Issue #221, opened 2016-12-08
IMLeagues


enzanki-ars:
### URL(s) where the issue occurs

http://www.imleagues.com/spa/member/player

### Describe the issue

Using all filters by uBlock, and Ads/Privacy/Malware domains categories, logging into IMLeagues will present the following page message: 

![image](https://cloud.githubusercontent.com/assets/8580211/21027537/31745690-bd5f-11e6-81d1-24cd3d07ed48.png)

The filter `@@||www.imleagues.com/spa/ads/show_ads.js$xmlhttprequest,domain=www.imleagues.com` fixes this issue (The script only contains the line `//no need to add anything`)

### Versions

- Browser/version: Google Chrome 55.0.2883.75 m (64-bit)
- uBlock Origin version: 1.10.0

### Settings

- All filters by uBlock, and Ads/Privacy/Malware domains categories.

### Notes
`@@||www.imleagues.com/spa/ads/show_ads.js$xmlhttprequest,domain=www.imleagues.com`
Fixes issue by allowing a js file with only the line `//no need to add anything`.

Relevant part of script the detects adblockers:

```js
// detect ad-blocking
var ad = document.createElement('imlAdvertisement');
ad.className = 'AdSense';
ad.style.display = 'block';
ad.style.position = 'absolute';
ad.style.top = '-1px';
ad.style.height = '1px';
document.body.appendChild(ad);

if (!ad.clientHeight) {
    $log.info('Ad is blocked..');
    try {
        $rootScope.$broadcast(constants.events.NOTIFY_AD_BLOCKED_EVENT, null);
    } catch (err) {}
} else {
    //recheck to prevent the case turn off blocker to login then turn on blocker
    //run this after 1 second since it may has some cost
    setTimeout(function() {
        $.getScript('/spa/ads/show_ads.js?v=134')
            .done(function() {
                $log.info('Ad is not blocked..');
            })
            .fail(function() {
                $log.info('Ad is blocked..');
                try {
                    $rootScope.$broadcast(constants.events.NOTIFY_AD_BLOCKED_EVENT, null);
                } catch (err) {}
            });
    }, 1000);
}
ga('set', 'dimension10', !ad.clientHeight);
document.body.removeChild(ad);
```

Another page uses

```js
setTimeout(function() {
    // detect ad-blocking
    var ad = document.createElement('imlAdvertisement');
    ad.className = 'AdSense';
    ad.style.display = 'block';
    ad.style.position = 'absolute';
[...]

PR #222, opened 2016-12-08
Fix anti-adblock on IMLeagues

enzanki-ars:
See #221 for description.


enzanki-ars:
Just discovered one last thing that needs to be fixed for the site to not detect the ad blocker.

migueldemoura:
https://github.com/uBlockOrigin/uAssets/issues/221#issuecomment-266155710

>Seems like a standard filter solution, please report to EasyList maintainers: https://forums.lanik.us/viewforum.php?f=62.

gorhill

PR #304, opened 2017-03-05
Add generic scriptlet to inject noop function.

### LINKED ISSUE #300: Generic filter to inject noop function by name?
Hi,

Would you mind adding generic filter to resources.txt which would inject noop function? This is common usecase to replace js functions.

Syntax could be:
`example.com##script:inject(noopfn.js, gtm4wp_track_downloads)`

I need this for this site https://naekranie.pl/recenzje/arrow-sezon-5-odcinek-14-recenzja because gallery doesn't load if `gtm4wp_track_downloads` is not defined. (was blocked before by easyprivacy). Defining it to noop fixes the issue. 

If generic idea is not acceptable could you add at least defuser for `gtm4wp_track_downloads` which would define it as noopfn. 

Here is direct link to blocked resource `https://naekranie.pl/wp-content/plugins/duracelltomi-google-tag-manager/js/gtm4wp-download-tracker.js`

Thanks,
Kacper

kasper93:
This can be helpful to unbreak sites after blocking certain scripts.

Fixes #300

Since there was no response in my issue. Here is PR, hope this time will be better :)

### DIFF
--- a/checksums/ublock0.txt
+++ b/checksums/ublock0.txt
@@ -1,10 +1,10 @@
 56b18eb692248fd0352066c62198c709 assets/ublock/filter-lists.json
 cadc15e72055199b194392c3e6814948 assets/ublock/badware.txt
 013363b068a39912b2261724b2a66306 assets/ublock/experimental.txt
-4535b21b211d4bd75eea6422c678e638 assets/ublock/filters.txt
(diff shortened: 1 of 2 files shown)


kasper93:
Closing because of the lack of real word use case. Can be reopened if such use case arise.

PR #328, opened 2017-03-24
Fix for additional MonkeyBroker checks

Akkuma:
At least one site is using the pbjs variable as another means of checking for MonkeyBroker

`https://www.hotslogs.com/Sitewide/HeroAndMapStatistics`

### Describe the issue

The below code is ran to stop uBlock and other blockers
```js
                function j14a0039a(source) {
                    
                    if (true)
                        window.location = "/Info/AdBloc" + "k?Source=" + source;
                    else
                        ga("send", "event", "AdFillMonkeyBroker", "Ad Blocked Acceptably", { nonInteraction: true });
                }

				try {
					var monkeyBrokerTemp = MonkeyBroker;
                }
            	catch(error) {
					j14a0039a("JavaScript Exception");
                }

            	if (typeof MonkeyBroker === "undefined")
                    j14a0039a("MonkeyBroker Undefined");
            	else if (typeof MonkeyBroker.regSlotsMap === "undefined")
					j14a0039a("regSlotsMap Undefined");
            	else if (typeof MonkeyBroker.defineSlot === "undefined")
					j14a0039a("defineSlot Undefined");
                else
                {
                    
                    var isAdExperienceLight = false;

                    try {
                        $(window).on("load", function(e) {
                        	if (typeof pbjs === "undefined" || pbjs.libLoaded !== true)
								j14a0039a("pbjs Undefined");
                            else if ($(".advertisementBannerMonkeyBrokerTop").height() === 0)
                                j14a0039a("Height 0");
                            else
                                setTimeout(function(){ if (typeof MonkeyBroker.slots === "undefined") j14a0039a("Slots Undefined"); }, 10000);
                                
                                
                        });

                        if (typeof MonkeyBroker !== "object" || typeof MonkeyBroker.addSlot === "undefined" || MonkeyBroker.addSlot.toString() === "function () {}")
                            j14a0039a("addSlot Undefined");
                        else if (!isAdExperienceLight)
                            ga("send", "event", "AdFillMonkeyBroker", "Ad Displayed", { nonInteraction: true });
                        else
                            ga("send", "event", "AdFillMonkeyBroker", "Ad Displayed Light", { nonInteraction: true });
                    }
            		catch(error) {
						j14a0039a("JavaScript Exception");
                    }
```
You'll be redirect away after 10s to a specific AdBlock page.

### Versions

- Browser/version: Chrome 56
- uBlock Origin version: 1.11.4

### Settings

- None

### Notes

hotslogs has been aggressively fighting blockers with multiple passes at stopping them and the latest either require firefox or cosmetic workarounds. Sit on the page for 10s at least before the redirect occurs.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -1516,6 +1516,7 @@ Object.defineProperty(window, 'is_block_adb_enabled', { value: false });
 d3pkae9owd2lcf.cloudfront.net/mb105.js application/javascript
 (function() {
 	var noopfn = function(){};
+	window.pbjs = { libLoaded: true };
 	window.MonkeyBroker = window.MonkeyBroker || {
 		addAttribute: noopfn,
 		addSlot: noopfn,
@@ -1524,7 +1525,8 @@ d3pkae9owd2lcf.cloudfront.net/mb105.js application/javascript
 		go: noopfn,
 		inventoryConditionalPlacement: noopfn,
 		registerSizeCallback: noopfn,
-		registerSlotCallback: noopfn
+		registerSlotCallback: noopfn,
+		slots: {}
 	};
 })();
 


gorhill:
> `https://www.hotslogs.com/Sitewide/HeroAndMapStatistics`

I loaded that URL in a tab, and the page rendered just fine. What is the issue and how can I reproduce it?

Akkuma:
The page gets redirected to https://www.hotslogs.com/Info/AdBlock after a `setTimeout`  of 10s

gorhill:
I am not seeing this -- using default settings:

a

gorhill:
I do get a immediate redirection with Nightly though.

Akkuma:
@gorhill Here is a video demonstrating it happening on a clean install uBlock in Firefox, complete with reset to defaults https://youtu.be/DyZ22OtvthI

gorhill:
I could reproduce on Chromium 56 -- I was testing with Chrome 59 (on which curiously the anti-adblock fails).

gorhill:
Ok code change looks fine to me, thanks for contributing.

gorhill:
I suppose the empty placeholder at the top of the page is not taken care by EasyList because users have to disable their blockers anyways with ABP?

I am thinking of adding `hotslogs.com###MainContent_advertisementBannerMonkeyBrokerTopReplacement` to uBO's filters.

Akkuma:
So one thing to keep in mind is that the current code is being extra sneaky by checking the height of elements that might be removed by blockers:

```js
else if ($(".advertisementBannerMonkeyBrokerTop").height() === 0)
                                j14a0039a("Height 0");
```

gorhill:
Yes I saw that after. How about `hotslogs.com##.advertisementBanner:style(height: 1px !important; min-height: 1px !important;)`?

mapx-:
classic filters:
https://forums.lanik.us/viewtopic.php?p=115278#p115278

okiehsch:
@Akkuma I am curious about your test with Firefox, because with https://github.com/uBlockOrigin/uAssets/commit/00704806d722cf0916c39ca2cbd22e1374edd5f5 I added
`hotslogs.com##script:inject(abort-on-property-write.js, FillMonkeyBrokerHorizontal1)` 
to the `uBlock filter` list.

I just tested `https://www.hotslogs.com/Sitewide/HeroAndMapStatistics` with  a clean 
Firefox 52.0.1 install and uBO as the only extension.
I dsabled  the `uBlock Filter` list and just added 
`hotslogs.com##script:inject(abort-on-property-write.js, FillMonkeyBrokerHorizontal1)`
to my own filter list. I can not reproduce any redirect.

gorhill:
@okiehsch I see, this explains why I could not reproduce with one browser but was able to reproduce with the other -- my uBlock filters were not up to date on all browsers, I did not notice your commit.

Given this, I feel like reverting the change in this pull request in the spirit of minimizing changes for what was already working fine.

okiehsch:
Hotslogs is not the only site using that script, so maybe the change can be useful for other sites,
just my two cents

PR #344, opened 2017-04-01
Update filters.txt

zero77:
Some new filters to block adds and other annoying objects on different websites.


gorhill:
Submit all cases with a formal issue, with all details for people to investigate whether there is ground to add filters to uBO's own filters.

PR #346, opened 2017-04-02
Aliexpres sponsored item filter

peterwilli:
There is an new annoying "sponsored item" in the Aliexpress results-list that is most of the time not what you're looking for. I.e the sponsored item doesn't have to match your filters you set when searching on Aliexpress (it could be more expensive than the max price you set, or have paid shipping when you turned on free shipping etc) In general it's quite confusing.

This filter blocks the sponsored items only, I didn't know where else to post it. Sorry if this is the wrong way, this is the first time I submit a filter.

selectie_009

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -1288,3 +1288,5 @@ vivo.sx##script:inject(abort-on-property-write.js, _0x773d)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/345
 math-aids.com##script:inject(abort-on-property-write.js, __drizzleSettings)
+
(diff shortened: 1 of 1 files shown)


gorhill:
Open an issue and provide all details for people to reproduce.

peterwilli:
@gorhill it's not a bug.

gorhill:
This is a filter issue. **Always** provide a URL. This will be looked into, and you might be told to go to EasyList, or a more specific filter than what you provided will be crafted.

peterwilli:
@gorhill alright, thanks. I understand now!

PR #643, opened 2017-08-23
Fix for the popunder at serienstream.to

grhio:


gorhill:
Thanks for reporting. Fixed with `abort-current-inline-script.js`, to defuse based on inline script content rather than one single variable name.

PR #700, opened 2017-09-18
A: http://vidzi.tv/ld7hetg6pleo.html

SMed79:
FYI, the above blocked host is an [admeasures](https://forums.lanik.us/viewtopic.php?f=62&t=33320) custome adserver.


okiehsch:
Thx, btw `vidzi.tv` is another site that now uses crypto-mining.

PR #706, opened 2017-09-20
Added jsecoin.com (Web Browser Bitcoin Mining)

SMed79:
```
http://kharasach.com/
https://nicheextensions.com/
http://greysanatomy.com.br/
http://iseeitnow.xyz/
http://uasweekly.com/
http://howdoyouplayit.com/
http://ihavenet.com/
```

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -2201,3 +2201,4 @@ trustedreviews.com##body > div#wrapper:style(padding-top: 0px !important;)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/690
 ||coin-hive.com^$third-party
+||jsecoin.com^$third-party


gorhill:
Thanks.

gorhill:
I had assumed that these coin-mining scripts would all use web workers, but this is not the case here.

devJ0n:
I don't agree that jsecoin should be blocked.

It's a nice and ethical implementation where the end-users easily can opt-out by clicking on a link on each site which uses JSEcoin to monitize their site.

End-user are also able to globally opt-out/opt-in by visiting:
https://jsecoin.com/privacy/

Read about JSEcoin and make a better decision:
https://jsecoin.com/

PR #710, opened 2017-09-22
Add missing pubads methods

gazs:
This pull request adds methods listed in the [GPT documentation](https://developers.google.com/doubleclick-gpt/reference#googletagpubadsservice) but missing from the mocked pubads object:
- getTargeting
- getTargetingKeys
- setForceSafeFrame
- setSafeFrameConfig


gorhill:
You have one or more actual cases where these missing methods cause an issue?

gazs:
Ah, sorry, `http://www.vogue.de/mode` will be triggered on browser window resize

gorhill:
> http://www.vogue.de/mode will be triggered

Thanks.

However, you will first need to change `setForceSafeFrame` and `setSafeFrameConfig` to return `noopthisfn`, as per reference documentation.

gorhill:
Thanks.

PR #730, opened 2017-09-29
Update for tisoomi (uBlock#1044)

SMed79:


SMed79:
Sorry for the above noise.

PR #747, opened 2017-10-07
Added ppoi.org

SMed79:


okiehsch:
Thanks, but I just realized that  `EasyPrivacy` has added a whole bunch of `miner` filters, including
`ppoi.org`

@gorhill should filters be added to`resource-abuse` even if they are already part of `EasyPrivacy`?

gorhill:
My opinion is that yes, they should be added.

PR #748, opened 2017-10-08
Add webmine.cz to the list

ondrejsevcik:

One example of such a website is pratele.nikee.net. After a few minutes, my computer's battery is drained and my CPU usage is very high for this tab -> you can see in the console that it's because of this miner script. 

### Describe the issue

webmine.cz provides script for other websites to mine cryptocurrencies in your browser while you're reading content of the website.

### DIFF
--- a/filters/resource-abuse.txt
+++ b/filters/resource-abuse.txt
@@ -33,3 +33,5 @@
 ! https://github.com/hoshsadiq/adblock-nocoin-list/issues/32
 ||ppoi.org^$third-party
 ||projectpoi.com^$third-party
+
+||webmine.cz^
--- a/filters/resource-abuse.txt
+++ b/filters/resource-abuse.txt
@@ -34,4 +34,4 @@
 ||ppoi.org^$third-party
 ||projectpoi.com^$third-party
 
-||webmine.cz^
+||webmine.cz^$third-party


gorhill:
Thanks.

However, use the `third-party` option please to avoid breaking the site `webmine.cz`.

ondrejsevcik:
Added. I'm not sure how to use it so I hope it's ok. Correct me if I'm wrong, thank you.

SMed79:
@ondrejsevcik Static filter syntax https://github.com/gorhill/uBlock/wiki/Static-filter-syntax

gwarser:
Should this really be blocked?

https://webmine.cz/

>This is a visitor friendly miner, which will show a prompt to allow or decline mining on their computer. Therefore it shouldn't be adblocked or detected via antivirus software as discussed in our manifest.

Also https://github.com/keraf/NoCoin/issues/69 linking to https://github.com/keraf/NoCoin/issues/67#issuecomment-338442418

>I'm not sure it would be a good idea to block AuthedMine as the intention behind it is good. The goal of the extension is mainly to block sneaky miners and provide the user an option to opt-in when the site doesn't. This is exactly what AuthedMine does and I think adding it to the blacklist would be counter productive. This would also push miner services to just find new ways to bypass mine blockers.


And I don't see pratele.nikee.net connecting to it.

gorhill:
For sure when the pull request was made the site `pratele.nikee.net` was mining without user content.

See <https://webmine.cz/api>:

> Auto Start Miner
> 
> We provide a miner which is able to start automatically without disrupting visitors by any prompts.
> 
> However, keep in mind that this kind of mining is often treated badly by adblockers and antiviruses.

The visitor-friendly one has a different domain: `authedwebmine.cz` which is not in _"Resource abuse"_.

gwarser:
Yes, I checked this more, and when you go to  `webmine.cz/miner` mining is started without asking.

ondrejsevcik:
I don't see a reason why it shouldn't be blocked. They might ask for a user consensus today, but it may change again without any notice. One of the reason we use adblocker is to protect website visitors from such a "badass" behaviour. After all, there is no benefit for the user.

m9ra:
We are doing our best to prevent any possibilities of mining via authedwebmine.cz domain without user confirmation and we won't ever change it. Team webmine.cz.

PR #847, opened 2017-11-19
Update README.md

KonoromiHimaries:


gorhill:
I will decline. This just adds a 3rd-party request when visiting the front page for what is a completely non-essential piece of information.

jspenguin2017:
It's not quite 3rd-party, GitHub will always cache images and serve them from their server.

PR #872, opened 2017-11-29
Fix #871

### LINKED ISSUE #871: 100percentfedup.com Anti-adblock message

100percentfedup.com

### Describe the issue

Anti-adblock messae at the top of the screen. 

### Screenshot(s)

capture

### Versions

- Browser/version: Chrome
- uBlock Origin version: 1.14.19.102

### Settings

Default lists 

### Notes
The following filters fix the issue and remove the empty div space at the top of the screen on my end.

`100percentfedup.com##script:inject(abort-on-property-write.js, Math.random)`
`100percentfedup.com##p`

yourduskquibbles:
Fixes #871

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -2559,3 +2559,7 @@ magesy.be##script:inject(abort-current-inline-script.js, adBlockDetected)
 ! https://forums.lanik.us/viewtopic.php?p=128630#p128630
 tohax.com##script:inject(abort-on-property-write.js, Fingerprint2)
 @@||tohax.com^$generichide
+
+! https://github.com/uBlockOrigin/uAssets/issues/871
+100percentfedup.com##script:inject(abort-on-property-write.js, Math.random)
+100percentfedup.com##p


gorhill:
That is at most an annoyance, there is no list for annoyances on this repo. Please report to Fanboy Annoyance or Adguard Annoyance.

yourduskquibbles:
ok thanks, will add to my [Web Annoyances filter list project](https://github.com/yourduskquibbles/webannoyances).

gorhill:
I see you've put a lot of work in there, and over a long period of time. I am tempted to include it as a stock filter lists but it's getting crowded in that section.

mapx-:
better adding
```
100percentfedup.com##.modaal-overlay
@@||100percentfedup.com^$generichide
```

yourduskquibbles:
@gorhill Wow, that's awesome! My first thought is that if it became a stock filter list, it should NOT be enabled out of the box.

If you were going to add it to available lists, I would probably recommend adding it as an optional (disabled by default) Annoyance list and let people opt-in to it if they want that type of functionality.

If it was enabled by default I think it may piss people off since it alters how sites are visually displayed. I can just imagine it being interpreted poorly as it changes users established workflows and 'desired' site behavior that users have become used to, e.g. The headers taking up part of the screen as you scroll. 

There is also a cat-and-mouse game that could develop if it were enabled by default that I've been thinking of opening an issue tracker for on uBO. (i.e. uBO ':style(foo !important;)' filters don't work when the site uses '!important;' on the same 'foo' element in their CSS. I assume due to race condition, but trying to think how uBO filters with '!important;' could have higher priority than the site's elements using the '!important;' tag.

okiehsch:
>  it should NOT be enabled out of the box

I am pretty sure he does not intend to make an annoyance filter list part of uBO's default setup 😉

Just fyi, your filter `100percentfedup.com##script:inject(abort-on-property-write.js, Math.random)`
will break quite a bit of site functionality, for eample comments, dropdown menus, embedded twitter videos, etc.
If you want to use a `script:injection` filter to fix the annoyance, I would suggest
`100percentfedup.com##script:inject(abort-current-inline-script.js, setInterval, Blocker)`
or simply use the filters suggested by `mapx-`.

yourduskquibbles:
Thank you @mapx- and @okiehsch! I will use the filters suggested by @mapx- and leave the script inject filters to you guys :)

PR #876, opened 2017-12-01
Added coinhiveproxy.com

SMed79:
`https://csgoconfigs.com/`


okiehsch:
To prevent `csgoconfigs.com` from starting the mining I will also add
`csgoconfigs.com##script:inject(abort-current-inline-script.js, m, CH.Anonymous)`

PR #937, opened 2017-12-10
Update hoshsadiq/adblock-nocoin-list#64

SMed79:


SMed79:
@mapx- Could you please remove the filter option `first-party`
see https://github.com/hoshsadiq/adblock-nocoin-list/commit/02885ba1c651cd1092788222dcc155afa1af5b43#commitcomment-26178881

PR #955, opened 2017-12-10
Add remove-frame-content.js scriptlet

vobruba-martin:
I use this scriptlet to remove some ads on `novinky.cz` and `aktualne.cz`. Problem with these sites is that their CSS is loaded from an iframe so it is not possible to block ads which are loaded into iframes with rules like:

```
||novinky.cz$subdocument
novinky.cz##iframe
```

This script removes `<head>` & `<body>` elements if it detects that it is run inside an iframe (tested on Chrome 62 / Win10 and FF 57 / Linux).

I hope this can help on other sites too. Please comment.

Thanks!

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2046,3 +2046,70 @@ set-constant.js application/javascript
 		}
 	});
 })();
+
+
+# This removes BODY & HEAD elements if run inside framed content.
+remove-frame-content.js application/javascript
+(function() {
+	var log = console.log.bind(console);
+	var removedHead = {
+		async: false,
+		result: false
+	};
+	var removedBody = {
+		async: false,
+		result: false
+	};
+	var url = 'unknown';
+	try {
+		url = window.location.href || url;
+	} catch (e) {};
+	function inIframe() {
+		try {
+			return window.self !== window.top;
+		} catch (e) {};
+		return true;
+	}
+	function removeHead() {
+		try {
+			// This is sometimes causing 100% CPU usage.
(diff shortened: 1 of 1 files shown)


vobruba-martin:
Rebased

uBlock-user:
What are the params?

vobruba-martin:
No params are needed. Use it like this `novinky.cz##script:inject(remove-frames-content.js)`

uBlock-user:
That will remove all iframes, I wouldn't want that.

filters/resources.txt
```diff
@@ -2046,3 +2046,67 @@ set-constant.js application/javascript
...
+
+# This removes BODY & HEAD elements if run inside framed content.
+remove-frame-content.js application/javascript
+(function() {
+	var log = console.log.bind(console);
+	var removedHead = {
+		async: false,
+		result: false
+	};
+	var removedBody = removedHead;
```
jspenguin2017: This will not copy the object, both variable will point to the same object in memory.

vobruba-martin: Thanks!

gorhill:
I need one or more (preferably) real-world cases (specific URLs etc) where I can see the code at work. I want to fully understand the issue which this pull request is trying to solve to find out whether I agree this is the right solution.

vobruba-martin:
Using this scriptlet I was trying to fool antiadblock system which are using czech media sites. Since I don't have much time to test it all again, I think we can close this PR. Thanks

PR #1050, opened 2017-12-19
Add Swedish i18n to Facebook sponsored content

Zalastax:
It might be a good idea to add i18n for all languages on Facebook

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -591,7 +591,7 @@ facebook.com###stream_pagelet [id^="hyperfeed_story_id_"]:has(span._4dcu)
 ! https://github.com/uBlockOrigin/uAssets/issues/722
 facebook.com##.ego_column:if(a[href^="/campaign/landing"])
 ! https://forums.lanik.us/viewtopic.php?p=128997#p128997
-facebook.com##[id^="hyperfeed_story_id_"]:if([id^="feed_subtitle_"] a[href]:matches-css-after(content:/Gesponsert|Sponsored/))
+facebook.com##[id^="hyperfeed_story_id_"]:if([id^="feed_subtitle_"] a[href]:matches-css-after(content:/Gesponsert|Sponsored|Sponsrad/))
 
 ! https://forums.lanik.us/viewtopic.php?f=62&t=32142
 ||ruiguozhi.cc^


gorhill:
Could you try this new filter instead?

    facebook.com##[id^="hyperfeed_story_id_"]:if([id^="feed_subtitle_"] a[class^="g__"]:matches-css-after(content:/./))

I am trying to improve efficiency and broadness.

Zalastax:
That does not work for me. Facebook minimizes their class names so it's really tricky to find a rule that won't break by every minor change they make. I don't see any readable class names with more specificity than `feed_subtitle`.

mapx-:
Could you test:

`facebook.com##[data-testid="fbfeed_story"]:if([href*="[is_sponsored]"])`

Zalastax:
That does work! Any risk that a URL could contain `&ft[is_sponsored]=0` meaning we would block unsponsored content? `facebook.com##[data-testid="fbfeed_story"]:if([href*="[is_sponsored]=1"])` seems smarter, right?

Just a heads up, I did stumble upon one sponsored post without `is_sponsored` in the URL. After refreshing the page I couldn't find any more cases, even though I got the same post presented again, this time filtered away.

mapx-:
I tested a FB page in italian and another one in english (2 different systems) and seems working fine

gorhill:
Good find.

I think it helps to use the element name in such case, `a`, I am pretty sure this makes `querySelector` faster, as theoretically, `href` could be an attribute on any element -- `querySelector` can't assume the attribute is only present on specific elements.

Zalastax:
Facebook is unfortunately not consistent and I've found sponsored links without `is_sponsored`. The URL is https://www.facebook.com/Hearthstone.en/?hc_ref=ARSz_WNFzUor4i8M6ypp_OdrdjkQqb-jhjQGrAjAWWFkcb9iWzE5rsLuPc4swmG9sHs&fref=nf, and this is the HTML of the element: https://gist.github.com/Zalastax/54811deb62011e4b409e31bbdc08112f

gorhill:
You left out too much useful information in your Gist. The HTML code starting from `id="hyperfeed_story_id_[...]"` (inclusive) needs to be provided.

Zalastax:
I've updated the [gist](https://gist.github.com/Zalastax/54811deb62011e4b409e31bbdc08112f) and here's a screenshot of the ad: ublock_fb
It's fairly rare that something slips through, so it might still be best to use the current rule.

Zalastax:
@gorhill The recently added filter is way too lax! It basically removes every single post on my Facebook feed.
I've uploaded [the HTML](https://gist.github.com/Zalastax/2deafc707a9dc8064af8a0dd3ee3a10f) of a [post](https://www.facebook.com/groups/242675022610339/permalink/759626937581809/) that should not have been filtered.

What caused you to add that rule? For me, is_sponsored has worked well.

One pattern I've noticed is that every normal post has numbers in it, either x hours, or x date. What if we modify the regex to: `/^[^0-9]+$/`?

gorhill:
> What caused you to add that rule?

I've seen [complaints](https://www.reddit.com/r/uBlockOrigin/comments/7muu6z/facebook_sponsored_ads_bar_finally_blocked/) that uBO was blocking Facebook ads, and having to support every language seems like an impossible task. I went we what I saw in my timeline, but then I am not a user of Facebook. I think it's best if I leave it to people who uses it to come up with filter solutions.

PR #1461, opened 2018-02-03
Update recipes_en.txt

pwd-github:
Added discord recipe


gorhill:
I think you need to add `_ 1st-party script`: blocking all scripts by default should be deemed common enough to warrant adding a rule to ensure the recipe will work for those who do block all scripts everywhere by default.

pwd-github:
Thanks @gorhill . It was just a quick and easy one to add to make sure I got it right and understood it properly. I've updated the "patch" and will now start thinking of more common sites to add.

gorhill:
Thanks.

PR #1468, opened 2018-02-03
Outlook

pwd-github:
Outlook/Hotmail


gorhill:
This needs fixing, conflicts arose as a result of committing Proton recipe.

If you have multiple recipes, it will be easier if you commit them all in a single pull request.

pwd-github:
Okie doke, I'll just collate them and add them all together next time.

gorhill:
If you have any idea how things can be improved just share. I still consider all this to be experimental and I sure want to make all the right choices before it's too late.

For now on my side I am working on extending the UI to make it easier to craft recipes.

PR #431, opened 2017-05-31
This fixes #430

anewuser:
This fixes #430.

Issue #430, opened 2017-05-31
freeroms.com


anewuser:
### URL(s) where the issue occurs

http://www.freeroms.com/dl_roms/rom_download.php?title=e_t_-the_extra_terrestrial&system=Atari&game_id=3131

### Describe the issue
This sites shows a suspicious download link to an .exe file by default, and sets a timer to show the actual download link.

A standard rule to hide the link (`freeroms.com##a.download`) breaks the alignment of  the filesize/votes/rating columns. Just setting the element `visilibity` to `hidden` instead doesn't have this problem.

### Versions

- Browser/version: Firefox 53.0.3
- uBlock Origin version: 1.12.4

PR #1472, opened 2018-02-04
pcworld.com - Amazon leftover

anewuser:

https://www.pcworld.com/article/3081525/software-productivity/how-to-organize-your-gmail-using-mulitple-inboxes.html

### Describe the issue

"Shop Tech Products at Amazon" leftover header at the end of the article.

### Versions

- Browser/version: Firefox 58.0.1
- uBlock Origin version: 1.15.4

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -4859,3 +4859,6 @@ engageme.tv##script:inject(abort-current-inline-script.js, stop, adblock)
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/7v0h2t/any_help_blocking_these_messages/
 @@||bludv.com^$generichide
+
+! pcworld.com leftover
+pcworld.com###amazon-bottom-widget
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -4857,8 +4857,8 @@ backin.net##script:inject(window.open-defuser.js)
 ! https://github.com/jspenguin2017/uBlockProtector/issues/840
 engageme.tv##script:inject(abort-current-inline-script.js, stop, adblock)
 
-! https://www.reddit.com/r/uBlockOrigin/comments/7v0h2t/any_help_blocking_these_messages/
+! https://github.com/uBlockOrigin/uAssets/issues/1472
 @@||bludv.com^$generichide
 
-! pcworld.com leftover
+! pcworld.com Amazon leftover
 pcworld.com###amazon-bottom-widget


anewuser:
Sorry, ignore this.

anewuser:
@gorhill Can you please fix or delete this? I accidentally made a [bad commit](https://github.com/uBlockOrigin/uAssets/pull/1472/commits/d15c60236b1b66112182c966a611ea1644ccfe60) along with [the one I wanted to submit](https://github.com/uBlockOrigin/uAssets/pull/1472/commits/404734b2fed2f8f6fb9ab6a9615d465345402e3b).

You had told me there's [no need to open a separate issue](https://github.com/uBlockOrigin/uAssets/pull/431), but then is it okay to just describe it in the comment that precedes the filter (e.g. `! pcworld.com - Amazon leftover`), since there's no link to add? I made this mess because I had manually checked the future ID of this pull request to add it to the comment, and accidentally typed it in the wrong place, but I know that could've caused other problems too.

gorhill:
Sorry I can't delete issue, only GitHub people can.

When I said "no need to open a separate issue", I meant to just put all the information you put in #430 into #431 (hence no need for #430) -- pull requests are like plain issues, except they come with code changes. But maybe I misunderstood and you opened #430 _before_ deciding to provide a fix in #431, in such case disregard my "no need to open a separate issue" -- I had assumed you opened #430 just to get a reference issue for #431 (you created #431 less than one minute after #430).

In case what I say here is even more confusing, in other words: if you know of an issue and have a fix, just opening a pull request with the details about the issue is enough.

anewuser:
>I had assumed you opened #430 just to get a reference issue for #431

That's what I did, and I got that all the info can be added to pull requests. My question is: what should I add to comments preceding filter rules?

For example, here's the commit I made:

`+! https://github.com/uBlockOrigin/uAssets/issues/430
+freeroms.com##a.download:style(visibility:hidden)`

What should it say in place of `! https://github.com/uBlockOrigin/uAssets/issues/430`?

okiehsch:
`! https://github.com/uBlockOrigin/uAssets/pull/431`

anewuser:
@okiehsch but how can I refer to a pull request before making it? Don't you get the ID just after confirming the commit?

gorhill:
Ok I see the problem, to have the code change refers to the pull request which is itself the code change.

gorhill:
Well I guess one way to do this is to just modified the pull request afterward, I always use squash-and-merge so that everything appears as a single commit. Is this something complicated to do? (can't tell given that on my side I can push directly).

okiehsch:
The next opened issue will be `https://github.com/uBlockOrigin/uAssets/issues/1500`, so if I open a pull request right now ,  `https://github.com/uBlockOrigin/uAssets/pull/1500` should be the reference link, or am I missing something?

anewuser:
@okiehsch That's what I tried to do here (and pasted the URL in the wrong place).

The problem with that is that somebody else could post another issue before you posted it. @gorhill's solution is simple, just edit it again after being certain about the pull request ID.

okiehsch:
OK, I did indeed miss something ;) and yes that looks like the simplest solution.

PR #1477, opened 2018-02-05
uMatrix rules: update Github rules

ArchangeGabriel:
Those are required for some features (see added comments):
– attaching files to comments;
– viewing diff of some binary files.


gorhill:
Thanks.

PR #1479, opened 2018-02-05
uMatrix Recipe: Netflix

warnellw:
This recipe unbreaks Netflix.

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -54,6 +54,15 @@ Google Maps
 		! needed for satellite view
 		no-workers: _ false
 
+Netflix
+	netflix.com *
+		_ 1st-party *
+		_ 1st-party cookie
+		_ 1st-party script
+		_ 1st-party xhr
+		_ nflxext.com script
+		_ nflxvideo.net xhr
+
 Outlook (Formerly Hotmail)
 	live.com *
 		_ 1st-party *


warnellw:
Should any base uMatrix rulesets be assumed when creating recipes? I added the 1st party rulesets as observed from other recipes, however, they are not needed if the default uMatrix rulesets are used (i.e. allowing all first party requests/resources).

gorhill:
> Should any base uMatrix rulesets be assumed when creating recipes?

I need to add documentation about this, and I so far what I have decided is that we should use the more granular rules for non-css/image (passive) content, but only if they really are needed. This also includes the per-switch (many sites uses web workers now).

> however, they are not needed if the default uMatrix rulesets are used

uMatrix will import only the rules which are needed, it tests the target cell to find out whether a rule from the recipe is needed. The rules are applied from top to bottom, so we need to start with the broader one (`_ cdn.com *`), then the narrower ones (`_ cdn.com script`).

gorhill:
Thanks.

anewuser:
I think that it'd be more practical to assume that 1st-party requests (but not web workers and referers) are whitelisted. That would keep the list cleaner and make creating and testing new rules easier.

Users who blacklist 1st-party content by default are extra cautious and less likely to enable pre-made recipes, anyway.

gorhill:
> I think that it'd be more practical to assume that 1st-party requests (but not web workers and referers)

Ok I agree that `_ 1st-party *` should be removed. I am still unsure about more specific 1st-party rules. I worry that users who configured uMatrix to block scripts out of the box  as documented in the wiki or even other uMatrix guides I've seen elsewhere (when uMatrix is proposed as a replacement of NoScript) may end up having unmatched expectations about a recipe not un-breaking a site. How about we at least keep `_ 1st-party script` as part of recipes when needed?

ArchangeGabriel:
I agree about script. I know a lot of people that block all scripts by default (starting with me) and even often cookies.

anewuser:
My fear is that people will start adding dozens of domains to the recipe list(s) just to whitelist 1st-party cells, but I like the compromise of keeping just `1st-party script`.

PR #1482, opened 2018-02-06
uMatrix Recipe: Netflix

warnellw:
These additional rules are more granular and fix assumptions made in the first version of the recipe.


gorhill:
Which specific resources are needed for `nflximg.net *` and `nflxso.net`?

warnellw:
Those domains are only needed for css/images. I commented on my other pull request (#1482) as to my reasoning for what I assumed the minimum resources a wildcard would cover.

PR #1484, opened 2018-02-06
uMatrix Recipe: Wikia

anewuser:
Wikis are displayed as text-only without this. 

Example: http://linux.wikia.com/wiki/Linux_Wiki

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -112,6 +112,12 @@ Twitter with account
 		_ 1st-party xhr
 		_ twimg.com *
 		_ twimg.com script
+		
+Wikia
+	wikia.com *
+		! wikis are displayed as text-only without this 
+		_ nocookie.net *
+		_ nocookie.net script
 
 Vimeo as 3rd-party
 	* player.vimeo.com


gorhill:
Probably also need:

    _ 1st-party *
    _ 1st-party script
    _ www.adumen.com xhr

?

With the two first ones, some features are not working (dropdown menus), and without the 3rd one, I get a full page warning about blocker.

anewuser:
The official uBlock filters have scriptlets that take care of that message.

Should uMatrix recipes whitelist stuff that uBO fixes by default?

gorhill:
> Should uMatrix recipes whitelist stuff that uBO fixes by default?

I think we should assume uMatrix is used without uBO (or could be used with another blocker which does not deal with the issue).

Though note that as opposed to what I said elsewhere in the current case we do not want more than `_ www.adumen.com xhr` given the purpose of request.

gwarser:
https://old.reddit.com/r/uMatrix/comments/9frhjv/default_ruleset_recipe/

Use `wikia.nocookie.net` insetead of `nocookie.net`?

matrix

PR #1485, opened 2018-02-06
uMatrix Recipe: Twitch

warnellw:
uMatrix recipe to unbreak Twitch.

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -73,6 +73,26 @@ Proton
 		_ 1st-party script
 		_ 1st-party xhr
 
+Twitch
+	twitch.tv *
+		no-workers: _ false
+		_ 1st-party *
+		_ 1st-party cookie
+		_ 1st-party frame
+		_ 1st-party script
+		_ 1st-party xhr
+		_ amazonaws.com *
+		_ amazonaws.com xhr
+		_ jtvnw.net *
+		_ media-amazon.com *
+		_ ttvnw.net *
+		_ ttvnw.net script
+		_ ttvnw.net xhr
+		_ twitchcdn.net *
+		_ twitchcdn.net script
+		_ twitchsvc.net *
+		_ twitchsvc.net script
+
 Twitter no account
 	twitter.com *
 		_ 1st-party *


gorhill:
Isn't the rule below missing?

    _ twitchsvc.net *

warnellw:
I was under the impression that the wildcard (*) should only be assumed (at a minimum) for passive css/images based on a user's specific global rules. `twitchsvc.net` only serves scripts, which is why I excluded the wildcard (as no passive content was required from that domain). 

For reference, below is the base ruleset that I was using to create recipes. It assumes a user has explicitly blocked everything including first party resources, except css/images. With these rules, whitelisting a 3rd party domain (*) only allows css/images (at a minimum, assuming a user's ruleset is very strict). Then granular rules followed allowing scripts, xhr etc.

```
* * * block
* * cookie block
* * frame block
* * media block
* * other block
* * script block
* * xhr block
* 1st-party css allow
* 1st-party image allow
```

gorhill:
> I was under the impression that the wildcard (*) should only be assumed (at a minimum) for passive css/images based on a user's specific global rules.

I consider that the natural way of using uMatrix is to stick to set rules for all types (using the wildcard) for any given domain. I consider that typically using specific-hostname/specific type rules are strictly to override broader block rules (for example the built-in `1st-party * frame allow`).

So `_ twitchsvc.net *` should come first (because natural way of using uMatrix), then `_ twitchsvc.net script` for those who block all scripts by default.

I know there are people out there that will set a very specific rules on every single cell for which there is a count, but really that is just not the way I intended ruleset to be managed as the natural way to do so.

warnellw:
> I consider that the natural way of using uMatrix is to stick to set rules for all types (using the wildcard) for any given domain.

For the sake of keeping the recipes as short as possible (this one is becoming quite large), it might make sense to consider a short-hand way to write the recipes. Take the following as an example:

```
Example
	example.com *
		_ 1st-party *
		_ 1st-party script
		_ 1st-party xhr
		_ 1st-party frame
		_ cdn-1.com *
		_ cdn-1.com script
		_ cdn-1.com xhr
		_ cdn-2.com *
		_ cdn-2.com xhr
		_ cdn-2.com frame
```

Could be written as:

```
Example
	example.com *
		_ 1st-party * script xhr frame
		_ cdn-1.com * script xhr
		_ cdn-2.com * xhr frame
```

At any rate, I will update this recipe with `_ twitchsvc.net *`.

ArchangeGabriel:
There is a proposed (and accepted by @gorhill, just not yet implemented) enhancement for uMatrix to support things like `example.com cdn-1.com frame,script,xhr` to write more concise rules. So this could be added here as well.

PR #1489, opened 2018-02-07
uMatrix Recipe: Dailymotion

anewuser:


gorhill:
This rule:

    _ googleapis.com script

Should really be:

    _ ajax.googleapis.com script

We should stick as much as possible to effective base domain: `googleapis.com` is a public suffix, not a base domain, as per PSL.

Not sure we should include

    _ 1st-party cookie

I could play fine a video I randomly picked. If the rule is to allow signing in, then we should create a separate recipe for "Dailymotion with account".

anewuser:
I remove the cookie line per https://github.com/uBlockOrigin/uAssets/pull/1479#issuecomment-363830907.

What about having a list of recipes for CDNs covered by Decentraleyes?

@Synzvato

PR #1490, opened 2018-02-07
uMatrix Recipe: Steam

anewuser:


gorhill:
These two rules are wrong:

    steamstatic.com *
    steamstatic.com script

They are missing the scope placeholder `_`.

Also, on my side randomly clicking a page with a video, `_ steamstatic.com media` seems to be needed.

anewuser:
Sorry, I block only 3rd-party `cookie`, `script` and `frame` by default, and forgot to add rules for `media` when I copied rules for this and IMDb.

gorhill:
I can't find where, but you asked me somewhere if you can add more recipe files for other languages, the answer is yes.

PR #1513, opened 2018-02-09
uMatrix Recipe: Update YouTube 3rd-Party

warnellw:
Some websites require resources from `ytimg.com` to render YouTube videos. Example: 
https://www.theverge.com/2018/2/9/16988630/spacex-falcon-heavy-launch-report-video

I've only noticed resources coming from the subdomains (`i.ytimg.com` and `s.ytimg.com`), so I kept the rules as strict as possible. I am also happy to update the rules to the following:

```
_ ytimg.com *
_ ytimg.com script
```

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -220,3 +220,5 @@ Youtube as 3rd-party
 		_ youtube.com frame
 		_ googlevideo.com *
 		_ googlevideo.com xhr
+		_ ytimg.com *
+		_ ytimg.com script


anewuser:
We'd rather use the rules in your update suggestion instead of dissecting sub-domains, unless the domain is listed here: https://publicsuffix.org/list/public_suffix_list.dat

Note that we're also assuming that users have uBlock Origin installed at that same time, so you can also file issues for specific ad/tracking resources being whitelisted by recipes, if any.

gorhill:
> Note that we're also assuming that users have uBlock Origin installed at that same time, so you can also file issues for specific ad/tracking resources being whitelisted by recipes, if any.

I don't want such assumption to be made, uMatrix should be considered on its own for recipe-crafting purpose.

anewuser:
I misread what you had said elsewhere, sorry.

But do you agree that for cases like this it's better to have what it's in his suggestion? I.e.

    _ example.com *
    _ example.com script

rather than 

    _ a.example.com image
    _ b.example.com script

?

warnellw:
@anewuser Thanks for the link to the public suffix list.

@gorhill This should be ready to go now.

gorhill:
> But do you agree that for cases like this it's better to have what it's in his suggestion?

Really, it depends. If there are reasons to believe `ytimg.com` could cause unwanted/unnecessary network requests to some specific subdomains to be allowed, narrowing is ok then. It's a case by case decision. For example, I allowed `player.cnevids.com` for Ars Technica because I didn't find that more than this one subdomain of `cnevids.com` was necessary. In other cases, for example `googlevideo.com` on Youtube, it makes sense. It all depends on our findings about what uMatrix reports.

PR #1524, opened 2018-02-10
uMatrix Recipes: Popular 3rd-party CDNs

warnellw:
This is a work in progress. I'm opening the pull request now to get feedback on how these recipes should be properly implemented, if at all. These CDNs were pulled directly from the [Decentraleyes wiki](https://github.com/Synzvato/decentraleyes/wiki/Frequently-Asked-Questions#for-umatrix-and-ublock-origin-non-easy-mode-users), and @pwd-github has also expressed interest for these recipes in gorhill/uMatrix#30.

How should CDNs that have multiple domains (see Microsoft and Yandex) be implemented? I can separate them into multiple recipe blocks, but it seems keeping them together makes the most sense (in this case).

I could also use help finding real-world example websites that use these CDNs.

Thanks!

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -22,6 +22,16 @@ Ars Technica
 		_ d2c8v52ll5s99u.cloudfront.net script
 		_ dp8hsntg6do36.cloudfront.net *
 
+Baidu CDN as 3rd-party
+	* libs.baidu.com
+		* libs.baidu.com *
+		* libs.baidu.com script
+
+Cloudflare CDN as 3rd-party
+	* cdnjs.cloudflare.com
+		* cdnjs.cloudflare.com *
+		* cdnjs.cloudflare.com script
+
 Dailymotion
 	dailymotion.com *
 		_ 1st-party script
@@ -68,6 +78,11 @@ Google account sign in
 		_ gstatic.com *
 		_ gstatic.com script
 
+Google CDN as 3rd-party
+	* ajax.googleapis.com
+		* ajax.googleapis.com *
+		* ajax.googleapis.com script
+
 Google Maps
 	google.* *
 		_ 1st-party script
@@ -88,6 +103,11 @@ IMDb
 		! required to upload images
 		_ 1st-party frame
 		_ amazonaws.com script
+
+jsDelivr CDN as 3rd-party
+	* cdn.jsdelivr.net
+		* cdn.jsdelivr.net *
+		* cdn.jsdelivr.net script
 		
 JSFiddle
 	jsfiddle.net *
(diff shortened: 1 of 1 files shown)


gorhill:
I am skeptical about the need for such rules: the purpose of recipes is to unbreak stuff, and there is not guarantee that any of these recipes will un-break anything. Most of them are just the equivalent of clicking the domain itself in the matrix UI, there is not much added value.

My view on how to use Decentraleyes along with uMatrix is pretty simple: don't configure Decentraleyes for uMatrix. Just use uMatrix as if Decentraleyes was not there, and whenever some CDN needs to be allowed in uMatrix, it's just nice that Decentraleyes is there possibly catching and fulfilling some allowed requests.

pwd-github:
My opinion is that these rules should still be bound to Decentralyes in one way or another, so something like

```
Decentraleyes
 	*
 	 	* cdn.madeupsite.com script allow
 	 	* cdn.madeupsite2.com script allow
```

warnellw:
@gorhill 

> Most of them are just the equivalent of clicking the domain itself in the matrix UI, there is not much added value.

Right, I agree with this with regard to locally scoped rules. I made a mistake with the scoping, and really meant for these recipes to be global scope (which I've now corrected in the commit). However, I don't think global scope is working correctly in recipes. With the corrected global scope, uMatrix no longer shows an available recipe.

Using Google's CDN with the below recipe, and visiting Gmail (while logged out), I show no available recipes even though uMatrix has blocked a script from ajax.googleapis.com.

```
Google CDN as 3rd-party
	* ajax.googleapis.com
		* ajax.googleapis.com *
		* ajax.googleapis.com script
```

Is this syntax correct for globally scoped rules?

gorhill:
You recipe works fine for me. I found a page from HN's front page which listed `ajax.googleapis.com` and the recipe was available and I could apply it: <https://www.tnhh.net/posts/google-talk.html>.

warnellw:
My mistake; I was still testing with uMatrix v1.3.3b7. It's now working with v1.3.3b8 that was released a few hours ago.

I think these recipes could be useful for folks because many, many sites use these CDNs. It might be more convenient to only have to allow a CDN once, rather than allowing it on every website for which it is required. If anything, the recipes would save space in the `recipes_en.txt` document, as these rules could be omitted from all other recipes.

gorhill:
> I was still testing with uMatrix v1.3.3b7. It's now working with v1.3.3b8 that was released a few hours ago.

No changes related to recipes in b8. To be sure I tested with b7, and as expected it too works fine with it.

anewuser:
When @gorhill first announced recipes,  I also suggested covering Decentraleyes in some lost comment not many people must have seen. However, I had in mind creating a **separate recipes file for it**, which would be disabled by default, and not submitting them to the general English recipes file.  

I suggested this because at first I thought that recipes would add rules automatically, but since users will still have to manually import recipes one by one, now I agree with @gorhill that there is not much added value in offering this list in the form of recipes.

It's more practical to keep a list of public CDNs in the old uMatrix format (e.g. `* jsdelivr.net script allow`) for everyone interested in it to simply copy and paste them into the "My Rules" tab, like the list in [Decentraleyes's FAQ](https://github.com/Synzvato/decentraleyes/wiki/Frequently-Asked-Questions#for-umatrix-and-ublock-origin-non-easy-mode-users)   (which is outdated). 

We can have  CDNs fully supported by Decentraleyes  at the top, and add a warning before the ones not supported by it. If @gorhill believes /uAssets/ isn't the place for this, I'll maintain it elsewhere.

warnellw:
I am going to close this pull request as the consensus seems to be that these recipes shouldn't be included. I still think that there are various reasons to keep them (see my previous comment), and that they are more useful than simply saving a single click. However, maybe a different, more elegant solution will be implemented in the future.

Issue #1510, opened 2018-02-09
netlify.com


jayvdb:
### URL(s) where the issue occurs

`http://gci-leaders.netlify.com/`

### Describe the issue

An extremely broad rule is affecting all sites hosted by Netlify.
See https://github.com/easylist/easylist/issues/898 for most of the background

### Screenshot(s)

i18n not loading
![screenshot from 2018-02-09 08-28-01](https://user-images.githubusercontent.com/15092/36007292-63d73074-0d74-11e8-84bb-7b0d15e5bc1c.png)

### Versions

- Browser/version: Firefox 58.0
- uBlock Origin version: 1.15.4

### Settings

- Defaults

### Notes

https://github.com/easylist/easylist/issues/898

okiehsch:
I already adjusted a filter in uBO-unbreak to disable the broad rule in EasyPrvacy
https://github.com/uBlockOrigin/uAssets/blob/3a5ef36c5d615c52e7b48cda2616c8b38412423f/filters/unbreak.txt#L816-L817

If you update your uBO-unbreak list, the issue should be fixed.

The problem is that as long as EasyPrivacy does not remove `netlify` from that rule any change to it
makes the `badfilter` in uBO-unbreak obsolete.
Meaning as soon as EasyPrivacy adds a new domain to that rule the `badfilter` has to be edited accordingly.

jayvdb:
It seems that if I update the "EasyPrivacy" rulelist in UBlock preferences, the problem resurfaces.  is it possible that `uAssets/filters/unbreak.txt` is not being re-run when the rulelist is updated after the plugin is operational?

jayvdb:
> The problem is that as long as EasyPrivacy does not remove netlify from that rule any change to it
makes the badfilter in uBO-unbreak obsolete.

Ouch.
I assume this is because of a limitation in the syntax.
Is there a issue tracking this generic "maintainability" problem?

okiehsch:
> Is there a issue tracking this generic "maintainability" problem?

No, this is not normally an issue.
The `badfilter` option prevents another filter and only the functional form of that filter, to be loaded in memory, for example if a list would somehow add the filter
`||assets-cdn.github.com^$script`, adding `||assets-cdn.github.com^$script,badfilter`
would instantly solve any problems and such a filter does not change it's form, usually.

In this specific case one filter,  `*$websocket,xmlhttprequest,domain=,` causes the issue and that filter will change it's form every time a new domain has been added.

PR #1525, opened 2018-02-11
Remove netlify.com unbreak

jayvdb:
Unbreak rule added for https://github.com/uBlockOrigin/uAssets/issues/1510
is no longer needed due to the rule being removed upstream for
https://github.com/easylist/easylist/issues/898

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: [here]

### Settings

### Notes


okiehsch:
Thx.

anewuser:
Note that there's no "upstream". uAssets does expect users to have EasyList and other lists like the "Peter Lowe’s Ad and tracking server list" enabled, since they're enabled by uBlock Origin by default and it'd be a waste of time to recreate all rules they already have, but we aren't officially associated to or depend on them in any way.

If anything, _uAssets_ is "upstream", since it's the home of the only filters officially made for and provided by uBlock Origin (and its forks).

PR #1575, opened 2018-02-20
Adding recipes for Shopify and reCaptcha

pwd-github:
reCaptcha credit goes to /u/ZaphodBeebblebrox from the /r/uMatrix Reddit community

Shopify is a popular store backend used on a lot of medium sized sites.

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -131,6 +131,20 @@ Proton
 	protonmail.com *
 		_ 1st-party script
     		no-workers: _ false
+		
+reCaptcha
+	* google.com
+		_ google.com *
+		_ google.com script
+		_ google.com frame
+		_ gstatic.com *
+		_ gstatic.com script
+		
+Shopify
+	* shopify.com
+		_ cdn.shopify.com *
+		_ cdn.shopify.com script
+		_ cdn.shopify.com xhr
 
 Soundcloud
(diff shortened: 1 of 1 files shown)


ArchangeGabriel:
From my experience, reCaptcha also requires `xhr` for `google.com`.

ArchangeGabriel:
And I have this for Shopify:
```
cdn.shopify.com script allow
checkout.shopify.com frame allow
checkout.shopify.com script allow
checkout.shopifycs.com frame allow
checkout.shopifycs.com script allow
elb.deposit.shopifycs.com xhr allow
```

pwd-github:
My Shopify works well with what's in the patch. I've added the xhr to google.com though.

ArchangeGabriel:
Well as long as checkout and payment is not concerned, I can live with only `script` for `cdn.shopify.com` (well `image` and `css` as well, but I have left the default enabled for all sites currently, though I might change that in the future).

warnellw:
reCaptcha now requires web workers in some implementations:

recaptchawebworkers

ArchangeGabriel:
The worker is not strictly required, in the sense that it works without it. I’m opening a separated PR for reCaptcha only.

ArchangeGabriel:
https://github.com/uBlockOrigin/uAssets/pull/1997

PR #1623, opened 2018-02-25
dirpy.com

anewuser:

http://www.dirpy.com/
http://www.dirpy.com/studio?url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DYE7VzlLtp-4

### Describe the issue

- Front page ad section.
- Download page fake download/watch buttons.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -5386,6 +5386,6 @@ goldshorten.com##.banner-inner
 itdmusic.site##script:inject(popads-dummy.js)
 itdmusic.site##script:inject(abort-on-property-write.js, Fingerprint2)
 
-! dirpy.com
+! https://github.com/uBlockOrigin/uAssets/pull/1623
 dirpy.com##a[href*="bit.ly"]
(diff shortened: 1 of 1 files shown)


gorhill:
I don't see any ad on that site -- which by the way is blocked by both Firefox and Chromium.

anewuser:
Dirpy has been around for years. I didn't know it was blocked because I don't use the "Block dangerous and deceptive content" option on Firefox, but it seems it has been blocked exactly because of the fake buttons I've hidden with my commit  ([Firefox uses the Google blacklist](https://support.mozilla.org/kb/how-does-phishing-and-malware-protection-work#w_unwanted-software)):

> The site http://www.dirpy.com contains harmful content, including pages that:
> 
>     Try to trick visitors into sharing personal info or downloading software



 https://transparencyreport.google.com/safe-browsing/search?url=http:%2F%2Fwww.dirpy.com


## Front page
dirpy1

## Download page
dirpy2

PR #1735, opened 2018-03-18
Update #1733

SMed79:
* http\://www.linkeasy.org/js/home/5282.js
* http://i.imgur.com/rXCH4G5.png


mapx-:
pretty good for easylist

mapx-:
@ryanbr

SMed79:
So, why you have added `||hipercontas.com.br^` ??
The script from `linkeasy.org` is which open the popup !!

mapx-:
the main issue there was the popup, fixed using 
`universeseries.org^$csp=script-src 'self' 'unsafe-inline' data: *.fbcdn.net *.facebook.com *.google-analytics.com`

the `hipercontas` filter was for the ads

Issue #1722, opened 2018-03-16
fruitstreams.com


KonoromiHimaries:
`http://streamango.com/embed/keebnatkanmlsobk`

popads in other sites
`http://alltube.pl/riverdale/chapter-twenty-eight-there-will-be-blood/odcinek-15-sezon-2/137855`

to reproduce click streamango button and play the video
![2018-03-16-riverdale - odcinek 15 sezon 2 alltube - filmy i seriale online](https://user-images.githubusercontent.com/5884000/37519810-d981497e-291a-11e8-8109-584d12cb20df.png)

<details><summary>my settings</summary>

my filters
https://gist.github.com/KonoromiHimaries/48e9a12ec92d5ef105073ec1485326d3

Nano Adblocker 1.0.0.34
Nano Defender 13.43
Chrome 65.0.3325.162

</details>

uBlock-user:
patched in uBlock Filters already.

```
streamango.com##script:inject(openload.co.js)
streamango.com##script:inject(abort-on-property-read.js, BetterJsPop)
streamango.com##script:inject(window.open-defuser.js)
```

KonoromiHimaries:
@uBlock-user not worked in other sites

uBlock-user:
Works perfectly fine on uBlock Origin, you're using Nano Adblocker.

uBlock-user:
Works perfectly fine on uBlock Origin, you're using Nano Adblocker.

KonoromiHimaries:
right

KonoromiHimaries:
i still have a sometimes popup

uBlock-user:
can't reproduce it on my end. I click play it plays, nothing else happens.

KonoromiHimaries:
![untitled](https://user-images.githubusercontent.com/5884000/37524307-be8370ca-2929-11e8-9977-80d657d9c741.gif)

uBlock-user:
Add

```
||liczniki.org^
||ininmacerad.pro^
||debadu.com^
||puserving.com^
||ffinwwfpqi.gq^
```

KonoromiHimaries:
@uBlock-user still have a popads

uBlock-user:
Not enough blocking on your end, all those domains are blocked along with popads on my end and everything goes flawless...

mapx-:
test these filters
```
[...]

PR #1745, opened 2018-03-19
Update #1722

SMed79:


okiehsch:
That is a webrtc popup, so the best fix would be 
`example.com##script:inject(nowebrtc.js)`

I can't reproduce the popups, see https://github.com/uBlockOrigin/uAssets/issues/1722#issuecomment-374420293 where exactly do they occur on your end?

SMed79:
https://github.com/uBlockOrigin/uAssets/issues/1722#issuecomment-374444371

PR #1760, opened 2018-03-22
https://forums.lanik.us/viewtopic.php?f=62&t=40286 (bestarticles.me & tellycolors.me)

SMed79:


okiehsch:
Can you also add
`bestarticles.me,tellycolors.me##.adsbygoogle` to deal with the huge white spaces caused by  the generichide filter?

PR #1761, opened 2018-03-22
fix https://forums.lanik.us/viewtopic.php?f=62&t=40269 (usgamer.net)

SMed79:
https://forums.lanik.us/viewtopic.php?f=62&t=40269


okiehsch:
The anti adblock message is aready fixed, if you use a default setup, by
````
@@||usgamer.net^$generichide
@@||fastly.net/ad2/$image
````
the only way to reproduce, on my end, is to block `0914.global.ssl.fastly.net/ad2/img/x.gif`.
So, if you can reproduce, what filter does block `0914.global.ssl.fastly.net` requests.

SMed79:
> if you can reproduce, what filter does block 0914.global.ssl.fastly.net requests.

An entry in my hosts file (a custom Pi DNS).

Sorry for the noise.

Issue #770, opened 2017-10-18
add to resource abuse list pls


Zhangsun321:
https://authedmine.com/
coin miner.. 'opt in my ass'....

jspenguin2017:
The opt in is not bypassable by the site using it after a quick analysis, it is shown in a secure `iframe`. The whole point of resource abuse list is so that mining becomes opt-in, I don't think this as a threat until they start abusing it.

okiehsch:
I agree with @jspenguin2017, I had a look at `https://authedmine.com` yesterday and unless someone can provide an example of a site using `authedmine` without an explicit opt-in, I will not add it to the resource abuse list.

Zhangsun321:
ok.... understood...

uBlock-user:
The notice is cookie based. As long as the cookie and site data is NOT removed after being permitted once, it won't ask and will start mining.

jspenguin2017:
They said on their website that the cookie is valid for up to 24h.

lain566:
@okiehsch This, for example

`https://mylink.st/OqQzF7J`
<details>

![ublock _ _registro_de_peticiones_de_red_-_google_c_2017-10-19_23-41-12](https://user-images.githubusercontent.com/27441315/31795692-17a6ae14-b527-11e7-9d64-a5d580a61309.png)
</details>

jspenguin2017:
Well, if you click the Captcha, it's like giving consent. It does not load CryptoNight before you click it. 
![image](https://user-images.githubusercontent.com/7283682/31796992-30c7939c-b4e9-11e7-91dc-f55de626c915.png)

okiehsch:
Correct me if I am wrong but it looks to me like all mylink.st-links will only be displayed after you mine for some time, so blocking `authedmine.com` makes `mylink.st` non-functional.

jspenguin2017:
I'll exclude `authedmine.com` from my filters until I can see by myself that it mines automatically without letting you know. 
I consider a Captcha box to be enough of a consent.

lain566:
@jspenguin2017 Of course, I clicked, but not everyone is informed about this.

jspenguin2017:
@lain566 
The box itself is enough of a consent IMO, disabling uBO was used as a "consent box" because for those automatic miners we didn't really have other choices. But the side effect is that trackers and other stuff are unblocked as well, and when the Captcha finishes, you get redirected and you'll have a white list that you need to go into uBO settings panel to remove. 
[...]

PR #1777, opened 2018-03-23
Allow mining on mylink.li

mylinkadmin:
Thank you.

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: [here]

### Settings

### Notes

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -721,7 +721,7 @@ energy.de#@##ad_home
 @@||honeybadger.io/*/honeybadger.min.js$script,domain=clark.de
 
 ! https://github.com/uBlockOrigin/uAssets/issues/946
-@@||authedmine.com^$domain=mylink.nu|mylink.st
+@@||authedmine.com^$domain=mylink.li
 
 ! https://github.com/gorhill/uBlock/issues/3321
 @@||livechatinc.com^$generichide


ryanbr:
Personally if a site enforces coin mining to allow use, then adblockers shouldn't let it run. the problem isn't with the list, its an issue of the site and its bad design.

okiehsch:
I asked gorhill before I merged and he is fine with opt-in, i.e. `authedmine`.

SMed79:
> if a site enforces coin mining to allow use, then adblockers shouldn't let it run.

As i asked here https://github.com/uBlockOrigin/uAssets/issues/770#issuecomment-340907369

ryanbr:
"I have a charity website that relys on coin mining for income" (which was another excuse someone used in forums), which was still denied. If a site needs or use Coin mining, it should be blocked just like any other server. uBo is using a grey area to allow coin miners is bad for users.

Its either you block miners, or you don't, there isn't a "sometimes".

okiehsch:
> the right question is:  Can you access to the contents without mining? 

To clarify and to paraphrase gorhill:
If a filter blocking a miner breaks a specific site, and if the miner is shown to be entirely opt-in with no anti-user workaround, then an exception filter is acceptable to unbreak the site -- just like filter list maintainers accept to create exceptions for some other ubiquitous 3rd parties to unbreak sites.

ryanbr:
The difference here is they're raping your cpu in the process, and for the most part getting around an advert might involve a minor whitelist or a filter adjustment(s).

If a site is willing injecting coin miner without a whiltelist, then it should remain broken. The onus should be on the website to change not the lists, afterall they're the one's injecting the nasty code. I'm not expecting ubo to change their policy, I'm just outlining our thoughts on coin miners.

PR #1781, opened 2018-03-24
Update https://forums.lanik.us/viewtopic.php?p=134258#p134258

SMed79:
https://github.com/uBlockOrigin/uAssets/commit/7c39feed565bc3213950c7681a0f373ae4e804b6


mapx-:
I'm not sure the frame is only on FB site.

SMed79:
Safety first...

mapx-:
sure, it's safer, but if don't know where is that frame..

PR #1784, opened 2018-03-24
Allow "opt-in" mining from jsecoin.com

zolzomia:


gorhill:
Such generic whitelisting is not acceptable.

PR #1903, opened 2018-04-07
https://forums.lanik.us/viewtopic.php?f=62&t=40409 (fuckadblock 3.2.0)

SMed79:
```
https://www.cougarboard.com/
https://www.cougarboard.com/js/min/blockadblock-1441918196.js
```
https://forums.lanik.us/viewtopic.php?f=62&t=40409


okiehsch:
Thx.

Issue #1730, opened 2018-03-18
Cannot view nsfwyoutube.com with uBlock enabled, piguiqproxy.com


nero1024:
### URL(s) where the issue occurs

`http://www.nsfwyoutube.com`

### Describe the issue

Website NSFW YouTube fails to render when uBlock is enabled. Site is used for watching age-restricted YouTube videos without having to sign-in and mess up your recommendations list.

### Screenshot(s)

![clipboarder 2018 03 17-002](https://user-images.githubusercontent.com/20376599/37561553-1d77cce8-2a1f-11e8-99e2-0e5524502d3f.png)

### Versions

- Browser/version: Google Chrome Version 64.0.3282.186 (Official Build) (32-bit)
- uBlock Origin version: 1.15.2

### Settings

- see the attached image

### Notes

If I whitelist/allow piguiqproxy.com the site renders as normal however I get bombarded with a bunch of shady looking banner ads. I just want to watch age-restricted YouTube videos incognito without having to mess up my recommendations.

okiehsch:
Update your uBO-filter list, you can then noop `piguiqproxy.com`, or keep the block does not matter either way, the site will render fine and no "shaydy" looking ads will appear.

PR #1956, opened 2018-04-14
Update #1730 (nsfwyoutube.com)

SMed79:


mapx-:
What issue do you want to fix ? I see nothing new on that page.
Cant see `smcheck` there

PR #1971, opened 2018-04-15
fix #719 twitch.tv video ads

### LINKED ISSUE #719: twitch.tv

`https://www.twitch.tv/directory/all` &ndash; Click any of the live video channels.

### Describe the issue

Recently Twitch updated their ad technology causing advertisements to play at the start of any video stream. Previously it was trivial to skip them with AdBlock or similar, causing content to play immediately in lieu of the ad, but currently there appears to be no working solution to skip ads.

### Versions

- Browser/version: Firefox 55.0.3
- uBlock Origin version: v1.14.10

### Settings

- Default

pcworld:
### URL(s) where the issue occurs / Describe the issue

The issue affects video ads playing when loading a livestream on twitch.tv. These kinds of ads are not very reproducible, whether or not and when Twitch injects ads appears to be almost random. However, when Twitch decides that I'm to be targeted with video ads on some specific livestream, I can often F5 dozens of times and the ad will still always try to load (I never watched an ad to the end, I suppose it lets you watch the stream then and maybe the ad will not be shown anymore after F5). Note that these kinds of ads might only appear on "partnered" livestreams. Also, they're only appearing when the stream is actually running and not when it's offline. I'm including some random livestream I found where I could reproduce ads *and* my fix:

`https://www.twitch.tv/xchocobars`

Twitch's ad targeting is really weird, e.g. right now I can easily reproduce ads in above-mentioned livestream in Firefox (I have reloaded that tab dozens of times), even when starting new incognito sessions, but I do not get any ads in Chromium right now.

Twitch performs Fetch requests to URLs like `https://usher.ttvnw.net/api/channel/hls/xchocobars.m3u8?allow_source=true&baking_bread=true&baking_brownies=true&baking_brownies_timeout=1050&fast_bread=true&p=4237533&player_backend=mediaplayer&rtqos=control&sig=...&token=...` to retrieve a playlist. As @pixeltris found out and used in their webextension https://github.com/pixeltris/TwitchAdBlock, the `baking` parameters seem to be responsible for showing the pre-livestream video ads. Note that the presence of them does not indicate whether or not you will be delivered ads, but *without* their presence you won't get ads from my tests. With the above-mentioned livestream, I got video ads in Firefox pretty much every time reloading that page, and they always went away when adding my `inject` rule, and came back when removing it. I have reproduced that dozens of times.

Note that beside the video ads that appear when loading the stream, "partnered" Twitch streamers have the ability to press a button and show ads for presumably all/most viewers. I do not know whether my pull request fixes these ads as well, as this situation is not easy to encounter/reproduce.

My script hooks into the function of the Twitch player that loads the playlist from a URL that is passed as the first and only parameter and removes the `baking` parameters. I'm using [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to manipulate the URL, which is [not yet available in M$ Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/8993198/), however according to that bug report it is assumed to be included in its next release, and since that issue should be covered by my try-catch block, I do not see the need to change my usage of `URLSearchParams`.

### Screenshot(s)

[Screenshot of (paused) video ad](https://user-images.githubusercontent.com/1506734/38775116-4cde873a-407a-11e8-8796-220e1295930b.png)

### Versions

- Browser/version: Firefox 59.0.2, Chromium 65.0.3325.181
- uBlock Origin version: 1.15.24 on both browsers

### Settings

Nothing that should affect my PR. On Firefox I made a clean install of uBO and only changed the selection of lists, activated all the privacy features and enabled advanced user mode to add my own resources file.

### Notes

For people who want to test my PR: Enable [advanced mode](https://github.com/gorhill/uBlock/wiki/Advanced-user-features) and change the [`userResourcesLocation`](https://github.com/gorhill/uBlock/wiki/Advanced-settings#userresourceslocation) directive in the advanced mode configuration to:  
```
       userResourcesLocation https://gist.githubusercontent.com/pcworld/c3e36195f533ee0f32deec66d10bce76/raw/dd9e3dd17e1f1dcc3468c1088261a187c36b5f30/resources.txt
```
Note that by doing that you trust GitHub not to manipulate the code in there.
Then add the following line to the "My filters" tab of the dashboard and press "Apply changes":
```
twitch.tv##script:inject(twitch-videoad.js)
```
The linked resources file includes an additional log line. When you have configured everything correctly, when loading a livestream the F12 console should print "Removing ad parameters from playlist request". Though note that doesn't mean that Twitch would have played an ad otherwise. To test whether my PR works for you, you need to make sure that you are consistently delivered ads by Twitch (you should probably F5 or pause the ad before it finishes playing), *only then* enable the inject filter, reload the page a few times to ensure that you don't get ads anymore, disable the filter and then verify that you're getting ads again. If you have performed these steps it would probably be helpful if you post a comment on this PR whether this has worked for you.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2314,15 +2314,14 @@ damoh-defuser.js application/javascript
 
 twitch-videoad.js application/javascript
 (function() {
-	var ourMediaPlayer = undefined;
+	var ourMediaPlayer;
 	Object.defineProperty(window, 'MediaPlayer', {
 		set: function(newMediaPlayer) {
-			if (typeof(ourMediaPlayer) !== 'undefined')
-				return;
+			if ( ourMediaPlayer !== undefined ) { return; }
 			var oldLoad = newMediaPlayer.MediaPlayer.prototype.load;
 			newMediaPlayer.MediaPlayer.prototype.load = function(e) {
 				try {
-					if (e.startsWith('https://usher.ttvnw.net/api/channel/hls/')) {
+					if ( e.startsWith('https://usher.ttvnw.net/api/channel/hls/') ) {
 						var url = new URL(e);
 						url.searchParams.delete('baking_bread');
 						url.searchParams.delete('baking_brownies');
@@ -2330,7 +2329,7 @@ twitch-videoad.js application/javascript
 						e = url.href;
 					}
 				} catch (err) {
-					console.error('Failed to bypass Twitch livestream ad');
+					//console.error('Failed to bypass Twitch livestream ad');
 				}
 				return oldLoad.call(this, e);
 			};
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
     var observer = new MutationObserver(cleanVideoAsync);
     observer.observe(document.documentElement, { childList: true, subtree: true });
 })();
+
+
+twitch-videoad.js application/javascript
+(function() {
+	var ourMediaPlayer = undefined;
+	Object.defineProperty(window, 'MediaPlayer', {
+		set: function(newMediaPlayer) {
+			if (typeof(ourMediaPlayer) !== 'undefined')
+				return;
+			var oldLoad = newMediaPlayer.MediaPlayer.prototype.load;
+			newMediaPlayer.MediaPlayer.prototype.load = function(e) {
+				try {
+					if (e.startsWith('https://usher.ttvnw.net/api/channel/hls/')) {
+						var url = new URL(e);
+						url.searchParams.delete('baking_bread');
+						url.searchParams.delete('baking_brownies');
+						url.searchParams.delete('baking_brownies_timeout');
+						e = url.href;
+					}
+				} catch (err) {
+					console.error('Failed to bypass Twitch livestream ad');
+				}
+				return oldLoad.call(this, e);
+			};
+			ourMediaPlayer = newMediaPlayer;
+		},
+		get: function() {
+			return ourMediaPlayer;
+		}
+	});
+})();
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -6631,3 +6631,6 @@ yourlust.com##iframe[src="/im/footer.html"]
 ! https://github.com/uBlockOrigin/uAssets/issues/1970
 audioz.cc##script:inject(abort-current-inline-script.js, document.getElementById, adblockplus)
 audioz.cc##script:inject(bab-defuser.js)
+
+! https://github.com/uBlockOrigin/uAssets/issues/719
+twitch.tv##script:inject(twitch-videoad.js)


filters/resources.txt
```diff
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
...
+
+
+twitch-videoad.js application/javascript
+(function() {
+	var ourMediaPlayer = undefined;
+	Object.defineProperty(window, 'MediaPlayer', {
+		set: function(newMediaPlayer) {
+			if (typeof(ourMediaPlayer) !== 'undefined')
+				return;
+			var oldLoad = newMediaPlayer.MediaPlayer.prototype.load;
```
pcworld: This might break the site if Twitch removes the `MediaPlayer.MediaPlayer.prototype.load` property. So another try-catch around this and the following statement might be in order.

filters/resources.txt
```diff
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
     var observer = new MutationObserver(cleanVideoAsync);
     observer.observe(document.documentElement, { childList: true, subtree: true });
 })();
+
+
+twitch-videoad.js application/javascript
+(function() {
```
pcworld: I don't see a security issue here, but since this script is site-specific, should I include a check at the beginning that verifies that `document.domain === "twitch.tv"`? For that to work securely no site scripts must be allowed to run before this script, while I think that's true it's not mentioned [in the documentation](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#scriptlet-injection).

filters/resources.txt
```diff
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
...
+	var ourMediaPlayer = undefined;
+	Object.defineProperty(window, 'MediaPlayer', {
+		set: function(newMediaPlayer) {
+			if (typeof(ourMediaPlayer) !== 'undefined')
+				return;
```
kasper93: Why? I see no reason to return here, just patch new object and assign.

filters/resources.txt
```diff
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
...
+			var oldLoad = newMediaPlayer.MediaPlayer.prototype.load;
+			newMediaPlayer.MediaPlayer.prototype.load = function(e) {
+				try {
+					if (e.startsWith('https://usher.ttvnw.net/api/channel/hls/')) {
+						var url = new URL(e);
+						url.searchParams.delete('baking_bread');
+						url.searchParams.delete('baking_brownies');
+						url.searchParams.delete('baking_brownies_timeout');
+						e = url.href;
```
kasper93: From my testing this is indeed enough to bypass the preroll video ads. Not sure about ads pushed by the streamer during playback tho. But those can be dealt with later if needed.

filters/resources.txt
pcworld: I want to prevent patching the same object multiple times. I suppose `ourMediaPlayer === newMediaPlayer` is better. But I'll wait with further commits until I know whether the maintainers are in general fine with merging this PR.

gorhill:
Sorry for the delay, I will try to find the time to pull this today.

filters/resources.txt
```diff
@@ -2310,3 +2310,34 @@ damoh-defuser.js application/javascript
...
+					}
+				} catch (err) {
+					console.error('Failed to bypass Twitch livestream ad');
```
jspenguin2017: I suggest stating where the log came from: 
```
console.error('uBO > Failed to bypass Twitch livestream ad');
```

filters/resources.txt
jspenguin2017: This is a style suggestion, it's ultimately up to you and gorhill: 

`typeof` is not a function, it is a keyword, to stay consistent with the style of other scriptlets, it should be: 
```
if (typeof ourMediaPlayer !== 'undefined') { return; }
```

gorhill:
First thing I tried is to get ads on `twitch.tv`. So far, no such "luck". So I am going to need help testing this.

In any case, some points:
- Always [jshint](http://jshint.com/) scriptlets and fix whatever is reported.
- Always use the braces for block of code, even for single line
- I rather not have the `console.log` in there (commenting it out is fine once debugging is done)

RoyCurtis:
I've started testing on this. In my experience, ads have sometimes loaded mid-stream, so... I'm dedicating two hours to this. I'll edit this comment as I keep an eye on Twitch.

I ran the script through JSHint and only had to make one change. Else, there were no syntax errors and it loaded fine for me on Chrome and Firefox. Thank you very much, @pcworld !

# Test environment

* Windows 10 Pro 1703 desktop
* United Kingdom IP, British Telecom (BT) as ISP
* All testing done in incognito/private browsing mode, with uBlock allowed to run
* uBlock 1.16.0, [reset to default settings](https://i.imgur.com/65S5zDO.jpg) and [default filter lists](https://i.imgur.com/PWrTEjr.jpg)
* HTML5 player used on twitch.tv (by default)

## Test procedure

_Note: If I encounter a single ad, the test is considered a failure and rest of procedure abandoned_

1. Open incognito window
1. Navigate to https://www.twitch.tv/rifftrax (always-on stream with ads), watching for any pre-roll ads
1. Watch for 20 minutes, watching for any ad breaks
1. Hard refresh (CTRL+SHIFT+R) up to 10 times, each time watching for any pre-roll ads
1. Close window; repeat procedure up to 2 more times

## Patching procedure

1. [In uBlock settings, check "I am an advanced user" and click gears](https://i.imgur.com/gs18Vhk.jpg)
1. Change `userResourcesLocation` to `https://gist.githubusercontent.com/RoyCurtis/6e6dd016d9d127c6cc7a2c30f582782a/raw/4e26389eb1dc064c210b4810672241481e1312c8/resources.txt` and hit "Apply changes"
1. [In uBlock's "My filters", add the pcworld's new filter lines and hit "Apply changes"](https://i.imgur.com/dAtyZYP.jpg)

For this test, I made the following changes to my own copy of the patch:

* Changed `var ourMediaPlayer = undefined;` to `var ourMediaPlayer;`, as reported by JSHint
* Added braces for single statement `if` block, as gorhill requested
* Added a debugging `console.log` to line 25

# Chrome

* Version 66.0.3359.117 (Official Build) (64-bit)
* [Extensions installed](https://i.imgur.com/3P14A3V.jpg)

## ...with default settings and filters

[**Failure** ❌](https://streamable.com/fx3r4); ad shown immediately on stream load. This test was done for comparison sake, to make sure ads are actually getting past uBlock's defaults, without pcworld's patches.

## ...with pcworld's patches

In-progress...

- [x] [`twitch-videoads.js` loads successfully](https://i.imgur.com/qPXwFO5.jpg)
- [x] No pre-roll ads
- [ ] No ads after 20 minutes
- [ ] No ads after 10 hard refreshes
- [ ] No ads after repeating procedure two more times

# Firefox Developer Edition

* Version 60.0b14 (64-bit)
* [Addons installed](https://i.imgur.com/FXpbWdS.jpg)

## ...with default settings and filters

[**Failure** ❌](https://streamable.com/i59dm); ad shown immediately on stream load. This test was done for comparison sake, to make sure ads are actually getting past uBlock's defaults, without pcworld's patches.

## ...with pcworld's patches

In-progress...

- [x] [`twitch-videoads.js` loads successfully](https://i.imgur.com/HE7nYJP.jpg)
- [x] No pre-roll ads
- [ ] No ads after 20 minutes
- [ ] No ads after 10 hard refreshes
- [ ] No ads after repeating procedure two more times

gorhill:
Thanks.

TheEnhas:
This appears to no longer work on some streams... Twitch must have changed something once again.

https://www.twitch.tv/twitchpresents Commercial breaks in-between episodes are not blocked anymore. They were before during the Doctor Who marathon but currently (the Pokemon marathon) they're not. Pre-roll ads are still blocked however.

RoyCurtis:
I've noticed this today too, but haven't yet done a proper test. As before, I noticed ads appearing on https://www.twitch.tv/rifftrax . uBlock has been working since this PR's merge, up until today.

TheEnhas:
It's being complained about on Reddit, and other alternatives like the Twitch5 player have been completely broken to not work at all anymore (not just blocking ads, they won't even play the video). Something has changed just today.

In the past hour just from loading random streams a few pre-roll ads have slipped through as well. The injector script definitely no longer seems to be working. Also, sometimes when I first try to load a screen I get a "unless you have a time machine this content is unavailable" error and have to reload the page, after which it loads.

I have no idea what Twitch is doing.

Hrxn:
> https://www.twitch.tv/twitchpresents Commercial breaks in-between episodes are not blocked anymore. They were before during the Doctor Who marathon but currently (the Pokemon marathon) they're not. Pre-roll ads are still blocked however.

Can't reproduce right now, no commercial breaks in-between episodes for me here.
uBO on Chrome stable with default settings, pretty much..
One possible explanation: GeoIP, might depend on your current location..

verlain3:
Well, Twitch changed something indeed as i'm getting ads once again. It started a few days ago, but i could get away with it by just loading the bookmark again and the stream would load without showing the ad. Today though, it doesn't start any stream before showing me ads. The moment i pick a stream, it loads an ad every single time.

Psojed:
This week I started getting ads too. I get exactly two ad videos, both are Prime-related ads. I get them only when I first open up a stream and I get them on ANY channel I open, be it something big like the mentioned twitchpresents or a random smaller streamer. After I sit through these two ads, they don't show up anymore until the next day.
It seems unlikely to me that only two different ads would be displayed, so I assume the workaround in this topic is still working, but twitch added a new way to load these specific ads for prime.

gorhill:
We know ads are now getting through Twitch. Just as before, what is wanted is a pull request to solve the issue, just adding "me too" comments won't advance the issue and just create noise for those watching to the issue.

gorhill:
A new pull request is welcome whenever a solution is found.

PR #1981, opened 2018-04-15
Add recipe for Slack.

TheVirtualMachine:

`https://slack.com/get-started`
`https://*.slack.com` where the `*` is any Slack workspace.

### Describe the issue

Many pages on `slack.com` either fail to load or only partially load.

For any subdomains where an Slack instance resides, the page displays an error on how it failed to load. This is what is shown in the screenshot below.

### Screenshot(s)

screenshot

### Versions

- Browser/version: Tested on Firefox 59.0.2 (64-bit) on Fedora Linux, but issue likely exists on all browsers.
- uMatrix Origin version: 1.3.6

### Settings

- None that should effect `*.slack.com`.

### Notes

`_ slack-edge.com *` is necessary for Slack pages to load at all.
`_ slack-imgs.com image` and `_ slack-msgs.com *` are needed for a Slack workspace to fully load and be functional.


gorhill:
Your recipe did not work with my setup, there is no rule to unblock scripts from `a.slack-edge.com`. The documentation requires that `script` rules are needed if scripts are necessary on a site, because it is expected many will use a global `* * script block` in uMatrix.

TheVirtualMachine:
Those commits should fix the issue.

gorhill:
Thanks.

PR #1991, opened 2018-04-17
Added rule for Twitch as 3rd Party

jpdasma:
uMatrix rule for Twitch as 3rd Party


uBlock-user:
https://github.com/uBlockOrigin/uAssets/blob/master/recipes/recipes_en.txt#L149

jpdasma:
Hello, I think that one is for the domain twitch.tv. What I did is for the rules to watch twitch in other websites. (same as Youtube as 3rd Party)

jpdasma:
Hello, you could use this http://multitwitch.tv/beyondthesummit . This is a website that would allow to watch mutliple twitch stream. Note that there is a possible bug in uMatrix since it also shows the recipe for Twitch (the first party one). I think it's because it's parsing  twitch.tv inside multitwitch.tv (in the name)

jpdasma:
For clips, you can use this to test https://www.reddit.com/r/DotA2/comments/8da9db/713_offlane_in_a_nutshell/

gorhill:
> Note that there is a possible bug in uMatrix since it also shows the recipe for Twitch (the first party one).

Confirmed. Fixed in <https://github.com/gorhill/uMatrix/commit/6958c4ed26ab514cc1faf84d8f2ddf3be5f8cad5>.

PR #1997, opened 2018-04-17
uMatrix: Add reCaptcha recipe

ArchangeGabriel:
Note, I did explicitly use `www.` subdomains to avoid pulling in the unnecessary cookie from `google.com`. No worker allowed, in my experience even when there is one it is not necessary to allow it for reCaptcha to actually work.

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -78,7 +78,15 @@ Google Maps
 		_ maps.googleapis.com script
 		! needed for satellite view
 		no-workers: _ false
-		
+
+Google reCaptcha
+	* www.google.com
+		_ www.google.com *
+		_ www.google.com script
+		_ www.google.com frame
+		_ www.gstatic.com *
+		_ www.gstatic.com script
+
 IMDb
 	imdb.com *
 		_ 1st-party script


gorhill:
Do you have a test case? I would like to try it first.

ArchangeGabriel:
Yes, sure: http://www.bdmust.be/user, just below “CAPTCHA”. You don’t need to have an user account.

ArchangeGabriel:
First, you have a `www.google.com script` resource. Upon allowing, you now have a `www.gstatic.com script`. Then allowing this gives you a `www.google.com frame`. Which in turns makes `www.google.com xhr`. Didn’t look at what/when CSS and images are brought in.

You can also reproduce this behaviour on https://www.oneplus.com/account/sign-in, but you won’t see the reCAPTCHA before login. Maybe dummy IDs work though.

gorhill:
Ok, it worked fine. However:
- It needs a better name, "reCaptcha" alone is not enough. I suggest "Google reCaptcha".
- No need for the `_ www.google.com xhr` given the presence of `_ www.google.com *`, only frame/script should be assumed blocked.
- Use a space to separate terms, not a tab, the tab should be used just for indentation.

ArchangeGabriel:
Fixed, and moved to suit alphabetic order.

PR #2058, opened 2018-04-23
fix #2057

### LINKED ISSUE #2057: Issue with /services/analytics/* pattern

`Not released yet, issue discovered in QA`

### Describe the issue

We have some URLs containing <custom_path>/services/analytics/<custom_path>

### Screenshot(s)

services-analytics

### Versions

- Browser/version: Chrome and FF
- uBlock Origin version: 1.16.0

### Notes

Line to be removed: https://github.com/uBlockOrigin/uAssets/blob/15214cfdef6fb8166cbee25dcfad164f4154b83e/thirdparties/easylist-downloads.adblockplus.org/easyprivacy.txt#L2611

alphabetek:
Fix for #2057

### DIFF
--- a/thirdparties/easylist-downloads.adblockplus.org/easyprivacy.txt
+++ b/thirdparties/easylist-downloads.adblockplus.org/easyprivacy.txt
@@ -2608,7 +2608,6 @@
 /serve/mtrcs_
 /server.php?request=track&output=
 /server/detect.php?
-/services/analytics/*
 /services/counter/*
 /services/counters/*
 /services/pixel.html?


gorhill:
EasyPrivacy is not maintained here, this is merely a mirror, and used only at first install time, afterward the list is pulled from EasyList official URL.

alphabetek:
As I have mentioned issue has been discovered in QA so I have no working link to provide for now. We would like to avoid blocking users in production, thus fixing it now will be very beneficial.

gorhill:
> thus fixing it now will be very beneficial

You are removing a filter benefiting all users of EasyPrivacy (which is used by other top blockers) for a secret issue in the QA department of an unnamed organization (not that it would make a difference naming it). Can't you see how wrong that is?

Ask your QA people to add this filter in their instance of uBO:

    /services/analytics/*$badfilter

alphabetek:
> for a secret issue in the QA department of an unnamed organization
I think there is a misunderstanding, the issue will be faced by end users and has been discovered by QA people in QA stage. So it is not an internal issue, it is a preventive approach.

> Can't you see how wrong that is?
I don't see anything wrong about that, what is wrong IMHO, is denying a very generic pattern: /services/analytics/*, which could be used by any organization.

Thank you for the advice, I will contact EasyPrivacy team.

PR #2071, opened 2018-04-24
Add recipe for Reddit

MrTimscampi:
This adds a recipe allowing Reddit (reddit.com) to work properly (Up/down votes and embedded media).

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -140,6 +140,13 @@ Proton
 		_ 1st-party script
 		no-workers: _ false
 
+Reddit
+	reddit.com *
+		_ 1st-party script
+		_ v.redd.it xhr
+		_ v.redd.it media
+		_ redditstatic.com script
+
 Slack
 	slack.com *
 		_ 1st-party script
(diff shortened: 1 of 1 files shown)


zaphodbeebblebrox:
Things such as imgur images, imgur gifs, gfycat gifs, etc. are often embedded in reddit. Shouldn't they be part of the recipe as well?

MrTimscampi:
Indeed, I hadn't looked close enough at the requirements for recipes. I'll amend this asap.

gorhill:
> Things such as imgur images, imgur gifs, gfycat gifs, etc. are often embedded in reddit

They should be presented as separate 3rd-party recipes (ex. "Imgur as 3rd-party")

MrTimscampi:
Ah, well, then this should be good to go, as it enables all Reddit-hosted content to work.

gorhill:
Problem is that your ruleset recipe does not work for the strictest ruleset which must be supported: <https://github.com/gorhill/uMatrix/wiki/Ruleset-recipes#contributing-recipes>.

MrTimscampi:
@gorhill This one should be good now.

PR #2078, opened 2018-04-24
Add recipes for 4chan

MrTimscampi:

https://www.4chan.org/

### Describe the issue

Makes 4chan work (Quoted posts, posting, general browsing).
reCaptcha isn't detected on 4chan, so there are two recipes: with and without reCaptcha.

### Screenshot(s)

n/a

### Versions

- Browser/version: Firefox 59.0.2
- uBlock Origin version: 1.16.0

### Settings

n/a

### Notes

n/a


gorhill:
There is already a recipe for 3rd-party Google re-Captcha, there is no need to duplicate these rules in recipes of every single site using Google re-Captcha.

Issue #2107, opened 2018-04-27
The built-in Unbreak filter breaks Disqus comments on cyclingnews.com


biwqn:
### URL(s) where the issue occurs

Any article off of `http://www.cyclingnews.com/`.
For example:
`http://www.cyclingnews.com/races/tour-de-romandie-2018/stage-3/results`
`http://www.cyclingnews.com/features/giro-ditalia-chris-froome-defiant-despite-looming-salbutamol-case`
`http://www.cyclingnews.com/news/richie-porte-shows-improving-form-in-romandie-tt`

### Describe the issue

The built-in Unbreak filter actually breaks Disqus comments on cyclingnews.com. If you go to any of those 3 links above, and have the Unbreak filter enabled, Disqus comments at the very bottom of the article do not load. When the Unbreak filter is disabled, the Disqus comments successfully load.

This has been happening for many weeks, maybe months now, even with latest Ublock origin versions.

### Screenshot(s)

N/A, as very easy to see. Make sure to go to the very bottom of each article page to see the Disqus comments or lack thereof.

### Versions

- Browser/version: Google Chrome/66.0.3359.139 (Official Build)
- uBlock Origin version: 1.16.2

### Settings

- No default changes, easy mode.

### Notes

Turning off/on the Unbreak filter and loading those articles (shift-control/command-r to not use the browser cache).

okiehsch:
The filter breaking the comments was added to fix the navigation bar not responding to click events.
That has the unintended side effect of breaking the comment section I will remove that filter and add
`@@||google-analytics.com/analytics.js$script,domain=cyclingnews.com`
which will also fix the navigation bar.

geoffdutton:
#2108 Is a proposed solution to not whitelisting google-analytics.com. Adding `ga.loaded = true` will probably help in general anyway.

gorhill:
Meanwhile this can be tested with dev build and a dynamic filtering rule to block GA (this will override the exception filter).

PR #2108, opened 2018-04-28
Fix for #2107

geoffdutton:
#2107 
`http://www.cyclingnews.com/races/tour-de-romandie-2018/stage-3/results/`

### Issue and Findings
The disqus comments were being packed in a first party script that was being blocked. However, that included `MONEITIZER101` which pulled in "Daily Deals". This also re-broke navigation because it was checking if `ga.loaded === true` before forwarding to the intended URL. Finally, there was a 1 second delay before the forward would take place (to ensure the click was tracked by google analytics), so I sped that up.

### Notes
I added `ga.loaded = true` to the google-analytics.js redirect.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -392,6 +392,8 @@ google-analytics.com/analytics.js application/javascript
 		return [];
 	};
 	ga.remove = noopfn;
+	// https://github.com/uBlockOrigin/uAssets/issues/2107
+	ga.loaded = true;
 	w[gaName] = ga;
 	// https://github.com/gorhill/uBlock/issues/3075
 	var dl = w.dataLayer;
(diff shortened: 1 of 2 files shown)


geoffdutton:
I also added `cyclingnews.com##.global-banner` because it was extra whitespace at the top.

gorhill:
Shouldn't the blocking filter in "uBlock filters -- Unbreak" be removed instead of creating an exception?

I tried to find official documentation regarding `ga.loaded` and couldn't find any. However I found this article: <https://hacks.mozilla.org/2016/01/google-analytics-privacy-and-event-tracking/>. So apparently it's not officially documented but it is being used in examples in official documentation.

okiehsch:
I already removed the bocking filter.
https://github.com/uBlockOrigin/uAssets/commit/f0ef1109ec5cf6e1b99bf25676e14bef49520089

gorhill:
Ok. Now the next problem is that the change in the neutered GA scriptlet will be seen by all users only with the next release of uBO, because it's fetch from `./web_accessible_resources/`. This means I can pull the changes here, but we can't yet remove the GA exception filter until the modified GA scriptlet is widespread.

gorhill:
Thanks.

Issue #2144, opened 2018-04-30
tune.pk


KonoromiHimaries:
![2018-04-29-ust rukanime com 002 _ tune pk](https://user-images.githubusercontent.com/5884000/39407233-d0a6b148-4bc2-11e8-908e-036cf0e6da5e.png)
`https://tune.pk/video/7747748/ust-rukanimecom-002`

- video has stopped
- video has muted
- cookie alert

<details><summary>my settings</summary>

my filters
https://gist.github.com/KonoromiHimaries/c05bf7f6dd6dcb2d3beb090871985baa

 uBO 1.16.2
 ND 13.61
 Chr 66.0.3359.139
</details>

geoffdutton:
It seems like their servers are just slow. I don't see any cookie alert either. What specifically is broken?

KonoromiHimaries:
@geoffdutton i watch the anime and player has stopped and muted on this same time in this point
![2018-05-01-ust rukanime com 002 _ tune pk](https://user-images.githubusercontent.com/5884000/39461039-83f4b088-4d08-11e8-9afc-123bc18ca0f2.png)

geoffdutton:
Ah I see what you're saying now. I will try to take another look today.

geoffdutton:
Before I commit, can you try this: `tune.pk##script:inject(abort-on-property-read.js, TuneOverlay)`

ghajini:
@mapx- , when playing video, go towards(before) second yellow dots, at 2nd yellow dot, player will autopause......

mapx-:
I tried, I have only 1 dot (it seems to be in the second position), after 8-9 seconds is starting again

PR #2148, opened 2018-05-01
Fix for #2144

geoffdutton:


geoffdutton:
Nevermind, this doesn't work. I just had run out of ads to be shown.

PR #2164, opened 2018-05-02
Nukes skype from the internet

pedrogarcia:
Microsoft is forcing skype integration in all of its services. Skype integration makes pages slow, and it is very annoying. This change nukes skype from the internet.


mapx-:
1. It could be eventually seen as an annoyance
2. provide example pages where we can see such annoyance, resources used to load skype etc

PR #2189, opened 2018-05-05
Fixing gelbooru.com; manipulation in filters

konecat:
Exceptions have been added for no reason. Now removed. Gelbooru admin most likely fiddles with filters. Do not allow changes to gelbooru section without reviewing.


kowith337:
`gelbooru.com/ads.js` is needed'to allow because the bottom ads near page navigations will re-appear if the `abvertDar` cannot read exactly as `readOurTOS`, we once blocked it because they only check that value not to be blank (`undefined`) and inject constant value instead.

PR #2386, opened 2018-05-27
Moved filter (d58b17e)

smed79:


smed79:
Ok but i still think this is not the right place for these filters.

okiehsch:
Well the sie was broken because of the filter
`||adsnative.com^$third-party`
using an exception lead to ads, so the better fix to "unbreak" the site was to add
`||v.embed-cdn.com/v8/player.js$script,domain=streamable.com`

Why do you thnk it should be part of `uBO-filters`?

smed79:
It's rather an anti-adblock mechanism for me.

what are we going to do if all sites add a dependency to adservers to load the contents?

okiehsch:
I can see your point, but I am not sure if this is an anti-adblock mechanism or simply a badly designed site.
I agree with you that a site that deliberately breaks itself for adblock users should be fixed in `uBO-filters`
I just think this is a fringe case.
If all

smed79:
Thank you for being kind to comment my point of view.

PR #2630, opened 2018-06-21
Patch window.name to empty string instead of null

gwarser:
Because `window.name` converts all assigned values to string representation.


mapx-:
@gorhill

gorhill:
Thanks.

FYI, I am not subscribed to notifications from uAssets, due to the sheer amount of activities in here, so you need to mention me if you want me to be notified.

PR #2767, opened 2018-07-04
M: Removed unnecessary filter (https://forums.lanik.us/viewtopic.php?f=91&t=41066)

smed79:
https://forums.lanik.us/viewtopic.php?f=91&t=41066


okiehsch:
Thx.

Issue #2664, opened 2018-06-25
NSFW porndig.com


KonoromiHimaries:
![2018-06-25](https://user-images.githubusercontent.com/5884000/41828343-02cfd19c-7835-11e8-9a19-ac5bd855785c.png)
`https://www.porndig.com/videos/99365/two-lesbians-wrestle-before-the-action.html`

broken site

Adblock Warning Removal List + Fanboy’s Annoyance List

<details><summary>my settings</summary>

my filters
https://gist.github.com/KonoromiHimaries/afab54402a6cc37a168c2e55a0ef6896

 uBO 1.16.10
 ND 14.58
 Chr 67.0.3396.87
</details>

okiehsch:
If I enable `Fanboy’s Annoyance List` the video plays but the picture is blurred. Disabling cosmetic filters fixes it on my end.

mapx-:
Even enabling Annoyance List still able to see the video.

okiehsch:
This is what I see with `Fanboy’s Annoyance List` enabled using Chrome.
![unbenannt](https://user-images.githubusercontent.com/16838044/41872767-333440dc-78c3-11e8-97be-492fbabf7b1d.JPG)

mapx-:
Weird, I tested in FF / chrome, no such issue for me.

KonoromiHimaries:
maybe a session of teamviewer?

mapx-:
However, @ryanbr should test this issue

okiehsch:
I found the issue. `Fanboy’s Annoyance List` has this filter `##.disclaimer_wrapper`.
This hides this disclaimer on my end.
![unbenannt](https://user-images.githubusercontent.com/16838044/41873327-f2ab3f96-78c4-11e8-898c-9f494e94ce3a.PNG)
I have to click Enter to see an "unblurred" video, once I have clicked Enter, the issue is fixed, even if I reenable `Fanboy’s Annoyance List`.

mapx-:
Maybe the disclaimer is enabled on regional / country basis.

KonoromiHimaries:
`porndig.com#@#.noadblock`

ryanbr:
Is porndig.com#@#.disclaimer_wrapper needed?

okiehsch:
On my end it still is needed.

Without the exception.
[...]

PR #2915, opened 2018-07-16
Update unbreak.txt

KonoromiHimaries:
https://github.com/uBlockOrigin/uAssets/issues/2664


zaphodbeebblebrox:
This was already fixed in the list that was causing the problem, so the filter is not needed.

KonoromiHimaries:
@zaphodbeebblebrox is still not working on my end

okiehsch:
@KonoromiHimaries I can't reproduce, even if I enable `Adblock Warning Removal List`.

In any case `uBO-unbreak` will only add filters if the breakage is caused by the default setup.

https://github.com/uBlockOrigin/uAssets/blob/11d6a02255ed21f15ffb57d3acf0bc6d51e249c9/filters/unbreak.txt#L4-L5

`Adblock Warning Removal List` is not enabled by default.

PR #3009, opened 2018-07-23
Stripe Recipe

anelder-stripe:
Stripe has three products (Stripe.js/v2, Stripe.js/v3, and Stripe Checkout, so I've written a recipe that hits all three.  These are hosted on JSFiddle and an exception was not written for their code.  You may need to include jquery or add an exception for scripts they host as well.

#### Stripe.js/v2
http://fiddle.jshell.net/yn56Lebv/2/show/

#### Stripe.js/v3
https://fiddle.jshell.net/ywain/8nobq41n/show/

#### Stripe Checkout
https://fiddle.jshell.net/ywain/1fvrtrdh/show/

### Describe the issue

Right now, uMatrix will break on any page that includes Stripe.  I wanted to write up a nice little recipe that will allow them to safely enable our scripts opportunistically.

### Notes

Please let me know if I did this wrong.  This is my first recipe!

fixes https://github.com/uBlockOrigin/uMatrix-issues/issues/32

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -188,11 +188,14 @@ Steam
 		
 Stripe as 3rd-party
 	* stripe.com
-		_ *.stripe.com *
+		_ stripe.com *
+		_ api.stripe.com *
 		_ api.stripe.com xhr
+		_ checkout.stripe.com *
 		_ checkout.stripe.com frame
 		_ checkout.stripe.com script
 		_ checkout.stripe.com xhr
+		_ js.stripe.com *
 		_ js.stripe.com frame
 		_ js.stripe.com script
 
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -189,21 +189,12 @@ Steam
 Stripe as 3rd-party
 	* stripe.com
 		_ stripe.com *
-		_ api.stripe.com *
 		_ api.stripe.com xhr
-		_ checkout.stripe.com *
 		_ checkout.stripe.com frame
 		_ checkout.stripe.com script
 		_ checkout.stripe.com xhr
-		_ js.stripe.com *
 		_ js.stripe.com frame
 		_ js.stripe.com script
-		_ q.stripe.com *
-		_ q.stripe.com image
-		_ stripe.network *
-		_ stripe.network image
-		_ m.stripe.network *
-		_ m.stripe.network frame
 
 Twitch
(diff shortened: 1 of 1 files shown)


ArchangeGabriel:
You are enabling too much things to get it working. For instance, q.stripe.com is used only for tracking/telemetry as it seems, so uMatrix users won’t enable this.

ArchangeGabriel:
I’ll try to verify the minimal working setup in each case.

gwarser:
No need for `image` and `xhr` rules if you already allow `*`.

Or maybe no need for `*` if you only want `images`/`xhr`, yes @gorhill ?

anelder-stripe:
@gwarser - Yeah, I thought that was weird too.  I sorta just tried to follow the precedent set by some of the other recipes in there.

@ArchangeGabriel - If you want to disable telemetry/analytics, just disable those three I mentioned explicitly.  Everything else is required to function.

anelder-stripe:
@ArchangeGabriel - Do you want me to re-submit w/o the `*` entries?  I'm not sure what to do there.  It should work without them.

gorhill:
When many subdomains of a base domain are needed, I consider it's best to just work at base domain level. `q.stripe.com` si blacklisted in one of the hosts file, so this is not a concern. Given this and given the guidelines in the doc, the recipe can be simplified as follow:

    Stripe as 3rd-party
    	* stripe.com
    		_ stripe.com *
    		_ stripe.com script
    		_ stripe.com frame

anelder-stripe:
@ArchangeGabriel - I updated it to remove any telemetry collection and removed all but one of the blanket `*` entries.  Is that better?

anelder-stripe:
@ArchangeGabriel + @gorhill - Can this be merged in or would you like me to take @gorhill's approach above?  I'm happy to do either.

ArchangeGabriel:
@anelder-stripe Sorry for the delay in answering. This is way better, and per the current policy, you can also remove `xhr` rules since they are implied by the `_ stripe.com *` one. ;)

ArchangeGabriel:
@gorhill Now that they are only two subdomains remaining, I think there is no need to simplify even more.

I for one always prefer more granularity than plain whitelisting (as I would even prefer rules to allow the strict minimum, but I understand this might not be desirable), but I think here there is a good compromise. ;)

gorhill:
Merged, thanks for contributing.

PR #3048, opened 2018-07-27
https://forums.lanik.us/viewtopic.php?p=139108#p139108

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=40054


okiehsch:
That won't work you have to change it to 
`tralhasvarias.*##script:inject(bab-defuser.js)`.

okiehsch:
Thx.

Issue #2593, opened 2018-06-17
[Anti-adblock] visioncloud.rocks


jspenguin2017:
### URL(s) where the issue occurs

`https://visioncloud.rocks/home`

### Describe the issue

Redirects to anti-adblock page

### Screenshot(s)

![image](https://user-images.githubusercontent.com/7283682/41503457-579aa804-7192-11e8-8d5f-bc67597ff768.png)

### Versions

- Browser/version: Chrome 69
- uBlock Origin version: 1.16.10

### Settings

English default + `spa, por: Adguard Spanish/Portuguese`

PR #3049, opened 2018-07-27
Update #2593

smed79:


okiehsch:
image

Already fixed in the regional list.

smed79:
Ok, thank you.

PR #3137, opened 2018-08-05
Mapbox as 3rd party

theel0ja:
Demo:
* Mapbox.js, raster tiles: https://theel0ja.info/mapbox.js-fix-logo-demo/before.html
* Mapbox GL JS, vector tiles: https://rawgit.com/theel0ja/26e78e4644a3b644abe504abbf2360bb/raw/d9822702cd13672c59bfb21270045c9d62fd1616/mapboxgl-demo.html
* https://map.snapchat.com/

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -115,6 +115,13 @@ JSFiddle
 		_ jshell.net *
 		_ jshell.net script
 		_ jshell.net frame
+		
+Mapbox as 3rd-party
+	* mapbox.com
+		_ api.mapbox.com script allow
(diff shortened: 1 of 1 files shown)


theel0ja:
BTW, Mapbox's map library Mapbox GL breaks if web workers are disabled.

gorhill:
Again, `_ api.mapbox.com xhr allow` should be `_ api.mapbox.com *` since `script` is also needed.

`allow` keyword is implicit, no need to declare it.

There is a blank line before the `no-workers` rule, the recipe will not parse properly.

PR #3245, opened 2018-08-16
4chan

theel0ja:


gorhill:
`_ 4cdn.org xhr` is not necessary with `_ 4cdn.org *`.

gorhill:
Thanks.

PR #3250, opened 2018-08-17
Delete duplicate filter entry

Flaxseeds:


okiehsch:
This is not a duplicate.

Add
`||github.com^`
go to `github.com`, allow temporarly and look at your logger.
image

As you can see the inline-script is not blocked by `example.com^`

Flaxseeds:
Understood. The [Adblock Plus filter redundancy checker](https://adblockplus.org/redundancy_check) seems to report pairs of filters as redundant when they actually aren't. Do you know of something that works better? Anyway, sorry for the insanely long delay.

kulfoon:
This one is better https://arestwo.org/famlam/redundantRuleChecker.html

PR #3378, opened 2018-08-31
Embedly as 3rd-party

theel0ja:
Demo page: `https://rawgit.com/theel0ja/f68348ec5f8982de962a29571c2ef77d/raw/embedly-demo.html`


gorhill:
Best to use

    _ cdn.embedly.com * allow

Instead of

    _ cdn.embedly.com css allow

The idea is that only one rule will be needed for whoever does not explicitly block script.

theel0ja:
Done.

gorhill:
Thanks.

PR #3379, opened 2018-08-31
Cloudflare Stream as 3rd-party

theel0ja:
`https://www.cloudflare.com/` -> "Learn how Cloudflare works" video uses Cloudflare Stream and this ruleset fixes it completely.


gorhill:
Thanks.

PR #3384, opened 2018-09-01
Added Tor entries to the main list

DandelionSprout:
As I have recently begun to experiment with .onion domains in Tor Browser (e.g. for the purposes of https://github.com/DandelionSprout/adfilt/blob/master/BrowseWebsitesWithoutLoggingIn.txt), I've learned that filter entries that apply to e.g. Facebook and NY Times, do not automatically apply to their .onion domains.

Since the .onion versions of websites use the same elements as their regular versions, it should usually be easy enough to e.g. convert these examples that I took from the main uAssets list:

```
facebook.com##.userContentWrapper:has(a[href*="/ads/"])
nytimes.com##:xpath(//div[starts-with(@id, "dfp-ad")]/../../..)
```

...to:

```
facebook.com,facebookcorewwwi.onion##.userContentWrapper:has(a[href*="/ads/"])
nytimes.com,nytimes3xbfgragh.onion##:xpath(//div[starts-with(@id, "dfp-ad")]/../../..)
```

I've added .onion domains to those entries that I could find, and I suggest that it could become some sort of policy to add .onion domains to any future additions to the uAssets lists.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -645,22 +645,22 @@ eurogamer.net##html:style(cursor: default !important;)
 ||bit.ly^$popup,domain=eurogamer.net
 
 ! https://github.com/uBlockOrigin/uAssets/issues/98
-facebook.com###stream_pagelet div[id^="hyperfeed_story_id_"]:has(a.uiStreamSponsoredLink)
+facebook.com,facebookcorewwwi.onion###stream_pagelet div[id^="hyperfeed_story_id_"]:has(a.uiStreamSponsoredLink)
 ! "People You May Know": EasyList tries to block these, might as well block them fully
-facebook.com###stream_pagelet div[id^="hyperfeed_story_id_"]:if(h6:has-text(People You May Know))
+facebook.com,facebookcorewwwi.onion###stream_pagelet div[id^="hyperfeed_story_id_"]:if(h6:has-text(People You May Know))
 m.facebook.com##article:has(footer > div > div > a[href^="/friends/center/?fb_ref="])
 ! https://www.reddit.com/r/uBlockOrigin/comments/58o3k6/facebook_ads_solution/
-facebook.com##.ego_section:has(a.adsCategoryTitleLink)
+facebook.com,facebookcorewwwi.onion##.ego_section:has(a.adsCategoryTitleLink)
 ! https://github.com/uBlockOrigin/uAssets/issues/507
-facebook.com###stream_pagelet [id^="hyperfeed_story_id_"]:has(span._4dcu)
+facebook.com,facebookcorewwwi.onion###stream_pagelet [id^="hyperfeed_story_id_"]:has(span._4dcu)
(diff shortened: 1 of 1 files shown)


ZaphodBeebblebrox:
@gorhill, thoughts?

gorhill:
We can add reported cases but I wouldn't want volunteers to be burden of the task of hunting for `onion` alternative of web sites. However is free to report such cases though, and it will be taken care of as any other filter issues.

DandelionSprout:
There's not all that many major websites that have .onion alternatives, let alone ones that work properly. Off the top of my head there's Facebook, NY Times, DuckDuckGo, The Pirate Bay, possibly Wikipedia, and virtually no one else.

gorhill:
What would be good is to provide links to these URLs such that we can load and see the results of the modified filters.

DandelionSprout:
Ah, okay. Here's the ones I can confirm first-hand. 🙂 

Facebook: https://facebookcorewwwi.onion (Example page: https://en-gb.facebookcorewwwi.onion/Sonic)
NY Times: https://www.nytimes3xbfgragh.onion (Example article: https://www.nytimes3xbfgragh.onion/2018/08/31/sports/colin-kaepernick.html?action=click&module=Editors%20Picks&pgtype=Homepage)
DuckDuckGo: https://3g2upl4pq6kufc4m.onion
The Pirate Bay: http://uj3wazyk5u4hnvtk.onion

The easiest way to test .onion domains for adblocking (as far as I am concerned), is to download Tor Browser and to install the newest `uBlock0.firefox.signed.xpi` file through the Extensions menu (It won't accept unsigned extension files).

ZaphodBeebblebrox:
Interestingly, `nytimes3xbfgragh.onion` redirects to `https://www.nytimes3xbfgragh.onion/interactive/2017/admin/100000006079518.embedded.html` for me. Can others reproduce, or does my installation of tor have issues?

gorhill:
> Can others reproduce

Me too, I just end up with a blank page.

So we shouldn't blindly add `onion` alternatives. Just like with any other filters, the case must be made that an `onion` version should be added.

DandelionSprout:
@ZaphodBeebblebrox I can confirm that it's the same on my end, though it wasn't like that 6 hours ago. I'll look into it more.

DandelionSprout:
Okay, so I have failed to figure out why nytimes3xbfgragh.onion redirects like. I can only presume that it's a server configuration issue at the NY Times, since the category pages (e.g. Sports) and the articles are unaffected.

And this was some bad timing for such an issue to occur. My experiences with Facebook over the past few weeks, are that its page elements are named identically on the regular and .onion versions, so I figured that it'd be easy to verify that the site experiences on regular and .onion versions are usually the same...

gorhill:
Good enough for me, I will trust you with Facebook (I won't bother creating an account just to test the filters), and for the NTY it's probably a temprary glitch -- at least I could see the filter working for the large ad placeholder on the front page before it redirects to the blank page.

PR #3481, opened 2018-09-15
Delete redundant filters in resource-abuse.txt

Flaxseeds:
"||coinhive.com^$third-party" at line 165 is a duplicate of the exact same filter at line 13. "||kiwifarms.net/js/Jawsh/xmr/xmr.min.js" at line 185 is redundant with "/xmr.min.js" at line 126.


okiehsch:
Thx.

PR #3517, opened 2018-09-17
Update for twitch ads #1971

pixeltris:
This removes the "unique_id" cookie (responsible for delivering twitch ads) before it is sent in the https://api.twitch.tv/api/channels/XXXXX/access_token request. This isn't the best fix as it hooks window.fetch but it works.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2345,10 +2345,10 @@ twitch-videoad.js application/javascript
 		}
 	});
 	var realFetch = window.fetch;
-	window.fetch = function() {		
-		if (JSON.stringify(arguments).includes("access_token"))
+	window.fetch = function(input, init) {
+		if (arguments.length >= 2 && typeof input === "string" && input.includes("/access_token"))
 		{
-			document.cookie = "unique_id=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=.twitch.tv;";
+			init.credentials = "omit";
 		}
 		return realFetch.apply(this, arguments);
 	}
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2349,6 +2349,9 @@ twitch-videoad.js application/javascript
 		if (arguments.length >= 2 && typeof input === "string" && input.includes("/access_token"))
 		{
 			init.credentials = "omit";
+			var url = new URL(arguments[0]);
+			url.searchParams.delete('player_type');
+			arguments[0] = url.href;
 		}
 		return realFetch.apply(this, arguments);
 	}
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2348,7 +2348,7 @@ twitch-videoad.js application/javascript
 	window.fetch = function(input, init) {
 		if (arguments.length >= 2 && typeof input === "string" && input.includes("/access_token"))
 		{
-			init.credentials = "omit";
+			//init.credentials = "omit";
 			var url = new URL(arguments[0]);
 			url.searchParams.delete('player_type');
 			arguments[0] = url.href;
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2319,6 +2319,7 @@ damoh-defuser.js application/javascript
 
 twitch-videoad.js application/javascript
 (function() {
+	if ( /(^|\.)twitch\.tv$/.test(document.location.hostname) === false ) { return; }
 	var ourMediaPlayer;
 	Object.defineProperty(window, 'MediaPlayer', {
 		set: function(newMediaPlayer) {
@@ -2348,13 +2349,12 @@ twitch-videoad.js application/javascript
 	window.fetch = function(input, init) {
 		if (arguments.length >= 2 && typeof input === "string" && input.includes("/access_token"))
 		{
-			//init.credentials = "omit";
 			var url = new URL(arguments[0]);
 			url.searchParams.delete('player_type');
 			arguments[0] = url.href;
 		}
 		return realFetch.apply(this, arguments);
-	}
+	};
 })();
 
 
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2344,6 +2344,14 @@ twitch-videoad.js application/javascript
 			return ourMediaPlayer;
 		}
 	});
+	var realFetch = window.fetch;
+	window.fetch = function() {		
+		if (JSON.stringify(arguments).includes("access_token"))
+		{
+			document.cookie = "unique_id=;expires=Thu, 01 Jan 1970 00:00:00 GMT;domain=.twitch.tv;";
+		}
+		return realFetch.apply(this, arguments);
+	}
 })();
 
 


ZaphodBeebblebrox:
@gorhill

mapx-:
@gorhill

filters/resources.txt
```diff
@@ -2344,6 +2344,14 @@ twitch-videoad.js application/javascript
 			return ourMediaPlayer;
 		}
 	});
+	var realFetch = window.fetch;
+	window.fetch = function() {		
+		if (JSON.stringify(arguments).includes("access_token"))
```
pcworld: Maybe a little more robust and faster:
`if (typeof arguments[0] === "string" && arguments[0].includes("/access_token"))`

pcworld:
I've tested this roughly and confirm it removes preroll ads. I have also tried removing the old `MediaPlayer.MediaPlayer.prototype.load` patch and preroll ads are still gone without it, but I did not test this extensively and I guess keeping it doesn't hurt either (Twitch still uses the "baking brownies" parameters on requests).

pixeltris:
Actually this doesn't work since the js is fighting to set the unique_id in various places in the code and therefore is very timing related. Opening streams in new tabs makes ads almost always appear. The alternative would be to replace getOrCreateDeviceID in https://static.twitchcdn.net/assets/core-45cc951fa1577878dbc1.js. Or create an actual web request hook such webRequest.onBeforeSendHeaders.addListener as but afaik uBlock doesn't allow for such control.

pcworld:
What about this? This prevents sending cookies altogether in the access_token request (which also removes preroll ads for me):
```js
var realFetch = window.fetch;
window.fetch = function(input, init) {
	if (arguments.length >= 2 && typeof input === "string" && input.includes("/access_token"))
	{
		init.credentials = "omit";
	}
	return realFetch.apply(this, arguments);
}
```

I suppose someone should test that this doesn't break functions related to the logged in user, like chat etc.

pixeltris:
For me it now appears that removing unique_id now has no impact? Can anyone else confirm? If so it would seem the only solution would be to do something similar to https://github.com/instance01/Twitch-HLS-AdBlock (which comes at a cost of an additional delay before the stream starts). Or if streamlink still somehow doesn't get served ads copy what they are doing with their requests.

pcworld:
> For me it now appears that removing unique_id now has no impact?

Yes I get preroll ads again. Not sure if they changed something though, given that apparently it took them months to fix the previous "baking brownies" workaround (I'm not even sure they fixed it on purpose). Btw it's possible that the `server_ads` variable in the response gives away whether ads will be injected. I'm pretty sure yesterday with the cookies workaround I got a `\"server_ads\":false` response (without any patch it was `true`), now I get `\"server_ads\":true` again with the credentials patch. With your new code @ a178f6be73720671b7fecc5714924cc10d891098 (removing `player_type`), it reverts to `\"server_ads\":false` again (lets see for how long).

For future reference this is the access_token response I get with a178f6be73720671b7fecc5714924cc10d891098 on https://www.twitch.tv/rifftrax (strings with <> were removed by me):
```
{"token":"{\"adblock\":true,\"authorization\":{\"forbidden\":false,\"reason\":\"\"},\"blackout_enabled\":false,\"channel\":\"rifftrax\",\"channel_id\":7832442,\"chansub\":{\"restricted_bitrates\":[\"archives\"],\"view_until\":1924905600},\"ci_gb\":false,\"geoblock_reason\":\"\",\"device_id\":\"<randomstring>\",\"expires\":1537310229,\"game\":\"Always On\",\"hide_ads\":false,\"https_required\":true,\"mature\":false,\"partner\":false,\"platform\":\"web\",\"player_type\":null,\"private\":{\"allowed_to_view\":true},\"privileged\":false,\"server_ads\":false,\"show_ads\":true,\"subscriber\":false,\"turbo\":false,\"user_id\":null,\"user_ip\":\"<myip>\",\"version\":2}","sig":"<hexstring>","mobile_restricted":false}
```

Note that additionally removing credentials from the access_token request *does* make a difference: `adblock` becomes `false` (i.e. it breaks their AdBlock detection) and `device_id` becomes `null`. I guess breaking their adblock detection is a good thing? On the other hand I even get `"adblock":true` in access_token with a new browser profile (even tried with a VPN)... No idea what they're doing there.

Off topic, is there any way to reload uBlock's userResourcesLocation without having to change its URL *and* disabling+enabling uBlock?

pixeltris:
I think they are actively making changes to how ads are being served at the moment so it's likely there will be more changes to come (hopefully with workarounds). I'm all for removing credentials if it doesn't have impact on chat and things like you mentioned. And yea making changes to the resources file is kind of annoying.

bogachenko:
https://i.imgur.com/GkqSYxE.png check

verlain3:
@pixeltris how does one test your changes? Need to put something into the "userResourcesLocation"?

pixeltris:
You can test it by going to the uBlock dashboard, checking "I am an advanced user", clicking the cog next to the checbox, then setting userResourcesLocation to https://raw.githubusercontent.com/pixeltris/uAssets/twitchads-update/filters/resources.txt

gorhill:
> then setting userResourcesLocation to `https://raw.githubusercontent.com/pixeltris/uAssets/twitchads-update/filters/resources.txt`

Best is to create a Gist for the one scriptlet which needs to be tested.

Juviax:
Adblock plus has implemented the fix from pixeltris, any reason ublock hasn't? Or is twitch bribing you too.

Juviax:
adblock plus implemented this nearly a week ago. It works fine.

mapx-:
> adblock plus implemented this nearly a week ago. It works fine.

details ? filters ?

gorhill:
> adblock plus implemented this nearly a week ago.

I don't see anything specific to ABP and `twitch.tv` in EasyList and neither in their own [ABP filters](https://github.com/abp-filters/abp-filters-anti-cv) list or their [snippets file](https://github.com/adblockplus/adblockpluscore/blob/master/lib/content/snippets.js). Unless you provide a source, it seems you are making that up.

gorhill:
Here is a Gist for the scriptlet: <https://gist.githubusercontent.com/gorhill/a47fe36d5f3da185f8c4d44d18ee022d/raw/911466e5d255a081c7126392186bf1a08ee85d4c/gistfile1.txt>

So try out, assign above URL to `userResourcesLocation` in [advanced settings](https://github.com/gorhill/uBlock/wiki/Advanced-settings#userresourceslocation).

Graeme-L:
> Also, need more feedback as to whether this works without any negative side effects -- so far not much feedback has occurred.

Been running it for several hours now visiting multiple streams to force situations where I would normally see ads. No pre-roll ads showing up and no mid-roll ads on streams that I know run them hourly.

Also not seeing any video hitches which has been a problem with other solutions I've tried.

TheEnhas:
No ads since testing the updated scriptlet.

verlain3:
I've been testing this since @gorhill posted the scriptlet two days ago and not a single ad was thrown at me on the multiple streams i've watched. Total stream playtime may be about 16 hours. Seems good so far, will keep testing it.

thinkpad4:
I to have tried the scriptlet and have not seen a twitch ad since

ZaphodBeebblebrox:
@gorhill, according to the people in here it seems to work well, time to add it?

gorhill:
I am waiting for the [changes](https://github.com/uBlockOrigin/uAssets/pull/3517#issuecomment-423740517) to be made before I merge the request.

pixeltris:
Oh jeez my bad, put the changes in now.

gorhill:
I will need to publish a new revision for Firefox, since the stable release of uBO on AMO is not allowed to update its `resources.txt` asset, it must use the packaged version at all time.

gwarser:
@gorhill can this be merged to dev builds?

gorhill:
The most recent versions of assets here is always merged with the dev build.

gwarser:
Ah! Ok, this was merged 8 days ago, but latest dev release was 10 days ago, that's why I don't have it yet.

Hrxn:
So what's the situation right now? It's included by default for the dev builds on Firefox and Chrome?
But not yet present in any stable/release (1.17.2 for Firefox, 1.17.0 for Chrome) versions?

gwarser:
Scriptlet is included in Firefox package in v1.17.2. In Chrome it auto-updates in 7 days interval and after ~10+ minutes in new installations (you can speed this up by forcing update in Dashboard -> Filter lists)

Hrxn:
Hm. Not a new installation on Chrome here, but it should be well past 7 days by now...

Is this still the correct file used here: `resources.txt`? In, for my example:
`C:\Users\Hrxn\AppData\Local\Google\Chrome\User Data\Default\Extensions\cjpalhdlnbpafiamejdnhcphjbkeiagm\1.17.0_0\assets\ublock\`?

gwarser:
No. This is original file extracted from installation package. You can view updated `resources.txt` using uBO asset-viewer: `chrome-extension://cjpalhdlnbpafiamejdnhcphjbkeiagm/asset-viewer.html?url=ublock-resources` (You can access this from uBO Dashboard, Filter lists, click any list and then change "url" parameter to `ublock-resources`.

krystian3w:
#5184

filters/resources.txt
TiagoGw: Thank god people like you exist, helping brainlet fucks like me <3

magnulen:
> This removes the "unique_id" cookie (responsible for delivering twitch ads) before it is sent in the https://api.twitch.tv/api/channels/XXXXX/access_token request. This isn't the best fix as it hooks window.fetch but it works.

Hey, seems to have stopped working today.. 😢

PR #3899, opened 2018-10-27
Adds missing scripts for Yahoo Mail

Miterion:

`mail.yahoo.com`

### Describe the issue
The webmailer only loads a few mails without these allowed scripts (currently only the first twelve mails are shown in my inbox)

### Versions

- Browser/version: Firefox 63.0 (64 Bit)
- uBlock Origin version: uMatrix 1.3.14

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -298,6 +298,7 @@ Yahoo Mail
 		_ 1st-party script
 		_ yimg.com image
 		_ yimg.com xhr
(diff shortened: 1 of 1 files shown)


ZaphodBeebblebrox:
@gorhill

gorhill:
You don't need to use the `allow` keyword, it is implicit.

Miterion:
@gorhill Fixed in 80fe2c7

PR #4004, opened 2018-11-02
Fix Youtube images when images aren't allowed

theel0ja:


gwarser:
The strictest ruleset for which recipes must un-break sites is:

- The global rule * 1st-party * allow is present
- The global rule * 1st-party frame allow is present
- All 3rd-party requests blocked by default (including images/css)
- All scripts blocked by default (including 1st-party scripts)
- All 3rd-party frames are blocked by default
- All stock hosts files are enabled

https://github.com/gorhill/uMatrix/wiki/Contributing-ruleset-recipes

gwarser:
Ouch, sorry:

>All 3rd-party requests blocked by default **(including images/css)**

Issue #4114, opened 2018-11-14
Google Analytics callback not being called


niklabh:
### URL(s) where the issue occurs

`https://www.cars45.com/`

### Describe the issue

According to google analytics calling ga with a function as the first argument will call that callback with tracker.
https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference

ga(function(tracker) {
  console.log(tracker); // Logs `undefined`.
});

uBlock is replacing ga with its own version `uAssets/filters/resources.txt#373` but callback is not called in its version.

### Versions

- Browser/version: [Chrome Version 70.0.3538.102 (Official Build) (64-bit)]
- uBlock Origin version: [uBlock origin 1.17.0]

### Notes

Please add callback behaviour to mocked ga as well.

mapx-:
@gorhill

gorhill:
What specifically is broken on `https://www.cars45.com/` that would be fixed with honoring the callback?

niklabh:
Confirm Booking not working as it waits for the callback to submit form.

gorhill:
> If it is called with a callback it is expected to be called with tracker

This is not that simple. If the neutered script starts to call a callback, it could very well lead to more breakage -- this is not something I can change lightly without evaluating how this affects sites overall. It's why I wanted to find out specifically about this one here, but your answer was to vague, I was never able to figure out the breakage.

niklabh:
The confirm booking button on cars45.com try to get google analytics clientId before submitting the form.

```
button.onclick = function () {
  window.ga(function (tracker) {
    formData.clientId = tracker.get('clientId')
    
    booking.create(formData)
  })
}

```

button was not working due to ga callback not being called.

mapx-:
> Now the issue is patched at cars45

closing, the PR will still stay opened

PR #4115, opened 2018-11-14
Calling callback if provided for google analytics mock

niklabh:

`https://www.cars45.com/`

### Describe the issue

According to google analytics calling ga with a function as the first argument will call that callback with tracker.
https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference

ga(function(tracker) {
  console.log(tracker); // Logs `undefined`.
});

Adding this behaviour to uBlock

### Versions

- Browser/version: [Chrome Version 70.0.3538.102 (Official Build) (64-bit)]
- uBlock Origin version: [uBlock origin 1.17.0]

### Notes
Added callback behaviour to mocked ga as well.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -376,6 +376,12 @@ google-analytics.com/analytics.js application/javascript
 			return;
 		}
 		var f = arguments[len-1];
+		if ( typeof f === 'function' ) {
+			try {
+				f(new Tracker());
+			} catch (ex) {
+			}
+		}
 		if ( typeof f !== 'object' || f === null || typeof f.hitCallback !== 'function' ) {
 			return;
 		}


niklabh:
issue https://github.com/uBlockOrigin/uAssets/issues/4114

mapx-:
@gorhill

gwarser:
@gorhill 

Reported here: https://www.reddit.com/r/uBlockOrigin/comments/j3jtaq/ublock_enabled_but_with_all_local_site_blocking/

Clicking on "Get started" button does nothing - should login(?).

`ga` is called with `handleButtonClick`:

image

Clumsy patched script is wokring https://gist.github.com/gwarser/ca69a02fa567a132896ee88b64523b67

gorhill:
Nice, thanks for the investigation. I see a reference about this in the [documentation](https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference):

> Invoking the ga() command queue function by passing it a function will schedule the passed function for execution at the next point in the queue.

I think the safest fix is to call asynchronously the passed function since this is how the real API behaves.

gorhill:
Err I didn't realize where I was commenting, it's what the pull request was about. Not sure why I was against it since the documentation is clear.

gwarser:
Yes, and no idea how it was working - my "patch" uses undeclared `f` variable.

PR #4121, opened 2018-11-15
fix #4120

### LINKED ISSUE #4120: Mother Jones ad block notice
https://www.motherjones.com/politics/2018/11/the-acting-attorney-general-helped-an-alleged-scam-company-hawk-bizarre-products/

### Describe the issue
Ad block notice, at least on first load. It looks like they set a cookie or something so it only happens once.

### Screenshot(s)
image

### Versions

- Browser/version: Chrome 70
- uBlock Origin version: 1.17.0

### Settings

- All default settings expect added Nano filters too

### Notes

Just need to unblock `ads.min.js`, which is a bait file.

geoffdutton:
Details are in #4120

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -13644,3 +13644,6 @@ bit-url.com##+js(noeval.js)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/4119
 andreadd.it##+js(abort-current-inline-script.js, document.getElementById, block)
+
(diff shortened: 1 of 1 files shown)


ZaphodBeebblebrox:
#4120

okiehsch:
Two things, just fyi

1. We consider such issues, i.e. you can click away the message and continue using the site, an annoyance and fix them in uBO-annoyances.
2. Your filter is not optimal, because it would allow all scripts with `ads.min.js` at the mentioned domain.
`@@||motherjones.com/wp-content/*/ads.min.js$script,first-party` is a preferable filter.

PR #4265, opened 2018-12-06
#4265 ebonus.gg

smed79:
https://forums.lanik.us/viewtopic.php?f=62&t=42245


okiehsch:
I can't reproduce, do you have to be logged in to reproduce?

smed79:
> do you have to be logged in to reproduce?

I think yes, I rely on the user test.

okiehsch:
Thx.

PR #4373, opened 2018-12-23
Prevent exception if eval called with non-string argument in bab-defuser

alhimik45:
From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval):
>If the argument of eval() is not a string, eval() returns the argument unchanged.

But `eval` wrapper in `bab-defuser` violates this contract: site's scripts get exceptions from `check` function if pass numbers or functions to eval:
```
Uncaught TypeError: s.indexOf is not a function
    at check (<anonymous>:23:57)
```

This pull request adds cheking of argument type.

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -768,7 +768,7 @@ bab-defuser.js application/javascript
 	};
 	var realEval = window.eval;
 	window.eval = function(a) {
-		if ( typeof a !== "string" || !check(a) ) {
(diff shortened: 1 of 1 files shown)


mapx-:
@gorhill

gorhill:
Thanks for catching this.

Could you just change the double-quotes to single-quotes to be consistent with the rest of the file?

alhimik45:
done

Issue #1870, opened 2018-04-02
idsly.bid [Anti-Adblock]


KiLL4FuNK:
URL(s) where the issue occurs

`https://idsly.bid/NLh51`

Describe the issue

Anti-Adblock

Versions

Browser/version: Chrome 65.0.3325.181
uBlock Origin version: 1.15.18

Settings

Default

zaphodbeebblebrox:
`@@||idsly.bid^$generichide`

okiehsch:
I am geo-blocked and if I access this site via proxy the captcha is broken, so fixing this is not straightforward on my end.
The site disables right click functionality and it also disables the shortcuts for opening the console or the source-code.
Does
````
idsly.bid##script:inject(abort-on-property-write.js, clickIE)
idsly.bid##script:inject(addEventListener-defuser.js, contextmenu, preventDefault)
idsly.bid##script:inject(addEventListener-defuser.js, keydown, disabledEvent)
idsly.bid##script:inject(nano-setInterval-booster.js)
idsly.bid##script:inject(setTimeout-defuser.js, getComputedStyle, 250)
````
fix all these issues without breaking anything?

zaphodbeebblebrox:
Yes.
`idsly.bid##script:inject(addEventListener-defuser.js, keydown, disabledEvent)` does not appear to be needed. I can right click and solve the captcha without it.

okiehsch:
Yes, I forgot them, if I add filters for a site I fix all ad-related issues, even if they can be fixed by "normal" filters.
Does
````
idsly.bid###link-view > a[href^="/go.php"]
idsly.bid##.col-md-3
````
work?

zaphodbeebblebrox:
Bottom one works, top one doesn't.  For the top you could add `idsly.bid##img[src="https://idsly.bid/download3.png"]`.

zaphodbeebblebrox:
Np. If you ever need someone with a U.S. IP, I'm willing to help.

SMed79:
> I am geo-blocked

https://forums.lanik.us/viewtopic.php?f=88&t=37860

PR #4389, opened 2018-12-25
Moved filters #1870 #3403 #3989 + minor change

smed79:


mapx-:
only the minor change ?

mapx-:
If you want prepare a commit with that minor change

PR #4589, opened 2019-01-10
Update "gamecopyworld.*"

smed79:
```
gamecopyworld.com
gamecopyworld.click
gamecopyworld.eu
```


okiehsch:
I don't think `gamecopyworld.com` has anything to do with `adfly`, so would it be not better to remove the
` gamecopyworld.click` entry?

okiehsch:
Thx.

PR #4736, opened 2019-01-26
Fix fanfiction.net CSS style filter missing semicolon

yourduskquibbles:
reported by uBO logger in firefox

capture

### DIFF
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -1535,7 +1535,7 @@ europixhd.*,topeuropix.net##+js(abort-current-inline-script.js, document.onconte
 zlomowisko.info##+js(abort-on-property-write.js, document.oncontextmenu)
 
 ! https://github.com/NanoAdblocker/NanoFilters/issues/260
-fanfiction.net##.storytextp:style(user-select: text !important)
+fanfiction.net##.storytextp:style(user-select: text !important;)
 
 ! https://github.com/NanoMeow/QuickReports/issues/590
 @@||xnxx.com^$generichide


yourduskquibbles:
Something weird about this. 

The logger found this error for me on firefox but not on chrome, can anyone test if that is the case for them as well?

If happens for others, I think it means the logger isn't catching all the same errors in chrome vs firefox.

yourduskquibbles:
@mapx- before merging were you able to test to see if you found the error on chrome logger?

mapx-:
yes, you're right, I get the error only in FF.
@gorhill

yourduskquibbles:
Regardless of browser suppoert for `user-select` shouldn't the filter before fixing `fanfiction.net##.storytextp:style(user-select: text !important)` have been found as an error in chrome and firefox since it didn't have the closing semi-colon?

gorhill:
The closing semi-colon is optional for the last declaration.

yourduskquibbles:
> 
> 
> The closing semi-colon is optional for the last declaration.

Interesting, if I'm interpreting that correctly 

uBO **would** properly parse and apply both `user-select` and `-webkit-user-select` in `example.com##*:style(user-select: text !important; -webkit-user-select: text !important)` 

but **wouldn't** parse and apply both `user-select` and `-webkit-user-select` in `example.com##*:style(user-select: text !important -webkit-user-select: text !important;)` ONLY `-webkit-user-select` based on where the semi-colon is missing from?

gorhill:
uBO does not parse CSS style directives, it just pass them to the browser and see if the browser accepted it ([code](https://github.com/gorhill/uBlock/blob/85f00622b4074dc836add661cb02b8458d1cdb93/src/js/static-ext-filtering.js#L654)). If the browser does not accept it, uBO reject the filter as invalid. So in the end it's the browser making the determination whether a CSS declaration filter is valid.

yourduskquibbles:
Thanks for the explanation.

gorhill:
Note that just adding a `;` as in your PR did not change anything on my side -- the filter is still rejected.

Issue #2275, opened 2018-05-14
intoupload.com [Anti-Adblock]


KiLL4FuNK:
URL(s) where the issue occurs

`https://intoupload.com/o3md2s0xt2a9`

Describe the issue

Anti-Adblock

![](https://k62.kn3.net/2/F/5/2/9/A/D7A.png)
Microsoft Windows 8.1 Version : Update 3 (64-bit)

![](https://k50.kn3.net/B/7/A/3/5/4/6AC.png)
Google Chrome Version : 65.0.3325.181 (Official Build) (64-bit)

![](https://k50.kn3.net/D/D/4/5/E/9/CDB.png) 
uBlock Origin Version: 1.16.4

Settings

Default

ghajini:
`https://intoupload.net/vaagml1ondl1`
popunder ads before download

ghajini:
antiadblock
wait a bit or on mouse movement adblock warning appears
`https://intoupload.net/gniw5u5d8gxi`

ghajini:
Bundled download 

`intoupload.net/sxoiw7vxhaqh`

Filters=
`intoupload.net##+js(ra, checked, #chkIsAdd)`

`intoupload.net##.checkbox`

mapx-:
Does not work for me.

first needing  `|https://$script,3p,domain=intoupload.net,badfilter`
If I use your remove-attr filter the download button will become unclickable (still needing to click `Download our offers` )

ghajini:
Work on my end opera browser/firefox android, ff desktop not tested yet

Without my custom filters=

`https://user-images.githubusercontent.com/20338483/76241735-71dbd600-625b-11ea-9cad-64821b5e54dd.png`


With custom filters=

`https://user-images.githubusercontent.com/20338483/76241772-84eea600-625b-11ea-8bd2-854475ebc2f0.png`

`https://user-images.githubusercontent.com/20338483/76241808-92a42b80-625b-11ea-978b-443f8631c04f.png`

ghajini:
`intoupload.net/kw2oxv46a6du`

Filter =

`intoupload.net##+js(acis, Math, zfgloaded)`

ghajini:
`https://intoupload.net/1bjxlg4jmsbr`

1. With all default settings ,can't download as simply clicking on it ,there is no download initiated

2. scriptlet filters needs update

fix=
adding below filters fixes issue on my end

`intoupload.net##+js(acis, $, preventDefault)`
[...]

PR #4746, opened 2019-01-28
HandyUserscripts/AntiAdware#53 (#2275)

smed79:


okiehsch:
Breaks the download button, whcih is the reason the filter was removed. https://github.com/uBlockOrigin/uAssets/commit/4cdb4b9d20a1404929a00a41f49ce27129b188da

okiehsch:
To reproduce go to `https://intoupload.net/suobit0rxks6` with your filter, and click the "Download Now" button after the timer reached zero.

smed79:
Ok, thank you.

PR #4961, opened 2019-02-20
Update Fingerprint2 stub

socram8888:

http://www.realstream.space/en/player/5c6d0b214cae2605560650/33/227/5c6d95a2d996b.801925/FCSevilla-SSLazio/960/540

### Describe the issue

Fingerprint2 has been updated and [its .get() method is now static](https://github.com/Valve/fingerprintjs2#usage). The web site in the example also requires the function to return something:
```javascript
<script>var cF = function(result) {location.replace('http://www.realstream.space/en/player/5c6d0b214cae2605560650/33/227/5c6dc08cac816/' + result);};Fingerprint2.get(function(components) {var h = Fingerprint2.x64hash128(components.map(function (pair) { return pair.value }).join(), 31);cF(h); });</script><br>&nbsp;&nbsp;&nbsp;Loading...</body></html>
```
If result is an empty string (thus the location is http://www.realstream.space/en/player/5c6d0b214cae2605560650/33/227/5c6dc08cac816/), the web site redirects to http://www.realstream.space/blocked instead of the expected stream.

### Versions

- Browser/version: Firefox 65.0.1
- uBlock Origin version: 1.18.4

### Settings

- N/A

### Notes

N/A

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -1719,10 +1719,15 @@ twitch-videoad.js application/javascript
 fingerprint2.js application/javascript
 (function() {
 	let fp2 = function(){};
+	let browserId = '';
+	for (let i = 0; i < 32; i++) {
+		browserId += '0123456789abcdef'[(Math.random() * 16) | 0];
+	}
+	fp2.get = function(cb) {
+		setTimeout(function() { cb(browserId, []); }, 1);
+	}
 	fp2.prototype = {
-		get: function(cb) {
-			setTimeout(function() { cb('', []); }, 1);
-		}
+		get: fp2.get
 	};
 	window.Fingerprint2 = fp2;
 })();


ghajini:
@gorhill

Yuki2718:
Another example with a hidden form: `https://the-japan-news.com/news/article/0006495058`

gorhill:
`filters/resources.txt` is no longer used by uBO, the scriptlets are located at <https://github.com/gorhill/uBlock/blob/master/assets/resources/scriptlets.js>.

For any example of sites which needs to be tested, I need to be given the filter to test, so I can reproduce the failure and can test that the change remove the failure.

Yuki2718:
> 
> 
> `filters/resources.txt` is no longer used by uBO, the scriptlets are located at https://github.com/gorhill/uBlock/blob/master/assets/resources/scriptlets.js.
> 
> For any example of sites which needs to be tested, I need to be given the filter to test, so I can reproduce the failure and can test that the change remove the failure.

Should I make a new issue? For now in short, `||the-japan-news.com/modules/js/lib/fgp/fingerprint2.js$script,redirect=fingerprint2.js` makes the site loading never ends.

gorhill:
> makes the site loading never ends

The site loads fine on my side, with or without the filter. I can't find `fingerprint2.js` in the logger.

llacb47:
@gorhill, on first navigation to the article, you are served the following.

```
<script src="/modules/js/lib/fgp/fingerprint2.js"></script>
<script>
    var fp = new Fingerprint2();
    fp.get(function(result, components) {
        if(typeof window.console !== "undefined") {
            document.cookie = 'fp=' + result;
            history.replaceState({}, "", "/news/article/0006495058");
            // location.href = "/news/article/0006495058?fp=" + result;
            document.getElementById("fp").value = result
            var f = document.forms["fpform"];
            f.method = "POST";
            f.submit();
            return true;
        }
        else {
            location.href = "/";
        }
    });
</script>
<form method="post" action="https://the-japan-news.com/news/article/0006495058"  name="fpform">
   <input type="hidden" name="fp" id="fp" value="" />
</form>
```

The fingerprint is then POSTed to the article URL and you are granted access. The fingerprint `fp` is stored as a cookie `fp` for later use (I believe).

See this screenshot from Chrome devtools.
image

gorhill:
Ok I could reproduce following @llacb47's advice to delete cookies, and after adding the `important` option to the suggested filter -- it turns out EasyPrivacy has an exception filter for that site. The suggested fix address the issue.

@socram8888 Do you want to make the change to  <https://github.com/gorhill/uBlock/blob/master/assets/resources/scriptlets.js>?

Yuki2718:
@llacb47 thx, appreciated!

PR #3133, opened 2018-08-05
Algolia as 3rd-party

theel0ja:
Recipe for Algolia.

Example sites that have Algolia:
* `https://cdnjs.com/`
* `https://preview.algolia.com/instantsearch/`

screen shot 2018-08-06 at 0 42 51

PR #5010, opened 2019-02-25
Unbreak Packagist.org

Cysioland:
The search on Packagist.org did not work, so I fixed it (or so I hope)


Cysioland:
Possibly will be rendered obsolete by #3133

PR #5031, opened 2019-03-01
Remove ppoi.org & projectpoi.com from the resource abuse list

nsuchy:
This pull request removes ppoi.org and projectpoi.com from the resource abuse list. Both names are now defunct and display a parked or domain for sale page and no longer need to be included. Thoughts?


okiehsch:
Thx.

PR #5036, opened 2019-03-01
Remove hatevery.info

nsuchy:
The domain hatevery.info is no longer registered and can be removed from the list.


mapx-:
Are we sure `lhnhelpouttab-current.min.js` is clean everywhere ?

nsuchy:
> Are we sure `lhnhelpouttab-current.min.js` is clean everywhere ?

I actually didn't notice that wasn't a comment, I added that part back.

mapx-:
thx

PR #5039, opened 2019-03-02
Add dnsleaktest.net to badware list

nsuchy:
The URL http://dnsleaktest.net/ is a look alike of http://dnsleaktest.com/ however instead of providing a website that looks up your current DNS Resolver it redirects you to malware pages trying to get you to install a fake adobe flash player update.


okiehsch:
Thx.

PR #5040, opened 2019-03-02
Add foxload.com to badware list

nsuchy:
This is a bit tricky, the website foxload.com targets German internet users through Bing Advertisements. It offers downloads for a lot of popular software applications though their motivations are unclear, the binaries have likely been tampered with or have adware bundled in the installer.  Could someone look at this further and merge if they think this qualifies as crapware?

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -165,3 +165,6 @@ upload4earn.org##+js(remove-attr.js, checked)
 
 ! dnsleaktest.net badware
 ||dnsleaktest.net^$document
+
+! foxload.com badware
+||foxload.com^$document


okiehsch:
The browsers you can download via that site are rather out of date.

"Foxload bietet Chromium in Zusammenarbeit mit dem Startseitenportal startxxl.com zum kostenlosen Download an. StartXXL ist das neue Portal für personalisierbare Startseiten. Foxload steht nicht in Verbindung mit Google. Die hier angebotene Version des Chromium Browser ist eine portable Edition von CrPortable."
Meaning that they offer the portable chromium version of `http://crportable.sourceforge.net`
which has last been updated `July 16, 2017`

The Firefox download is version `57.0.1`.

The site encourages users to also download the `StartXXL Add-on`, which I would consider crapware but they do not bundle it as far as I can see.
@gorhill what do you think should we add the the site or not?

gorhill:
Result of VirusTotal for Chromium exe:

b

For comparison, I didn't see such result with Chromium downloaded from `https://chromium.woolyss.com/download/en/`.

Result of VirusTotal for the plug-in they encourage users to install:

a

So I would say it should be added.

Also, these tests can be [run by anyone](https://www.virustotal.com/#/home/upload), so it would be nice if they are actually run by whoever is submitting such requests rather than asking us to investigate.

PR #5529, opened 2019-05-05
Also remove cookie for base domain when on 'www.' subdomain

addisr:
### Describe the issue

We started having a problem with [removing cookies](https://github.com/tomasko126/easylistczechandslovak/blob/d7ae01/filters_ublock.txt#L13) used for showing ads on sites serving their content under `www.` subdomain. Problem was that they were setting cookie domain without `www.` part (just `.example.com`) and `cookie-remover.js` was not deleting those cookies.

### Versions

- Browser/version: Firefox 66/Chrome 74
- uBlock Origin version: 1.18.16

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -2245,6 +2245,8 @@ cookie-remover.js application/javascript
 			let part1 = cookieName + '=';
 			let part2a = '; domain=' + document.location.hostname;
 			let part2b = '; domain=.' + document.location.hostname;
+			let domain = document.domain;
+			let part2c = domain && domain !== document.location.hostname ? '; domain=.' + domain : undefined;
 			let part3 = '; path=/';
 			let part4 = '; Max-Age=-1000; expires=Thu, 01 Jan 1970 00:00:00 GMT';
 			document.cookie = part1 + part4;
@@ -2253,6 +2255,9 @@ cookie-remover.js application/javascript
 			document.cookie = part1 + part3 + part4;
 			document.cookie = part1 + part2a + part3 + part4;
 			document.cookie = part1 + part2b + part3 + part4;
+			if ( part2c !== undefined ) {
+				document.cookie = part1 + part2c + part3 + part4;
+			}
 		});
 	};
 	removeCookie();


mapx-:
@gorhill

gorhill:
Can I have a specific URL where I can see such case?

addisr:
Yes, for example `idnes.cz` and [many more](https://github.com/tomasko126/easylistczechandslovak/blob/master/filters_ublock.txt#L13) as linked above.

gorhill:
I went to `https://www.idnes.cz/zpravy/zahranicni/velka-britanie-anglie-brexit-britove-pas-zadost.A190503_101433_zahranicni_jadv` and I am unable to see cookies matching `/^_?adb/`.

addisr:
Reload and you will get it (they set it with `beforeunload` event).

gorhill:
Reloading does not help. I browse the site and so far no instance of any cookie which matches `/^_?adb/.` Using a VPN for CZE does not allow to reproduce either.

addisr:
Did you enabled czech/slovak filter for this test? That would explain it because current rules also defuse `beforeunload` so its effectively blocked. The issues happens for users who previsously visited the site with adblock but without czech/slovak filter or when some race condition occurs and `beforeunload` is not defused.

So correct way to reproduce this is uBlock enabled (without czech/slovak filter), go to `idnes.cz`, reload (sets cookie), now you should see ads everywhere, enable czech/slovak filter, ads will still be visible because cookie is not deleted.

gorhill:
Ok I can reproduce without CZE list and with only `idnes.cz##+js(cookie-remover.js, /^_?adb/)`.

However I don't think hardcoding a specific subdomain is the way to go, because what if `www.com`?

Looking around I see that the base domain can be looked up using [`document.domain`](https://developer.mozilla.org/en-US/docs/Web/API/Document/domain), this is the better solution.

addisr:
Updated PR. Looks better?

gorhill:
You should check that `part2c` is different from `part2b` before using it.

addisr:
Hm, was testing `document.domain` on other sites and for some reason it sometimes contain `www.` part too. Does not look reliable.

gorhill:
Which sites?

addisr:
https://www.google.com/ for example.

gorhill:
Ok, the documentation does actually mentions that it is the domain from the current origin, so then the question is why `idnes.cz` is returned when the current origin is `www.idnes.cz`... But in any case, if this solves your current issue, we should go ahead with the fix, this is really a safe fix where we do not have to worry about setting a cookie in the wrong origin.

addisr:
It does fix the issue for now. I will make new PR if anything breaks again.

gorhill:
One more detai,: the documentation says:

> In Firefox, this property used to return `null` if the domain of the document could not be identified, although this has now changed as of Firefox 62

Since uBO is compatible with Firefox 55+, there should be a test against `null` before using `part2c`:

>     let domain = document.domain;
>     let part2c = domain && domain !== document.location.hostname ? '; domain=.' + domain : undefined;

Then:

>     if ( part2c !== undefined )

I suspect the page changes `document.domain` to `idnes.cz` (possibly to set the cookie), which is allowed as per documentation, as long as the effective domain does not end up being changed.

addisr:
:heavy_check_mark: done!

gorhill:
Thanks.

gwarser:
> because what if `www.com`?

I don't think cookies can be set on top level domain, that's why Public Suffix List exists.

gorhill:
Right. Still I don't feel that hard-coding a specific subdomain (`www`) was the way to go.

PR #5760, opened 2019-06-04
Block Google Docs logImpressions API

ealgase:
This pull request blocks the logImpressions API, an API that is used for tracking by Google Docs.

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -172,3 +172,6 @@ rediff.com##+js(remove-attr.js, onmousedown, [onmousedown^="return enc(this,'htt
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/bvlumg/oldredditcom_outbound_link_redirect/
 old.reddit.com##+js(remove-attr.js, data-outbound-url)
+
+! google docs tracking api
+||docs.google.com/*/logImpressions?$xmlhttprequest,domain=docs.google.com


fffelix-jan:
It causes the ad counter to increase like crazy when I use Google Docs! Should they get rid of this filter?

llacb47:
No

ealgase:
@fffelix-jan ideally, the netcheck.gif would be redirected to a local file, that's what I do for mine (so no requests that aren't necessary ever hit Google)

fffelix-jan:
Does this tracking affect my privacy a lot? Why do we have to block this?

okiehsch:
We have thousands of filters blocking tracking, we can't call a symposium on every filter to decide if it affects privacy a lot or only a little. That's why we simply block requests that do track unless it causes breakage
If you don't think a particular filter is necessary you can disable it by adding
`||example.com^$badfilter` to your filter list. https://github.com/gorhill/uBlock/wiki/Static-filter-syntax#badfilter

For the record I don't think all tracking is the same, i.e. bad,  like some privacy advocates do.

cd-CreepArghhh:
Why does it retry so many times? 2 hours of having a Google doc open blocked 1286 ads/other content.
screenshot of the blocked content

uBlock-user:
Those are not ads, those are network requests blocked, it retries because it fails, google configured it as such.

uBlock-user:
Those are not ads, those are network requests blocked, it retries because it fails or could be because it was set to sent specific data periodically and ends up getting blocked by uBO.

PR #5765, opened 2019-06-05
Update unbreak.txt

terrorist96:
`https://www.koat.com/amp/article/gun-rights-group-pushing-to-make-bernalillo-county-a-second-amendment-sanctuary/27081131` video broken. There's probably more sites affected.


okiehsch:
Thx.

Issue #3060, opened 2018-07-28
[badware] Bundled software on lots of inofficial sites


stonecrusher:
Inspired by
https://twitter.com/JusticeRage/status/1021815597972291591

All from the same scammer:

### URL(s) where the issue occurs
```
http://7zip.fr
http://appdownloads.co
http://audacity.es
http://audacity.fr
http://azureus.es
http://bluestacksdownloads.com
http://celestia.es
http://celestia.fr
http://clonezilla.es
http://clonezilla.fr
http://cloudwayapps.com
http://garagebandforpc.org
http://gimp.es
http://gparted.fr
http://greenshot.fr
http://handbrake.es
http://inkscape.es
http://inkscape.fr
http://keepass.com
http://keepass.fr
http://nc3354.nexylan.net
http://notepad2.com
http://paintnet.es
http://paintnet.fr
http://scribus.fr
http://stellarium.fr
http://thunderbird.es
http://unetbootin.net
http://unetbootin.org
```
### Describe the issue

The old trick:
- Grab a good domain
- Make it look official
- Offer a PUP downloadmanager instead of the clean installer

### Notes

I checked every single one manually with at least one download (Windows x64 installer) and it always points to the same downloadmanager (just with different filenames but same hash):
```
http://7zip.fr
http://appdownloads.co
http://azureus.es
http://bluestacksdownloads.com
http://celestia.es
http://celestia.fr
http://clonezilla.es
http://clonezilla.fr
http://garagebandforpc.org
http://gimp.es
http://gparted.fr
http://greenshot.fr
http://handbrake.es
http://inkscape.es
http://inkscape.fr
http://keepass.com
http://keepass.fr
http://notepad2.com
http://paintnet.es
http://paintnet.fr
http://scribus.fr
http://stellarium.fr
http://thunderbird.es
http://unetbootin.net
http://unetbootin.org

https://www.virustotal.com/#/file/a5616985e92ca7c1df3b132d2da2ef33c64f38ba2dca40445017037473d7d014/detection

===============
Remaining items from the list:

http://nc3354.nexylan.net
Nothing

http://cloudwayapps.com
Clothes shop, no downloads

http://audacity.es
http://audacity.fr
downloaded files look clean, but not latest version
```

Downloadserver mostly is `http://www.femmfa-gis.com/` so maybe block that too.
[...]

PR #5774, opened 2019-06-07
Added various look-alike domains for popular open source tools

vin01:
I observed that the past list for look-alike domains of popular tools like keepass, 7zip is missing some of the domains pointing to the same IP address i.e 185.46.229.39

https://greenshot.org/
https://open-office.fr/
https://songbird.es/
(skipping all the domains here)

### Notes

I have consciously updated the existing list instead of appending at bottom as it is the part of the same set of malicious domains. Please let me know if this shall be amended. The domains are obtained from passive DNS data which was also shared with [BSI-CERT](https://www.bsi.bund.de/DE/Themen/Cyber-Sicherheit/Aktivitaeten/CERT-Bund/certbund_node.html) [earlier](https://twitter.com/certbund/status/1127864403276091393).

Please review.

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -92,31 +92,86 @@ upload4earn.org##+js(remove-attr.js, checked)
 ! https://github.com/uBlockOrigin/uAssets/issues/3060
 ! https://www.bleepingcomputer.com/news/security/fake-websites-for-keepass-7zip-audacity-others-found-pushing-adware/
 ! https://www.virustotal.com/#/file/a5616985e92ca7c1df3b132d2da2ef33c64f38ba2dca40445017037473d7d014/detection
+! https://twitter.com/certbund/status/1127864403276091393
+||7zip.es^$document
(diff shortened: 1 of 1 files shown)


yourduskquibbles:
@gorhill Do you have any issue with adding these domains from [5774 PR](https://github.com/uBlockOrigin/uAssets/pull/5774) to [badware.txt](https://github.com/uBlockOrigin/uAssets/blob/master/filters/badware.txt) list? 

Related discussion from last year: #3060
Twitter link with more info (from [5774 PR](https://github.com/uBlockOrigin/uAssets/pull/5774)): https://twitter.com/certbund/status/1127864403276091393

gorhill:
No issue. Note that I was not part of the discussion in #3060, so I don't need to be involved in all decisions as to whether something needs to be added to badware, especially the case here is about obvious deception.

yourduskquibbles:
Thanks @vin01

PR #5851, opened 2019-06-19
Hide annoyances on medium based websites

GoGoris:
Hide sponsor topbar and popup asking you to subscribe to the newsletter

afbeelding
afbeelding


ealgase:
Thank you for this!

yourduskquibbles:
@GoGoris Can you show screenshot of what the topbar is and the overlay modal that you are hiding for reference back if the PR gets accepted?

GoGoris:
Done

yourduskquibbles:
I don't think `##.metabar>.u-tintBgColor,.u-borderTopLighter` is a good filter - it is the site header bar and provides a link back to the main landing page for the site.

Screenshot below
medium


A workaround would be to pin it in place with something like `:style(position: absolute !important;)` but I don't think that belongs in the uBO Annoyances list since it's changing the default site design behavior. That is a problem to be solved in a separate annoyances list that does CSS style modifications and users understand that is going to occur. 

@okiehsch what do you think?

GoGoris:
If that is the case I will remove the topbar and only keep the overlay rules.

okiehsch:
> A workaround would be to pin it in place with something like :style(position: absolute !important;) but I don't think that belongs in the uBO Annoyances list since it's changing the default site design behavior. 
...
@okiehsch what do you think?

I agree.

GoGoris:
OK I removed the filters for the topbar. Now the PR only contains the filter for the popup window asking you to subscribe.

ealgase:
You might want to undo that last commit.

okiehsch:
@yourduskquibbles I can't reproduce the "popup asking you to subscribe to the newsletter"
Merging the pull request is up to you.

yourduskquibbles:
I can reproduce after loading multiple articles

medium


But `medium.com,codeburst.io,blog.angularindepth.com,hackernoon.com##html:style(overflow-y: visible !important)` 

should be replaced with 

`medium.com,codeburst.io,blog.angularindepth.com,hackernoon.com##html.u-overflowHidden:style(overflow-y: scroll !important;)`

@GoGoris can you update the PR with the suggested filter above and then I will commit and also add additional medium based domains.

yourduskquibbles:
Thanks

PR #5853, opened 2019-06-20
remove hyperurl.co "sponsorship" ad/annoyance

ealgase:

http://hyperurl.co/d1zest

### Describe the issue

### Screenshot(s)

Without filter:
image
With filter:
image

### Notes

Arguably an ad, but seems more relevant to annoyances.

### DIFF
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -2337,3 +2337,6 @@ msn.com##p:style(-moz-user-select: text !important; -moz-user-drag: text !import
 
 ! glamourmagazine.co.uk anti adb annoyance
 glamourmagazine.co.uk##+js(set-constant.js, ads_not_blocked, true)
+
+! hyperurl.co iTunes sponsorship (e.g. http://hyperurl.co/d1zest)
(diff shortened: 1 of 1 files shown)


okiehsch:
That is an ad not an annoyance and the [readme](https://github.com/uBlockOrigin/uAssets/blob/master/README.md) applies.

"If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable."

ealgase:
> That is an ad not an annoyance and the [readme](https://github.com/uBlockOrigin/uAssets/blob/master/README.md) applies.
> 
> "If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable."

Yeah, I was wondering where this should go (it is on the edge IMO).  If only it was this easy to submit filters to EasyList.

PR #5876, opened 2019-06-25
Speed up verystream download waiting time

ealgase:
Verystream usually requires you to wait five seconds until downloading the file is possible.  This uses `nano-setTimeout-booster.js` to speed that up.


okiehsch:
Thx.

PR #5898, opened 2019-07-01
amp-geo

terrorist96:
Maybe we should consider whitelisting this across the board. Lots of sites need whitelisting. Who knows how many we are missing.

`https://tucson.com/news/local/california-urges-high-court-not-to-hear-arizona-s-challenge/article_2eda46be-c9fc-5e8d-8686-64e54693fb12.amp.html`


okiehsch:
Thx.

Issue #2280, opened 2018-05-14
egy.best


zaphodbeebblebrox:
### URL(s) where the issue occurs

`https://egy.best/movie/welad-rezk-2015/`

### Describe the issue

Anti-adblock (I have been informed that the red text is an anti-adblock message)

### Settings

- Default settings

### Notes

`@@||cdn.egy.best/static/global/js/adblock.js$script,first-party` perhaps?

smed79:
@mapx- mirror site `http://egybest.site/`

KiLL4FuNK:
download Anti-adblock

`https://egy.best/movie/heart-of-his-mother-2018/?ref=home-new`

smed79:
anti adblock `http://egybest.site/`

```
@@||cdn.egy.best/static/global/js/adblock.js$script,domain=egy.best|egybest.site
```
or
```
@@||cdn.egy.best/static/global/js/adblock.js$script
```

smed79:
egybest has been shutdown.

gwarser:
Still alive?

https://old.reddit.com/r/uBlockOrigin/comments/cecksg/httpsegybest_detect_ublock/

KiLL4FuNK:
anti adblock play video

`https://www.egy.best/movie/dark-phoenix-2019/?ref=home-new`

gwarser:
https://www.reddit.com/r/uBlockOrigin/comments/cwjlnf/the_new_egybest_for_arabictranslated_movies_is_a/

PR #5977, opened 2019-07-18
Restore 2023ac2

smed79:
https://github.com/uBlockOrigin/uAssets/issues/2280#issuecomment-512922459


okiehsch:
Are all of those filters still necessary?
Can you test the site with the filters I proposed at https://github.com/NanoAdblocker/NanoFilters/issues/373#issuecomment-512694410

PR #6140, opened 2019-08-20
update Stripe recipe

vrischmann:
Hi,

Stripe has used the domain `stripe.network` for some time in third party scripts. I've updated the recipe to allow it in addition to stripe.com


vrischmann:
I just rebased on master and added the line `_ hooks.stripe.com frame`

I add to enable it for Stripe to work on scaleway.com
scaleway_stripe

gorhill:
Thanks.

PR #6156, opened 2019-08-21
block medium.com "pardon the interruption" modal

samer:
### Describe the issue

Medium.com displays an annoying modal prompting you to signup after you've read a few articles.

The modal appears after you open a few articles (3, I think)

`https://medium.com/swlh/state-of-the-digital-nation-2020-venture-road-22de4377836`

### Screenshot(s)

pardon

### Notes

The filter is aggressive, but this was the only way to target the correct div. The class on the div is randomized in an effort combat adblocking. An alternative approach would be to target a particular string, but that is also prone to stop working the moment the copy inside the modal changes.


mapx-:
thx

Issue #6092, opened 2019-08-15
kachelmannwetter.com needs defuser scriptlet


TheRealCuran:
### URL(s) where the issue occurs

`https://kachelmannwetter.com/de/wetter/2950159-berlin`

### Describe the issue

When loading any page on `kachelmannwetter.com` with an adblocker (uBlock Origin and uMatrix in my case) enabled you will get redirected to `https://kachelmannwetter.com/de/site/adblockuser`. The reason for this lurks in `https://kachelmannwetter.com/js/function.js`, lines 160ff:
```javascript
    if ($('#paywall-account-type').attr('data-value') !== 'payaccount') {

        try {
            nx.cmd.push(function () {
                nx.trigger('rescan');
            });
        } catch(err) {};

        a457c035a6dd2ca7c69(function(ret, triggerText) {
            var adblocker = 'Unknown';

            if(ret) {
                adblocker = 'Enabled';
                var infocounter=0;
                $('.md').each(function() {
                    infocounter++;
                    $(this).css('background-color','#f7284b');
                    $(this).css('color','#ffffff');
                    $(this).css('height','100%');
                    $(this).css('font-size','13px');
                    $(this).removeAttr("data-nx-container");
                    if (!isGoogle()) {
                        $(this).html(adblock_txt);
                    }
                    else {
                        $(this).html(adblock_txt2);
                    }
                    if (infocounter>=2) {
                        return false;
                    }
                });
            }
            else {
                adblocker = 'Disabled';
            }

            if(typeof ga !=='undefined'){
                ga('send', 'event', 'Adblocker', adblocker, { nonInteraction: true});
            } else if(typeof _gaq !=='undefined'){
                _gaq.push(['_trackEvent','Adblocker',adblocker,undefined,undefined,true]);
            }

            console.log({adblocker: adblocker, trigger: triggerText, isAdblockInfoSite: $('#noadblock-info').attr('data-value') === 'true'});

            var urlpath = get_url_path();
            if (urlpath.length == 0) { urlpath = '/'; }
            if (adblocker === 'Enabled' && $('#stop-redirect').attr('data-value') !== 'true' && $('#noadblock-info').attr('data-value') !== 'true' && !isGoogle()) {
                if(location.href.indexOf('adpreview=true') !== -1)
                    location.href = get_url_path()+'/site/adblockuser?adpreview=true';
                else
                    location.href = get_url_path()+'/site/adblockuser';
            }
[...]

PR #6239, opened 2019-09-03
create filter to prevent blocker detection on kachelmannwetter

Busti:

`https://kachelmannwetter.com/de`

### Describe the issue

The Website redirects the request if an adblocker is detected by an obfuscated script. This PR sets a variable, which causes the scripts output to be ignored.

### Screenshot(s)

image

### Versions

- Browser/version: Google Chrome | 76.0.3809.132
- uBlock Origin version: v1.21.6

### Settings

- I am an experienced user

### Notes

Reverse engineered script which causes the redirection. An obfuscated snippet, which is likely a content-blocker-detector script calls an obfuscated function. However, the function checks for search-engine crawlers and, if it detects a crawler, ignores the output of the content-blocker-detector script.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -22346,3 +22346,5 @@ vidia.tv##+js(acis, $, undefined)
 ! https://github.com/uBlockOrigin/uAssets/issues/5435
 *$popunder,domain=multiup.eu|multiup.org,3p
 ||multinews.me^$all
+
+kachelmannwetter.*##+js(set-constant, isGoogle, trueFunc)


okiehsch:
I can't reproduce the redirection with a default setup.
Only filters for issues that can be reproduced with a default setup will be added to `uAssets`.

Busti:
Adguard added filters for it some days ago. My filters weren't updated properly.  
This method is different, so it might be relevant again in the future, but for now it is not needed anymore.

okiehsch:
The filters for that particular site have been changed numerous times lately, see https://github.com/uBlockOrigin/uAssets/issues/6092 
Your filter works on my end if I deliberately trigger the adblock detection, so it could be added in the future if needed. 👍

PR #6522, opened 2019-11-06
A: (webtor.io) uBlockOrigin/uAssets#6522

smed79:


okiehsch:
Can you specify what the issue is and how to reproduce?

mapx-:
> Can you specify what the issue is and how to reproduce?
> I don't see anything on my end that would be fixed by your filter.

at the top
image

okiehsch:
Thx, I don't know how I missed that when I first checked.

PR #6578, opened 2019-11-16
Update recipes_en.txt

Jan02:
google recaptcha requires www.googleapis.com xhr allow

### DIFF
--- a/recipes/recipes_en.txt
+++ b/recipes/recipes_en.txt
@@ -40,6 +40,10 @@ Cloudflare Stream as 3rd-party
 		_ embed.cloudflarestream.com script
 		_ videodelivery.net xhr
 
+Coub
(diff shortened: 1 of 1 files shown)


recipes/recipes_en.txt
```diff
@@ -40,6 +44,10 @@ Cloudflare Stream as 3rd-party
 		_ embed.cloudflarestream.com script
 		_ videodelivery.net xhr
 
+Coub
+	_ akamaihd.net * allow
+	_ coubsecure-s.akamaihd.net media allow
```
gwarser: - Invalid format - no condition specified.
- `allow` keyword is not needed.

uBlock-user:
Project archived, so closing.

PR #6696, opened 2019-12-18
Fix anti-adblock on aternos.org (Solution 1)

llacb47:
https://github.com/NanoMeow/QuickReports/issues/1643#issuecomment-566691583

This PR fixes anti-adblock on `aternos.org`. If this solution is not preferable, I have also created a PR (https://github.com/uBlockOrigin/uAssets/pull/6697) that fixes the anti-adblock by keeping the existing filters and whitelisting another script.

IMO this solution is preferable but it is easy for the website to break the cosmetic filter.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -4593,12 +4593,9 @@ themeslide.com##+js(nano-setInterval-booster.js, countdown, 1500)
 ! https://github.com/uBlockOrigin/uAssets/issues/1494
 ! https://github.com/NanoAdblocker/NanoFilters/issues/272
 ! https://github.com/NanoMeow/QuickReports/issues/25
-*$script,domain=aternos.org,3p,redirect-rule=noopjs
-@@||aternos.org^$generichide
-@@||amazon-adsystem.com/*/apstag.js$script,domain=aternos.org
-@@||snigelweb-com.videoplayerhub.com/videoloader.js$script,domain=aternos.org
-@@||static.h-bid.com^$css,script,domain=aternos.org
-@@*$xhr,3p,domain=aternos.org
+aternos.org##.ad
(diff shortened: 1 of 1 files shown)


filters/filters.txt
```diff
@@ -4593,12 +4593,9 @@ themeslide.com##+js(nano-setInterval-booster.js, countdown, 1500)
...
 ! https://github.com/NanoAdblocker/NanoFilters/issues/272
 ! https://github.com/NanoMeow/QuickReports/issues/25
-*$script,domain=aternos.org,3p,redirect-rule=noopjs
-@@||aternos.org^$generichide
-@@||amazon-adsystem.com/*/apstag.js$script,domain=aternos.org
-@@||snigelweb-com.videoplayerhub.com/videoloader.js$script,domain=aternos.org
-@@||static.h-bid.com^$css,script,domain=aternos.org
-@@*$xhr,3p,domain=aternos.org
+aternos.org##.ad
+aternos.org##+js(acis,$)
```
okiehsch: Can this be made more specific by adding a second argument?
The second argument can be any string of the inline script you want to abort.
https://github.com/gorhill/uBlock/wiki/Resources-Library#acisjs-

llacb47:
I'm hesitant to make `acis` specific to an inline-script containing a certain string since the site has broken our filters in the past. If they do it again, I think it should be reconsidered. There is only one inline-script that uses `$` anyway.

okiehsch:
Thx.

llacb47:
@okiehsch they broke the `acis` scriptlet. Can you change the second argument to `css` or remove it entirely? I am not able to make a PR right now.

llacb47:
Thanks.

PR #6699, opened 2019-12-18
Removed old whitelist for Discovery TV streaming sites

llacb47:
This PR removes an old whitelist for Discovery TV streaming sites that is no longer needed. 

If you want to test each site, see the test URLs in https://github.com/easylist/easylist/pull/4533


okiehsch:
No need to test the exception filters are redundant anyway seeing as the same filters are part of `EasyList`.

llacb47:
My Easylist PR has been merged FYI, so that whitelist is no longer present.

PR #6704, opened 2019-12-20
Avoid whitelisting all foxnews scripts, made more specific

ryanbr:
When just using unbreak, we found this whitelist is whitelisting too much on the site. Should be more specific. I adjusted to cover the "/ads.js" javascript.

This will allow Brave to block `https://global.fncstatic.com/static/isa/app/lib/google-funding-choices.js` while still using unbreak.txt


ryanbr:
Also for privacy reasons, its whitelisting some tracking scripts. not ideal

okiehsch:
Thx.

PR #6705, opened 2019-12-20
Fix anti-adblock on aternos.org (again)

llacb47:
Continuation of https://github.com/uBlockOrigin/uAssets/pull/6696


okiehsch:
Thx.

PR #6707, opened 2019-12-20
Allow block on aproxy on all platforms, and cover dronedj.com

ryanbr:
Applicable to all the related 9to5 sites, including `dronedj.com` and should also cover other platforms.

Similar commits here; `https://github.com/abp-filters/abp-filters-anti-cv/commit/9f16210`


okiehsch:
Thx.

PR #6757, opened 2019-12-28
[Badware] 1hos.cf

DandelionSprout:

Mostly `1hos.cf`.

### Describe the issue

Okay, so the situation is that *badmojr*, who is the lone maintainer of *1hosts*, which was slowly becoming a reasonably respected compilation list with more than 120 source lists and its own XDA forum thread, appears to have just randomly vanished off the face of the earth, allowing his then-current domain (`1hos.cf`) to lapse (leaving only a meager little GitHub mirror behind), which has now been snapped up by a domain parking seller thingie that I don't think have the best of intentions.

In about 90% of cases, clicking on any previously-legit link to that domain, e.g. `https://1hos.cf/complete/adblock.txt`, leads me to some fishy (though not outright malicious per se) parking place that looks like the below screenshot (when used with a Norwegian locale) when loading insecure sources is turned on.

In about 10% of cases, and which I have no idea how to specifically trigger, it insta-redirected me to `katie.tncrun.net`, and from there to the usual malware sites and such things.

### Screenshot(s)

image

### Versions

- Browser/version: Chrome 79.0.3945.88 64-bit for Windows
- uBlock Origin version: Nano Adblocker 1.0.0.130 + Nano Defender 15.0.0.170

### Settings

I admit to be *very* unsure as to how to best list every single extension change of any kind, so I'll just post a screenshot of my used lists as per usual: 
image

### Notes

The new *Windows Sandbox* tool seems to be pretty convenient to test out sites and links like these, as it makes me considerably less likely to insta-panic and close the browser before I've done enough research on such sites.

`1hosts.cf` seems to have been the first domain that the maintainer used. It too has lapsed, although it's not currently known to have been snapped up by anyone else.

I was unsure whether to use `||katie.tncrun.net^$all` or `||katie.tncrun.net^$document`. I'd go with the one that best prevents further redirections from happening, for which I've in previous cases had much luck with `$document`, but this time I got unsure if it'd block a super-obvious maldirection domain strongly enough.

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -273,7 +273,7 @@ majorgeeks.com##b:has(a[target^="reimage"])
 ||securemecca.com^$document
 ||hufilter.hu^$document
 ||fredfiber.no^$document
-! https://github.com/uBlockOrigin/uAssets/pull/[placeholder]
+! https://github.com/uBlockOrigin/uAssets/pull/6757
 ||1hos.cf^$document
 ||1hosts.cf^$document
 
@@ -446,5 +446,5 @@ howtoremove.guide###gray_de
 ! https://github.com/NanoMeow/QuickReports/issues/2522
 ||browser-update.*.top^
 
-! https://github.com/uBlockOrigin/uAssets/pull/[placeholder]
+! https://github.com/uBlockOrigin/uAssets/pull/6757
 ||katie.tncrun.net^$all
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -266,13 +266,16 @@ majorgeeks.com##b:has(a[target^="reimage"])
 ! https://github.com/uBlockOrigin/uAssets/issues/5187
 ||check-prizes-now2.life^$document
 
-! Lapsed domain that once hosted adblock lists, several of whom are now used for bad purposes
+! Lapsed domains that once hosted adblock lists, several of whom are now used for bad purposes
 ! https://github.com/uBlockOrigin/uAssets/issues/5307
 ||adblock.gjtech.net^$document
 ||spam404bl.com^$document
 ||securemecca.com^$document
 ||hufilter.hu^$document
 ||fredfiber.no^$document
+! https://github.com/uBlockOrigin/uAssets/pull/[placeholder]
+||1hos.cf^$document
+||1hosts.cf^$document
 
 ! Badware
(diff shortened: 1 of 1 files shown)


okiehsch:
Thx.

DandelionSprout:
You're welcome. 🙂

DRSDavidSoft:
It's sad to see @badmojr leaving the [1Hosts project](https://forum.xda-developers.com/android/general/badmojr-one-host-file-to-block-t3713360) behind like this, since it actually blocked many of the malware/adware domains that the other filters lists that I'd used didn't catch. Although IMHO this was kind of inevitable, when using [free `.cf` domains](https://www.freenom.com/).

hawkiee552:
I've noticed aliexpress.no redirecting to katie.tncrun.net, I got the warning form Ublock and had to investigate. It used to show the domain as parked, but now it redirects trough previously mentioned address and ends up on aliexpress.com.

DandelionSprout:
Checking out `http://aliexpress.no` in a hurry in *Windows Sandbox*, apparently it leads me through multiple redirection domains that are normally varying degrees of malicious (Some of them are usually *very* malicious), but which for once had the dignity to actually lead to `aliexpress.com` in the end, although it's very likely the redirections could be used for malignant tracking nevertheless.

Incomplete list of other such redirections that `aliexpress.no` led me to (*Not* to be blocked in uAssets until further notice, but could be added to independent lists):
```
ww2.siteplug.com (Usually the starting point
hhw.773.us
ww38.aliexpress.no
```

I strongly recommend that `tncrun.net` remain blocked as it currently stands, especially considering #6822. Alternately some very brave people out there with split-second reaction skills, could look into the exact `tncrun.net` URLs to see if the entries can be specified further.

llacb47:
Here's my list of domains I was redirected to - in order:

1. `aliexpress.no`
2. `park.above.com`
3. `forwrdnow.com`
4. either seeked.info or 7proof.com, which took me to rakuten.com (a page about aliexpress actually: `https://www.rakuten.com/signup_sem.htm?merchant_id=10644&src=msn&eeid=17881&utm_campaign=nb&pub=intg&utm_source=6423621368&utm_medium=intangoCTO&utm_content=intgo&utm_source=6423621368&utm_medium=intangoCTO&utm_content=intgo`

PR #6779, opened 2020-01-04
Fix popups on ebooksz.net

ryanbr:
Visiting: `http://www.ebooksz.net/2019/12/27/essential-mathematics-for-economics-and-business-2013/`

This will block the popups :)


okiehsch:
Thx.

PR #6830, opened 2020-01-15
Update syntax of rare cases

JustOff:
Update syntax of rare cases, as they were missed during the global search and replacement, and also are not detected by the converter embedded in firefox-legacy-1.16.4.14.


mapx-:
thx

PR #6930, opened 2020-02-06
Fix juragan-anime.net

nik2143:
This fix ads on juragan-anime.net


mapx-:
@Alex-302

`juragan-anime.net##+js(aopr, MANGO)`

Alex-302:
@mapx- does not URL blocking rule enough?

mapx-:
sure, just in case they change the domain ..

Alex-302:
Ok, will check later.

PR #6983, opened 2020-02-21

llacb47:
Last PR: https://github.com/uBlockOrigin/uAssets/pull/6949


Pokerkoffer:
This is needed for me as well: 
`aternos.org##+js(set, apstagLOADED, true)`

llacb47:
@pokerkoffer no, amazon-adsystem script is already whitelisted

Pokerkoffer:
> @Pokerkoffer no, amazon-adsystem script is already whitelisted

You are right, my bad. Privacy Badger blocked it.

PR #7013, opened 2020-02-27

llacb47:
Last PR: https://github.com/uBlockOrigin/uAssets/pull/6996

Note: the site added code to detect scriptlet injection, which will probably make things more difficult in the future.


okiehsch:
Thx.

PR #7029, opened 2020-02-29
Block Criteo tracking on tagesspiegel.de

ryanbr:
**Visiting:** `https://www.tagesspiegel.de/gesellschaft/panorama/coronavirus-weltweit-so-unterschiedlich-gehen-die-laender-mit-dem-virus-um/25595076.html`

### Describe the issue

We're blocking the following:

`<script type="text/javascript">`
`var Criteo = Criteo \|\| {};`
`Criteo.events = Criteo.events \|\| [];`
`Criteo.events.push(function() {`
`Criteo.DisplayAcceptableAdIfAdblocked({`
`"zoneid": 640113,`
`"width": 160,`
`"height": 600,`
`"containerid": "crt-sky-right",`
`"callbacksuccess": function() { adslotFilledByCriteo('sky-right', true); }`
`});`
`});`
`</script>`

### Versions

- Browser/version: Brave
- uBlock Origin version: 1.24.4

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -186,3 +186,7 @@ brillen.de##+js(acis, document.createElement, 'script')
 
 ! thesaurus.com / dictionary.com Adblock detection
 dictionary.com,thesaurus.com##+js(aopr, adDetection)
+
+! tagesspiegel.de Criteo tracking
+tagesspiegel.de##+js(acis, Criteo)
+


uBlock-user:
Criteo is already blocked, so what difference does this make ?

ryanbr:
Stops the inline scripts? Looking at the privacy aspect here.

uBlock-user:
> Stops the inline scripts?

Stopping or not stoping the inline-script doesn't stop the website from making a connection attempt to `static.criteo.net`, so all this filter does is stop the execution of this script.

ryanbr:
nvm. I'll push this to my own list.

PR #7037, opened 2020-03-01
Fix breakage on aternos.org

llacb47:
On certain pages, for example `https://aternos.org/software/v/snapshot/latest`, the ad placeholder covers buttons and impedes site functionality.


mapx-:
thx

PR #7047, opened 2020-03-02
Block ad elements on Bing

socram8888:

`https://www.bing.com/search?q=consoles`
`https://www.bing.com/search?q=travel`

### Describe the issue

Ads blocked by current visual filters are made visible again by a script in the page.

### Versions

- Browser/version: 73.0.1
- uBlock Origin version: 1.25.0

### Settings

No changes.

### Notes

The script in question removes the ".b_ad" class and uses CSS "::before" to insert the "Ad" label rather than a HTML element. Therefore the ads after processing are indistinguishable from a legit result by just looking at the HTML, and thus unblockable by means of CSS filters (ie from EasyList).

By removing the elements straight from the HTML the script finds no ads, and thus no elements are made visible again.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -18028,6 +18028,7 @@ bing.com###b_results > li[style]:not([class])
 bing.com###b_results:xpath(//span[(text()='d') or (text()='n')]):nth-ancestor(5)
 bing.com##.b_algo:has(p:matches-css-before(content: "Ad"))
 bing.com##+js(set, Blob, noopFunc)
+bing.com##^.b_ad
 
 ! https://github.com/NanoMeow/QuickReports/issues/1475
 ! https://github.com/NanoAdblocker/NanoFilters/issues/426


uBlock-user:
That will only serve Firefox users, `bing.com##.b_ad:style(position: absolute!important; left: -3000px!important;)` works on both.

socram8888:
> 
> 
> That will only serve Firefox users, `bing.com##.b_ad:style(position: absolute!important; left: -3000px!important;)` works on both.

nw

Not working here (note the four first results saying "Anuncio", Spanish for "Ad").

Again, the script removes the .b_ad class so filtering by CSS is not feasible. I have tried both with other filters enabled and with them disabled (to rule out some other ruleset inteferring) and neither worked.

socram8888:
I had also accidentally disabled the "My rules" switch so they weren't being applied lol. Your fix works perfectly fine.

uBlock-user:
I have tested on the English version of Bing.

socram8888:
It looks like it conflicts with EasyList (which is enabled by default). If I put the new rule manually in the "My filters" tab, this filter fails, and I get the ads back.

It works if the EasyList ruleset is disabled, though.

uBlock-user:
Some filters may need to be disabled --

```
bing.com#@#li.b_adBottom
bing.com#@#a[h$=",Ads"]
bing.com#@##b_results > li *:has(span:has-text(Ad))
bing.com#@##b_results > li:not(.b_algo):not(.b_ans):not(.b_pag):not(.aca_algo):not(.aca_algo_count):not(.b_msg)
```

mapx-:
all these ..
```
bing.com#@#.b_ad
bing.com#@##b_results > li:not(.b_algo):not(.b_ans):not(.b_pag):not(.aca_algo):not(.aca_algo_count):not(.b_msg)
bing.com#@#a[h$=",Ads"]
bing.com#@#li.b_adBottom
bing.com#@#.b_adLastChild
```

uBlock-user:
The ones I posted appear in my logger..

PR #7063, opened 2020-03-05
Add nowebsocketsend-if.js to resource library

smege1001:
This scriptlet is for blocking WebSocket messages containing tracking info made to a server that cannot be blocked, because it provides essential functions such as chatting, etc.


uBlock-user:
`resources.txt` is deprecated and no longer used. Do you a have test-case where this needs to be used to address the issue ?

smege1001:
vlare.tv uses WebSocket to send tracking info to the same server it uses for chatting, status change, etc.

PR #7100, opened 2020-03-15

llacb47:
Last PR: https://github.com/uBlockOrigin/uAssets/pull/7098


okiehsch:
Thx.

llacb47:
I came up with a better approach: https://github.com/uBlockOrigin/uAssets/pull/7101

PR #7108, opened 2020-03-16
Improved Facebook/Pornhub entry domain coverage even more

DandelionSprout:
I decided to attempt to improve the entries for `Pornhub` and `Facebook` furthermore (At least those that needed it; some entries in filters.txt had already been fixed beforehand so that they covered `pornhub.org` and `pornhub.net`), in connection to how I [was doing the same for EasyList](https://github.com/easylist/easylist/pull/5072).

I also noticed four entries in filters.txt that could be distilled into two `$all` entries, which I did.


okiehsch:
Thx.

PR #7205, opened 2020-04-07
Fix ads on boards.net sites

ryanbr:

Fixes ads on 
`https://xmenagerie.proboards.com/thread/1812/introduction`
`https://crossports.freeforums.net/thread/641/links-pl-message-boards`
`https://trongridlines.boards.net/thread/2/welcome-gridlines-programs`

### Describe the issue

https://forums.lanik.us/viewtopic.php?f=62&t=44538

### Screenshot(s)

### Versions

- Browser/version: `Brave/Version 1.5.123 Chromium: 80.0.3987.163 (Official Build) (64-bit)`
- uBlock Origin version: `1.24.4`

### Settings

### Notes

`<script>
            window.runAntiAdBlock = 1;
        </script>`

### DIFF
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -3711,6 +3711,9 @@ nonton78.com##+js(acis, onload, contextmenu)
 ! https://github.com/uBlockOrigin/uAssets/issues/661#issuecomment-609018989
 proboards.com##+js(aeld, scroll)
 
+! proboards https://github.com/uBlockOrigin/uAssets/pull/7205
+boards.net,freeforums.net,proboards.com##+js(set, runAntiAdBlock, 0)
+
 ! https://forums.lanik.us/viewtopic.php?p=153944#p153944
 deezer.com##+js(nostif, bait)
 deezer.com##.abp-banner
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -6897,6 +6897,7 @@ zoosintabues.com##+js(aopw, Fingerprint2)
 @@||proboards.com^$ghide
 /tortoise.min.js$domain=boards.net|freeforums.net|proboards.com
 boards.net,freeforums.net,proboards.com##+js(acis, $, vglnk)
+boards.net,freeforums.net,proboards.com##+js(set, window.runAntiAdBlock, 0)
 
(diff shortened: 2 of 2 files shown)


uBlock-user:
@ryanbr Why `window.runAntiAdBlock` instead of `runAntiAdBlock` ?

window.runAntiAdBlock is internally translated to `window.window.runAntiAdBlock`

ryanbr:
sure, updated

uBlock-user:
@gorhill isn't window.runAntiAdBlock translated to window.window.runAntiAdBlock anymore ? I'm not seeing any difference in Firefox, did something change with the scriptlet ?

uBlock-user:
With or without that filter, I still see ad placeholders which need 

`###banner-container, [id^="ad-"]` to fix.

gorhill:
> isn't window.runAntiAdBlock translated to window.window.runAntiAdBlock anymore ?

Yes -- there is no need to prefix with `window`.

`window` is a property of `window`, pointing to itself. See <https://developer.mozilla.org/en-US/docs/Web/API/Window/window>, so in the end even `window.window.window.runAntiAdBlock` would work.

uBlock-user:
Was this always like this since th beginning?, because I distinctly remember you telling someone on the tracker that using `window.runAntiAdBlock` will not work, a year or two ago

gorhill:
> I distinctly remember you telling someone on the tracker that using `window.runAntiAdBlock` will not work

I was wrong -- I didn't realize `window` was a property of `window`.

uBlock-user:
@okiehsch Should this go into Annoyances section because proboards was fixed there ? and the same anti-adblock script is used on the other two domains ?

okiehsch:
I can‘t test at the moment.
If the anti-adblock message is dismissable it is an annoyance if it is a „hard“ wall it should be fixed in uBO-filters.

uBlock-user:
I tested all three, it's dismissable on all three too and stays the same as a cookie named "aab" is lodged, so it should go in the Annoyances list.

@ryanbr Can you change the list to annoyances.txt ?

ryanbr:
Okay, done.

mapx-:
4 days ago I added in annoyances list:
`proboards.com##+js(aeld, scroll)`

I'll add the other 2 pages to the same filter.

llacb47:
Doesn't `aeld, scroll` have a bigger potential to break something?

PR #7194, opened 2020-04-04
Fix anti-adblock on aternos.org

llacb47:
Needed for mobile devices

Last PR: https://github.com/uBlockOrigin/uAssets/pull/7161

PR #7208, opened 2020-04-09

llacb47:
The old `calc` value caused Firefox to round `min-height` down to 0px, triggering anti-adblock.

Last PR: https://github.com/uBlockOrigin/uAssets/pull/7194


okiehsch:
Thx.

PR #7225, opened 2020-04-13
Fix suzylu.co.uk anti-adblocker filter

smege1001:

`suzylu.co.uk`

### Describe the issue

The anti-adblocker filter for suzylu.co.uk doesn't working anymore, because the regex that was used by this filter didn't include `AdBlock` (only the lowercase version of it)

### Versions

- Browser/version: Firefox 75.0

### Settings

Default settings


uBlock-user:
Too generic, can cause man false positives, fixed with https://github.com/uBlockOrigin/uAssets/commit/3f7a666acbd0d8a4f6b49c118a0ff4f93d342c9f

PR #7230, opened 2020-04-14

llacb47:
Well, only really a fix in Firefox. In chrome, you must click the button. I don't think there is any way to fix this in Chrome with uBO's current scriptlets.

The `pRxKtsfeNaSRHGiFYgtNRWrTntdJFgkArcyysNzSxrgViIW` is random for each user account, and the overlay cannot be hidden because the main content uses `!important` with inline styles. 

Last PR: https://github.com/uBlockOrigin/uAssets/pull/7208


okiehsch:
Fixed.

PR #7232, opened 2020-04-14
Fix anti-adblocker in vidlii.com

smege1001:

`vidlii.com`

### Describe the issue

The code has changed to bypass adblocker (message has changed (`Disable your adblocker!` -> `Disable Adblock`), the message is now in a element (`<addi>`))

There's also empty ad spots on the page

### Notes

Old versions of the anti-adblocker still exist in the page's source code

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -23831,6 +23831,7 @@ trendsderzukunft.de##+js(aopr, onload)
 ! https://github.com/NanoMeow/QuickReports/issues/3117
 @@||vidlii.com^$ghide
 vidlii.com##+js(nostif, cloudflare-app)
+vidlii.com##.adsbygoogle:style(display:none !important;)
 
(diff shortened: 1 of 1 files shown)


filters/filters.txt
```diff
@@ -23831,6 +23831,7 @@ trendsderzukunft.de##+js(aopr, onload)
 ! https://github.com/NanoMeow/QuickReports/issues/3117
 @@||vidlii.com^$ghide
 vidlii.com##+js(nostif, cloudflare-app)
+vidlii.com##.adsbygoogle:style(display:none !important;)
```
okiehsch, line 23834: Why the :style filter?
`vidlii.com##.adsbygoogle`
should work just fine.

smege1001: I forgot about that.

okiehsch:
Thx.

PR #7285, opened 2020-04-26
Fix adservers on ovagames.com

ryanbr:
Recently added some adservers; from this site `ovagames.com`

From this https://github.com/easylist/easylist/commit/6e893726d6bed9eb868818ae91236727bb0d6f70

This should limit these ad servers:
`||341073be6e9db7d2.com^`
`||e2706c6e2b426f74.com^`
..


okiehsch:
Thx.
The inline script triggering those requests is a generic one that we already defuse on dozens of websites..
https://github.com/uBlockOrigin/uAssets/commit/4b6f39cd36d134113495aeeeb44500835dafbec7

PR #7304, opened 2020-05-02
Fix sh.st: anti-adblock message

Yuki2718:

`http://sh.st/iA5JK`

### Describe the issue

Unobtrusive anti-adblock

### Screenshot(s)

sh

### Versions

- Browser/version: Brave 1.8.86

### Settings

- Default + Fanboy's Annoyances + uBlock Annoyances

### DIFF
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -3955,3 +3955,6 @@ suzylu.co.uk##+js(aeld, contextmenu)
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/54745
 ggwptech.com##+js(aopr, anOptions)
+
+! http://sh.st/iA5JK
+sh.st##+js(set, isAdsBlockerDetected, false)


mapx-:
I cannot reproduce FF / chrome

Yuki2718:
> 
> 
> I cannot reproduce FF / chrome

Reload several times.

Yuki2718:
BTW it seems a rule in uBlock filters `sh.st##+js(acis, parseInt, tabunder)` is outdated, maybe.

mapx-:
Did you disable brave protection (all of them) ?

Yuki2718:
Sure, if it blocked sth or upgraded connection to https the counter should appear.

brave

Of note, it's not 100% reproducible on my end too. Only occasionally.

mapx-:
Try a chrome portable installation too

Yuki2718:
> 
> 
> Try a chrome portable installation too

Okay, I can't reproduce on Chrome.

Yuki2718:
No idea, even with Brave's shields completely turned off and exactly the same entries on the logger, it only appears on Brave.

mapx-:
similar issue here: https://forums.lanik.us/viewtopic.php?f=100&t=44571

but in that case even using brave I cannot reproduce the same issue.

mapx-:
ok ... I tested now sh.st in brave and .. I get your same issue.

I watched the shields settings for the site (even if I disable globally like you any shields)
image

If I switch the orange button (disabling it) to grey then all is ok !

So, it seems brave has some other protections not explicitly listed in the global settings

Yuki2718:
> 
> 
> ok ... I tested now sh.st in brave and .. I get your same issue.
> 
> I watched the shields settings for the site (even if I disabled globally like you all shields)
> image
> 
> If I switch the orange button (disabling it) to grey then all is ok !
> 
> So, it seems brave has some other protections not explicitly listed in the global settings

Thank you for confirmation, it's a finding. Maybe I should report to or ask for Brave.

Issue #7233, opened 2020-04-14
t-mall.cmnw.jp


Yuki2718:
### URL(s) where the issue occurs

`https://t-mall.cmnw.jp/daily/kuji`
Account required

### Describe the issue

1. as stated in https://github.com/AdguardTeam/AdguardFilters/issues/53253 you have to wait for 15 sec to proceed.

2. Anti-adblock

### Screenshot(s)

For 1., see the AG issue. Attached SS for 2.
![tmall-ubo](https://user-images.githubusercontent.com/58900598/79177607-30cc7800-7e3e-11ea-8baf-d57c604b815b.png)

### Versions

- Browser/version: Brave 1.5.123 with responsive mode (as the AG issue was reported on mobile)
- uBlock Origin version: 1.24.4

### Settings

- Default

### Notes

For 1,
`t-mall.cmnw.jp##+js(set, sec, 0)`

For 2,
`t-mall.cmnw.jp##+js(set, adblock, false)`

If you can't create an account maybe I can help.

PR #7356, opened 2020-05-12
Update #7233

Yuki2718:
https://github.com/uBlockOrigin/uAssets/issues/7233

Removed the anti-adblock rule as it will not be fired any more due to a change in EL (`/adframe.` -> `/adframe.$~script`). Generalized another rule as per https://github.com/AdguardTeam/AdguardFilters/issues/53900 but haven't added other domains since AG Japanese is default-enabled.


mapx-:
thx

PR #7390, opened 2020-05-13
Fix sekai-kabuka.com right click

Yuki2718:

`https://sekai-kabuka.com/`

### Describe the issue

Annoyance: right-click disabled

### Versions

- Browser/version: Brave 1.8.95

### Settings

- Default + uBlock Annoyances

### Notes

There's a left-over of an ad which requires style syntax (e.g. `sekai-kabuka.com###况くく0o杤lﾞ:style(height: 0 !important)`) too as simple cosmetic rule breaks the site layout, but didn't add as it may be a job for AG Japanese filter.

sekai-kabuka


uBlock-user:
Resolve the conflict first by updating your annoyances list and then resubmit.

PR #7458, opened 2020-05-25
fix(7062): Unbreak viki player

vylmen:

https://www.viki.com/videos/1145603v (logged in)

### Describe the issue

Exception: pingAbFactor() is not a function

In JavaScript null is not a function, aside from that, whatever we
return must be truthy. So let's just return a function that returns
true.

Because of the thrown exception, the player plays about 1 second of
content in a continious loop, even if one has a paid account that
disabled ads.

### Screenshot(s)

### Versions

- Browser/version: Firefox 76.0.1 64 bit, MacOS
- uBlock Origin version: 1.27.4

### Settings

- Changed how the player interacts with adblock detector

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -3248,7 +3248,7 @@ kitguru.net##+js(aopr, anOptions)
 pornktube.*##+js(aeld, load, exoJsPop)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/7062
-viki.com##+js(set, VikiPlayer.prototype.pingAbFactor, function () { return true; })
+viki.com##+js(set, VikiPlayer.prototype.pingAbFactor, noopFunc)
 ! @@||pubads.g.doubleclick.net/gampad/ads*viki.com$xhr,domain=imasdk.googleapis.com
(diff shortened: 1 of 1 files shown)


uBlock-user:
I can't reproduce that when I tested, video plays fine, how is it getting stuck exactly ? Does it happens when you're logged in only ?

vylmen:
yes, logged in.
It gets stuck on this bit:
```js
            {
              key: 'onTimeUpdate',
              value: function () {
                if (!this.options.isLive) {
                  var e = this.player;
                  Math.ceil(e.currentTime()) >= e.duration() && e.duration() > 0 && this.emit(y.VIDEO_ENDED)
                }
                this.player.currentTime() >= this.abFactorCheckPoint && this.pingAbFactor()
              }
            },
```
Because with your commit is has been set to null, this.pingAbFactor() throws an exception. Another prerequisite may be that you have to resume a video so that it tries to forward and then hits this code. I'm not entirely sure about that, but the exception was in this bit.

uBlock-user:
How about `viki.com##+js(set, VikiPlayer.prototype.pingAbFactor, noopFunc)` ? Test that filter.

vylmen:
Updated

PR #7537, opened 2020-06-09
fix seoul.cs.land.to drag

Yuki2718:

`http://seoul.cs.land.to/`

### Describe the issue

Annoyance: drag disabled

### Versions

- Browser/version: Brave 1.9.76
- uBlock Origin version: 1.27.10

### Settings

- Default + uBlock Annoyances

### Notes


mapx-:
image

PR #7582, opened 2020-06-21
Targets anti-adblock-layer on t-online

tibor:

`https://www.t-online.de/nachrichten/deutschland/id_88096234/corona-ausbruch-bei-toennies-mitarbeiter-wohl-vor-quarantaene-abgereist.html
`

### Describe the issue

The element with the id targets ad block users to disable their adblocker.

### Screenshot(s)

Screenshot 2020-06-21 at 15 55 10

### Versions

- Browser/version: Chrome 83
- uBlock Origin version: 1.27.10

### Settings

### Notes


mapx-:
I cannot reproduce the issue

tibor:
Hi @mapx-, it’s dynamically inserted and only visible once. If you reload the page or return later, it won’t show up again.
Screenshot 2020-06-21 at 20 12 18

You have to remove the `Taabl` cookie or set its value to 0:
Screenshot 2020-06-21 at 20 16 42

uBlock-user:
Already fixed in uBO with `t-online.de##+js(set, abp, false)` in uBlock Filters, test on default settings.

PR #7637, opened 2020-07-05
Quora blur fix

smege1001:

`quora.com`

### Describe the issue

`.PageWrapper` is renamed to `#root` so it blurs the whole page again

### Screenshot(s)

screenshot


okiehsch:
Thx.

Issue #7691, opened 2020-07-22
Anti-adblock stickies on the bottom of many websites


DandelionSprout:
### URL(s) where the issue occurs

```
https://madensverden.dk/frikadeller/
https://bornholm.nu/
https://newsbreak.dk/drenge-faengslet-for-overfald-med-pistol-og-kniv/
https://gaffa.dk/
https://stiften.dk/
https://www.bilgalleri.dk/
https://www.recordere.dk/
https://kendte.dk/opdatering-v1-taarevaedet-kanye-west-begynder-valgkamp-med-kaotisk-moede/
https://www.gizmodo.jp/2020/07/216985.html
```

### Describe the issue

An English-language anti-adblock message sticky with Google-hosted image files, are shown on the bottom a lot of websites. It seems to be `||fundingchoicesmessages.google.com^$3p` in the main *uBlock Filters* list that is causing this, but I've had oddly much problems with verifying that it is the case.

### Screenshot(s)

![image](https://user-images.githubusercontent.com/22780683/88123006-4d5d1680-cbca-11ea-97da-a4442a4ed6b8.png)

### Versions

- Browser/version: Chrome 84.0.4147.89 x64 for Windows
- uBlock Origin version: Nano Adblocker 1.0.0.152 + Nano Defender 15.0.0.201

### Settings

![image](https://user-images.githubusercontent.com/22780683/88123071-6ebe0280-cbca-11ea-87ae-cb343d164108.png)
![image](https://user-images.githubusercontent.com/22780683/88123108-7f6e7880-cbca-11ea-9ada-1226cbbec954.png)

### Notes

Thankfully enough, the message overlays uses very similar `[style]` values, meaning that a generic non-advanced hiding rule to simply remove the banners should be pretty possible on paper. From the `bornholm.nu` example page:
![image](https://user-images.githubusercontent.com/22780683/88123269-d7a57a80-cbca-11ea-9ba7-1f7a5f771777.png)
I know that the `width:` percentage changes from site to site, varying from `101%` to `107%`. The rest of the values don't seem to change very often that I'm personally aware of.

I've added a makeshift specific entry in my Nordic list for the Danish example sites, so you may or may not want to turn the Nordic list *off* on your PCs while testing this, for simplicity's sake.

okiehsch:
>It seems to be `||fundingchoicesmessages.google.com^$3p` in the main uBlock Filters list that is causing this

That is a known issue.
[...]

PR #7696, opened 2020-07-23
Moved 1 entry from the Privacy list to the main list

DandelionSprout:
This is directly based on https://github.com/uBlockOrigin/uAssets/issues/7691, thus why I'm skipping the report template for this PR.

Although the uAssets team successfully argued for how I should've tested with a default uBO setup, I see no problems with moving the entry from one default list to another default list, since the entry turned out to be a lot more versatile than what was believed when it was first added 4 years ago.

I've turned on the "Allow edits by maintainers" button, if you guys want to modify this PR in any way without needing to ask me to do any such modifications.

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -3260,3 +3260,12 @@ vidtobo.com##+js(aopw, _pop)
 vidtobo.com##+js(aopr, LieDetector)
 ||supperopeningturnstile.com^
 ||gcxjczooe.com^
+
+! Examples of what is fixed by even an unfilled dummy API:
+! https://twitter.com/kenn_butler/status/709163241021317120
+! https://adblockplus.org/forum/viewtopic.php?f=10&t=48183
+! https://forums.lanik.us/viewtopic.php?f=64&t=32161
+! https://forums.lanik.us/viewtopic.php?f=64&t=30670
+! Also prevents many sites from trying to connect to "fundingchoicesmessages.google.com", thus preventing anti-adblock banners from appearing.
+! https://github.com/uBlockOrigin/uAssets/issues/7691
+||googletagmanager.com/gtm.js$script,redirect=googletagmanager.com/gtm.js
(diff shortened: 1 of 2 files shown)


mapx-:
"turned out to be a lot more versatile" is not a valid reason.

DandelionSprout:
Okay, so I should probably have formulated it as "Fixes way more things than just blocking tracking" instead. Is that a valid reason?

mapx-:
moving it between 2 default lists is completely useless.

okiehsch:
I already declined it https://github.com/uBlockOrigin/uAssets/issues/7691#issuecomment-662217428

The filter is redirecting a tracking scrip,t as a side effect that also fixes the google contributor anti adblock message on some sites.
`||googletagmanager.com/gtm.js` as a site specific filter is also part of `EasyPrivacy` becauses it is , you know, a **tracker**.

We use a redicrection for that request because simply blocking it will cause site breakage on many sites, so maybe you would have us put it in the the `unbreak` list next ?

DandelionSprout:
I see that the burden of explanation is on me, so I guess I can explain the admittedly unusual situation that I'm in regarding the lists I use.

Essentially, I've honestly been somewhat afraid of using privacy-specific lists, mostly due to some unclear memories I have from 2018 about EasyPrivacy breaking websites left, right and wide. I also don't actively mind trackers, although admittedly I have little experience with some of the nastier international trackers out there.

But on the other hand, I'm very big into anti-anti-adblocking, using tools like Nano Defender, https://raw.githubusercontent.com/AdguardTeam/AdguardFilters/master/AnnoyancesFilter/sections/antiadblock.txt, and https://raw.githubusercontent.com/Cybo1927/Hosts/master/Anti/Anti%20g00 .

This combo situation, which I understand is very unusual in the adblocking community, placed me in the dilemma about having to use a privacy list to utilise anti-adblock entries.

okiehsch:
> This combo situation, which I understand is very unusual in the adblocking community, placed me in the dilemma about having to use a privacy list to utilise entries with anti-adblock entries.

You can simply add `||googletagmanager.com/gtm.js$script,redirect=googletagmanager.com/gtm.js` to your personal filter list, no need to use the whole privacy list if you don't want too.

> Essentially, I've honestly been somewhat afraid of using privacy-specific lists, mostly due to some unclear memories I have from 2018 about EasyPrivacy breaking websites left, right and wide.

That is indeed the case, which is the reason other adblockers do not include it in their default setup.
I think uBO is the only one and that is also the reason we have a dedicated `unbreak` list, if you look through that list you will find that most breakages occur because some tracking script was blocked and needed to abe allowed to run for the site to work properly.

Yuki2718:
> 
> I have from 2018 about EasyPrivacy breaking websites left, right and wide. 

To me this is still somewhat the case after I've reported more than a dozen FPs. AGTPL has fewer FPs and uBlock Privacy even less, IIRC I've only got 2 FPs in uBlock Privacy and the one was exactly caused by GTM: `https://forums.lanik.us/viewtopic.php?f=64&t=44513&p=153945`. If not limited to uBlock Privacy, GTM is one of the most FP-prone rules. So it can partly be contradicting - moving the rule will increase the FP rate for people like you while it solves some anti adb.

PR #7764, opened 2020-08-13
Instagram login modal

llacb47:
Stops login modal when clicking on photos from instagram profile

Test here: `https://www.instagram.com/johncena/`


yourduskquibbles:
Thanks, great find

krystian3w:
IMO also disable pop-up after detect scroll down.

krystian3w:
https://github.com/AdguardTeam/AdguardFilters/issues/61429

PR #7780, opened 2020-08-17
Update Filters.txt to stop false positive blocking of Bugsnag JS

rob-race:
Bugsnag, an error catching service, provides their javascript library over a CloudFront URL. 

`http://d2wy8f7a9ursnm.cloudfront.net/bugsnag-2.min.js` 

It seems the CloudFront domain name is being blocked on this list and is causing issues rendering pages where Bugsnag JS is being used.

### Versions

- Browser/version: Latest Chrome on macOS
- uBlock Origin version: 1.28.4

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -1745,7 +1745,6 @@ jav.guru##.alignnone
 ||9anime.*^$script,1p
 @@||9anime.*/assets/$script,1p
 @@||9anime.ru/cdn-cgi/scripts/hcaptcha$script
-||d2wy8f7a9ursnm.cloudfront.net^
 ||soujoobafoo.com^
 ! https://www.reddit.com/r/uBlockOrigin/comments/e5b4h6/popups_and_ads_from_certain_websites_dont_get/
 ||bqiovml.com^


llacb47:
The blocking of bugsnag does not cause issue rendering pages in my experience.

rob-race:
Sure, and that may be application-specific at times. However, Bugsnag is a real use case of javascript to allow for a service being used. If not removing this entry, it there a different syntax to block the domain, except for the specific file mentioned above?

gorhill:
> causing issues rendering pages where Bugsnag JS

Can you provide specific cases where it's causing rendering issues?

Yuki2718:
IIRC, EasyPrivacy had been blocking Bugsnag but removed the rule. AdGuard started to block it relatively recently, which I feel somewhat inconsistent as they unblocked Crashlytics. But uAssets has its own criteria, and from what I understand - correct me if wrong - it's solely whether the thing is useful to <strong>user/client</strong> and NOT web master or web developer. So you need to provide an actual case that blocking it causes a problem to <strong>user</strong>.

Personally, what I've seen in this decade is majority of websites have been focusing only on conversion or revenue, having ignored user, which they call "improvements" but I "degradation".

PR #7825, opened 2020-08-26
Fix lookmovie icon breakage

llacb47:
`lookmovie.ag` is using play buttons from `nflxext.com` (netflix cdn)  (...?)

Example page: `https://lookmovie.ag/movies/view/8503618-hamilton-2020`

Reported by `https://old.reddit.com/iggwi7/`


okiehsch:
Thx.

PR #7829, opened 2020-08-26
Remove two obsolete rules

Yuki2718:

`https://vup.to/v/3k25asr2f395`
`https://streamp1ay.cc/2cjrsfxbyqww`

### Describe the issue

`@@||vup.to^$script,1p`
was probably for EasyList rule `/pop.js` which was removed so I guess it's no need any more.
`@@||streamplay.to/js/pop$script,1p`
also seems to be obsolete.

### Screenshot(s)

### Versions

- Browser/version: Firefox 79.0

### Settings

- Default

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -1469,7 +1469,6 @@ fux.com##+js(aeld, , /exopop|adb/)
 streamplay.*##+js(aopr, miner)
 streamplay.*##+js(aopw, Fingerprint2)
 steamplay.*##+js(aopr, btoa)
-@@||streamplay.to/js/pop$script,1p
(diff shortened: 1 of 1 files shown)


Yuki2718:
Maybe other exception rules for 1p script also need to be checked, and I think all rules including `Fingerprint2` as a property too - I haven't seen this property actually in use.

uBlock-user:
`@@||vup.to^$script,1p` is required because of `.to/ads/` in Easylist. Second case filter can be removed.

Yuki2718:
@uBlock-user Thanks, probably I should have made an issue than PR. Should I correct that or can you do on your side?

PR #7849, opened 2020-08-30
Fix https://github.com/NanoMeow/QuickReports/issues/4601 manga1000.com

Yuki2718:
https://github.com/NanoMeow/QuickReports/issues/4601


uBlock-user:
Did you test for any breakage ? `onload` is used by two other inline-scripts.

Yuki2718:
@uBlock-user As far as I can tell, everything is working fine.

PR #7915, opened 2020-09-15
Fix www.sony.jp can't switch tabs

Yuki2718:
The same issue as https://github.com/easylist/easylist/pull/6062 so please forgive me for skipping the required steps. Due to `tags.tiqcdn.com` blocked by Peter Lowe's another fix is needed on uBO.

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -3886,3 +3886,6 @@ hentaiplay.net###video_overlays
 !#if env_firefox
 @@||privacy-mgmt.com^$script,domain=handelsblatt.com
 !#endif
+
+! www.sony.jp can't switch tabs
+@@||tags.tiqcdn.com/utag/sony-marketing/main/prod/utag.sync.$script,domain=www.sony.jp


mapx-:
tested in firefox. working for me adding:

```
@@||tags.tiqcdn.com/utag/sony-marketing/main/prod/utag.sync.$script,domain=sony.jp
@@||assets.adobedtm.com/*/satelliteLib-$script,domain=sony.jp
@@||assets.adobedtm.com/*/mbox-contents-$script,domain=sony.jp
```

test again and please use `domain=sony.jp` and not `domain=www.sony.jp`

Yuki2718:
@mapx- Aren't they exactly the same as mine except for www? I committed in EP as these adobedtm are from EP. As to www, I would have done so if it's blocking rule, but no need to allow them in other subdomains such as `https://acafe.msc.sony.jp/photo/detail/item/000034718372cg?ltyp=15&lpag=1&ltag=&ltmd=&lcmr=&llen=&lphr=%E9%B3%A5`.

PR #7924, opened 2020-09-17
Unbreak Sentry while keeping grofers.com exception

piotr-cz:

We use Sentry.io in web application that is not closed to public and so cannot provide an URL.
It's API endpoint URL is requested usually only in case of unexpected events so only way to demonstrate problem would be via controlled error,

### Describe the issue

Sentry.io is an error tracking utility, one of it's platform allows to be used in web browsers.

In commit https://github.com/uBlockOrigin/uAssets/commit/6489ab5eaf2a838bcdf2ae74ec8542ebb897d16d sentry.io was added to the privacy list, apparently because the `grofers.com` site is misusing it to report user fingerprints (more information is in mentioned commit comments).

Unfortunately this breaks all web apps that use [@sentry/browser](https://www.npmjs.com/package/@sentry/browser) package as these are not able to send error events anymore.

This Pull requests changes filter restricts filter to be applied only in case of `grofers.com` site:

```diff
-||sentry.io^$3p
+||sentry.io^$3p,domain=grofers.com
```

### Screenshot(s)

Log from Chrome DevTools console
```
https://xxx.ingest.sentry.io/api/xxx/store/?sentry_key=xxx&sentry_version=7
net::ERR_BLOCKED_BY_CLIENT
```

### Versions

- Browser/version: Chrome/ Version 85.0.4183.102 (Official Build) (64-bit)

### Settings

- no changes to default settings

### Notes

none

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -222,7 +222,7 @@ frogogo.ru##+js(aopw, ADMITAD)
 ||artfut.com/static/tagtag.$script,3p,redirect=noop.js
 
 ! grofers.com sentry data-collection/fingerprinting
-||sentry.io^$3p
+||sentry.io^$3p,domain=grofers.com
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/62696
 ||freegames.io/log^


uBlock-user:
Already explained in https://github.com/uBlockOrigin/uAssets/commit/6489ab5eaf2a838bcdf2ae74ec8542ebb897d16d#r42198269

It's not about one domain, a tracking service regardless of what it's principle business is always blocked if not blocked in EasyPrivacy itself. Secondly, sentry not receiving the data is none of uBO's concern as uBO always represents the users.

piotr-cz:
Sentry is error reporting - not tracking utility.
Preventing it from functioning negatively affects uBlock users as errors that occur in browser won't be recorded - and in effect won't be fixed.

The problem is that it's extremely hard to present it's proper usage (developers try to minimize errors) while trivial to reproduce misuse (sending fingerprints like in grofers.com case).

Library by default doesn't report to API endpoint on visiting web site, only in unexpected cases.

The https://github.com/uBlockOrigin/uAssets/commit/6489ab5eaf2a838bcdf2ae74ec8542ebb897d16d#r42198269 comment you relate to describes an abuse,
while next one https://github.com/uBlockOrigin/uAssets/commit/6489ab5eaf2a838bcdf2ae74ec8542ebb897d16d#r42199597 describes proper use.

Please think it over.

gorhill:
@uBlock-user, [you said](https://github.com/uBlockOrigin/uAssets/commit/6489ab5eaf2a838bcdf2ae74ec8542ebb897d16d#r42198269):

> via a `POST` request when you visit `https://grofers.com/` or any other domain that employs your service.

You have examples of other domain?

I received an email from a Sentry person asking me to reconsider the broadness of the filter.

If requests are sent to `sentry.io` unconditionally upon visiting a site, I agree this is an issue from a user perspective. Even if the requests were sent only upon errors being raised in the browser, this can also be an issue since uBO by merely doing its work as a content blocker can cause many errors to be raised, and those should not be reported as site issues.

uBlock-user:
It's sent unconditionally on grofers and but on myanimelist.net it's conditional if we were to use a filter with `:remove()` operator.

gorhill:
I think for now we should narrow to `grofers.com`, but we will need to keep an eye on:

- Are we finding more sites where the unconditional `POST` upon visiting a page occur?
- Are we finding more cases where `POST` are triggered by uBO merely doing its job of blocking network requests by enforcing default filter lists?

uBlock-user:
> for now we should narrow to grofers.com

https://github.com/uBlockOrigin/uAssets/commit/9ab563e5fb6c2c8be78aa1ac47e429e057a0ead0

uBlock-user:
@gorhill Found another -- `https://www.rocketpunch.com/companies/amorepacific` unconditional `POST` requests, if you can't reproduce on first load, try opening other company pages and it will fire.

piotr-cz:
To test conditional event reporting on some of the `https://publicwww.com/websites/ingest.sentry.io/` sites, run this in DevTools console:

```js
setTimeout(() => { throw new Error('Debug') }, 1e3)
```

piotr-cz:
> @gorhill Found another -- `https://www.rocketpunch.com/companies/amorepacific` unconditional `POST` requests, if you can't reproduce on first load, try opening other company pages and it will fire.

Probably peple from Sentry organization could explain what's going on in specific cases /cc @lobsterkatie

dcramer:
Hey folks - founder/CTO of @getsentry here.

So Sentry is getting broader, but there's some things I'd like to add to the conversation:

1) Errors are mission critical to developers. That's our core product. If you see a POST to ``ingest.sentry.io`` its VERY LIKELY them recording an error. That error may or may not be visible to the end user, but without Sentry that team will _never_ know about those errors (unless you're raging at them on Twitter)

2. We have a newer product called Transactions which could be conflicted with Analytics. It doesn't do user tracking or cookie the end user, so its far less invasive in privacy, but I would accept a fair argument that it can be used in other ways. This product is fundamentally the same as our error monitoring, but it captures latency-related concerns and not just errors. This might be what you're seeing with the unconditional reporting.

What I would be interested in understanding is how we can work with the ecosystem to ensure folks can trust Sentry, and what improvements we might be able to offer our customers so that they can continue to.. frankly, develop software using Sentry. We've had similar conversations with other communities around this (e.g. pi.hole), and I realize there's a lot of folks with strong opinions, but genuinely we are just here trying to help software developers ship less defective software.

dcramer:
Aside I just reviewed what is being labeled as a fingerprint.. that's a stacktrace of an error. It has nothing to do with a user, and is simply the mechanism to reproduce the function call. It is not used to identify a user, and without be extraordinarily clever (in an unscalable and expensive way, which they are not) it cannot ever be used to do that.

gorhill:
The key is to look at this from the user's perspective: some information is sent to a 3rd party when visiting a given site.

Pretty much all 3rd parties which are blocked by default in filter lists used by uBO will (as some have) argue that what they collect is not for tracking/fingerprinting etc. This is irrelevant from a user perspective, since no user has the power to audit those claims, so they don't really factor in whether we should remove a block filters or not.

dcramer:
@gorhill are you suggesting that every third party service in the world should be blocked by uBlock? As a long time user that is not what I would have expected.

If that's not true, and it's only conditionally based on what the data is, or what its used for, I don't think its fair or relevant that a user simply _thinks_ its malice. We can educate folks, and we can prove what it is and isn't (even just from the JSON we send). It's also important to note that if someone wanted to, they could even audit how the data is used, and I could link them to the literal code and specifications for some of these things, as [Sentry's source is all public](https://github.com/getsentry/sentry)

I also think its worth understanding that when you're using these services, they're required to adhere to their terms of service (which they're required to disclose what they would do with Sentry, if its at all linked to a user's personal information). In general people _don't_ want to use Sentry and have it associate with user information due to the complexities of things like GDPR and CCPA these days. They actually prefer Sentry not store any identifiable information. I realize this latter bit may not satisfy folks, but it is important none the less.

dcramer:
Aside, I haven't looked at Grofers, but there are plausible explanations for a request being fired right away on a page load, with zero malice at all. JavaScript has a _lot_ of points of failure, and its a newer ecosystem, so its often that these companies legitimately have errors that happen consistently all over the place. We've certainly experienced it ourselves, and I see it on many big company websites (you'll often see it by just opening up Chrome's console)

gorhill:
In case you missed it, [the filter has been narrowed](https://github.com/uBlockOrigin/uAssets/pull/7924#issuecomment-694851062) to apply only to `grofers.com` at this point.

dcramer:
@gorhill yeah I saw and I appreciate that. I mostly want to help educate folks and understand the concerns. Generally speaking, when we've had this happen in the past, it happens again not too far along in the future. So I want to know if there's more we can do to help avoid those situations or be more pro-active. If we're good at this point that's fine, but we're also here if folks want more information.

travigd:
I found this issue when I noticed Sentry was being blocked on my app. I'll echo what @dcramer has said - Sentry is pretty essential to finding bugs when they happen, and importantly, tracking them back to actual lines of code (which is otherwise a difficult task given how complex frontend development infrastructure has become).

Also, thanks to everyone in this thread, it seems like it's been very productive/constructive despite the inherent subjectiveness of where these lines get drawn (I've seen lots of these quickly devolve into flame wars, so I'm very appreciative).

Yuki2718:
I appreciate the debate here and want to ask about one thing raised by gorhill which is not yet fully answered: why so many bug reports if you use contents-blocker? It's not only Sentry, Bugsnag too - if you use any blocker, particularly in hardened modes i.e. default-deny, I see bunch of bug reports sent here and there. I doubt they're all real bugs. Each packet is negligibly small in size, but they can be cumulated up and I'm in tight bandwidth restriction on mobile (I don't watch movies on mobile). If it's impossible to distinguish a real bug from non-bug, I still have a reason to block them.

dcramer:
@Yuki2718 I think thats fine as a personal decision, but that to me is not a good reason to block them for all users of uBO. The bandwidth concern is totally fair, but can be applied to just about everything that an application is doing. An analogy I'd pass: Should uBO also block _all_ other scripts, images, etc? 

I'd say instead you should stop doing business with sites that are you eating away at your bandwidth - or in general are low on the performance bar. I know that's not always possible, but it is a choice you have. If the internet started pushing it more (which to be fair, it is), all sites would have to be a bit more thoughtful about how large their assets are, and how much traffic they are consuming.

Yuki2718:
@dcramer Bandwidth is not a main point. I'm okay to send a real bug report to a nice site. What I don't like is that bug reports are fired everywhere every time (yeah, bit exaggerated but still) if you use contents blocker which I suspect are NOT real bugs. This should also be problem to your customer as these non-bug reports will be noise. So I asked if you can't do anything about this. Unlike other guys here I'm not a developer so have no idea.

dcramer:
Can't really comment on specifics since it's based on the customer's implementation, but from the developer lens there are a number of things that could cause these noisy errors. So while I agree they can be noise, knowing _how_ noisy can also help teams prioritize what to address. Teams may choose to filter them out (so you'd stop sending them to Sentry e.g. try..catch in simple terms), but only if it's not important to them. JavaScript is still the wild west in a lot of ways - it's not a very mature ecosystem which is one of the reasons why Sentry is so desperately needed by folks, but it's also why the errors which you might see as constant/noisy are so tough for developers to deal with.

Yuki2718:
@dcramer Okay, thank you for taking time to explain!

PR #7305, opened 2020-05-02
Fix affiliate.fc2.com: anti-adblock message

Yuki2718:
### URL(s) where the issue occurs

`https://affiliate.fc2.com/`

### Describe the issue

Unobtrusive anti-adblock

### Screenshot(s)

fc2

### Versions

- Browser/version: Brave 1.8.86
- uBlock Origin version: 1.26.0

### Settings

- Default + Fanboy's Annoyances + uBlock Annoyances

PR #7965, opened 2020-09-28
Fix google-funding on globo.com

ryanbr:

`https://www.globo.com`  has google funding warning

### Describe the issue

### Screenshot(s)

globo

### Versions

- Browser/version: Brave

### Settings

### Notes


mapx-:
@ryanbr did you test chrome / FF (not only brave) ?

mapx-:
it's brave => shields up but nothing it's blocked apparently (already happened if you remember and you reported with brave guys). If I turn off the shields no issue

image

mapx-:
@ryanbr same behaviour for `latimes`, it's brave fault (but using us ip)

Yuki2718:
I don't see the message neither on globo or latimes even with Brave + uBO double filtering, maybe regional matter? BTW https://github.com/uBlockOrigin/uAssets/pull/7305 was also Brave issue.

PR #7984, opened 2020-10-02
Update for twitch.tv #5184

pixeltris:
This is an update to the twitch.tv ads workaround. Solution by @diglarboard

### DIFF
--- a/filters/resources.txt
+++ b/filters/resources.txt
@@ -1699,7 +1699,7 @@ damoh-defuser.js application/javascript
 })();
 
 
-# https://github.com/uBlockOrigin/uAssets/pull/3517
+# https://github.com/uBlockOrigin/uAssets/issues/5184
 twitch-videoad.js application/javascript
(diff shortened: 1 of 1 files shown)


llacb47:
Wrong. This is where it needs to be changed, and only gorhill can do it.

https://github.com/gorhill/uBlock/blob/master/assets/resources/scriptlets.js

pixeltris:
Oh right I wasn't aware of the change in location of this. Why does this file still exist?

krystian3w:
for uBO Legacy or very old like 1.13.6

gorhill:
Can you open a pull request for <https://github.com/gorhill/uBlock/blob/master/assets/resources/scriptlets.js#L1329>?

pixeltris:
https://github.com/gorhill/uBlock/pull/3781

tigo9:
any update on this? it stopped working the last day or so

PR #7994, opened 2020-10-03
fix https://github.com/NanoMeow/QuickReports/issues/4778

Yuki2718:
Notorious (i.e. better to use denyallow) sites in japan but won't be fixed in AG Japanese.


okiehsch:
Thx.

PR #8005, opened 2020-10-05
Further update for javhay.net [NSFW]

Yuki2718:
Test URL: `https://javhay.net/tag/sex-toys/`


uBlock-user:
`javhay.net##+js(addEventListener-defuser, load, 2000)` already present in uBlock Filters. Can't reproduce.

Yuki2718:
Yes, no anti-adb as long as the scriptlet is. Can I add `||javnung.net^$3p` and remove `javhay.net##+js(addEventListener-defuser, load, 2000)` or no need?

uBlock-user:
No need to, player and video is hosted on that domain, will block `mimilcnf.pro`

PR #8045, opened 2020-10-17
Update badlists.txt

D0n-A:
Added alternative link (?) for Malwaredomains.

That it was not possible to install this filter from an alternative source.


gwarser:
Malwaredomains is here only because it was in default selection of lists before. This section of `badlists.txt` is used to remove stale entries from users custom filter lists section. This whole part of this file should be probably removed now, when we have 1.30.4 already released. @gorhill?

gorhill:
We can still wait, I would wait until the next release cycle before removing soft block of lists.

In any case, the pull request here is irrelevant to uBO, it's a list which was never part of stock filter lists, and it was found that the list from `malwaredomains.com` is no longer updated really.

gwarser:
> soft block of lists

"Soft" as "we don't have them to be blocked permanently"?

gorhill:
Yes, eventually people can import if they want, though it's inadvisable to import no-longer-maintained lists, we are not going to start to forbid these, that would be a burden. We forbid only those who are known to cause filtering issues.

Issue #8041, opened 2020-10-16
pixiv.net (ad placeholder cleanup)


paperclip-dayo:
### URL(s) where the issue occurs

`https://www.pixiv.net/artworks/*`
Example: `https://www.pixiv.net/artworks/84998112`

### Describe the issue

Clicking the bookmark button (heart) slides down a frame below the image with ad banner, UBO removes the banner but doesn't prevent an ad frame from being shown as an empty placeholder. Requires pixiv account (free).

### Screenshot(s)

![example](https://user-images.githubusercontent.com/69221940/96255956-7a9a4480-0fc0-11eb-95e2-32be9bbfc59d.jpg)

### Versions

- Browser/version: Edge 87.0.664.8
- uBlock Origin version: v1.30.5b1

### Settings

All "Ublock" filters + adguard base

### Notes

This has been an issue for quite a long time with vanilla UBO after pixiv made some UI changes (for the 100th time) but I've used NanoContrib filters to deal with empty placeholders. Since Nano project has been sold and turned into malware, it'd be nice to have some of its placeholder fixes incorporated into UBO.
Relevant filter list: `https://github.com/NanoAdblockerLab/NanoContrib/blob/master/dist/placeholder-buster.txt`
Currently using this part of the filter list as my custom filters deals with the placeholder issue but it might be worth reviewing in case there's redundant code or a better solution available before incorporating it into UBO filter list:
```
pixiv.net###js-mount-point-header.with-ad:style(min-height: 0px !important;)
pixiv.net###root > header + div + div:not(:has(a)):not(:has(img))
pixiv.net##.ads-top-info
pixiv.net##.area_new:not(:has(.area_title))
pixiv.net##.sc-AykKF:style(padding-bottom: 0px !important;)
pixiv.net##a[href$="=work_detail_remove_ads"]:upward(3)
pixiv.net##div:has(:scope > div > iframe:not([src])):matches-css(padding-bottom: 56px)
pixiv.net##iframe[height="520"]:not([src]):upward(1)
pixiv.net##iframe[height="60"]:upward(1)
pixiv.net##iframe[height="90"]:upward(1)
pixiv.net##li[size="2"][offset="1"]
```

uBlock-user:
Out of the list, these 4 are getting hits in uBO --

```
pixiv.net##a[href$="=work_detail_remove_ads"]:upward(3)
[...]

PR #8046, opened 2020-10-17
Update #8041 pixiv.net placeholder

Yuki2718:
https://github.com/uBlockOrigin/uAssets/issues/8041

Any test/review/improvements by those who use Pixiv more extensively like @paperclip-dayo and @DandelionSprout  are welcome.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -15678,17 +15678,13 @@ freebcc.org##+js(aopr, adBlockDetected)
 
 ! https://github.com/DandelionSprout/adfilt/issues/7#issuecomment-493395377
 ! https://github.com/uBlockOrigin/uAssets/issues/6726
-pixiv.net###js-mount-point-header.with-ad:style(min-height: 0px !important;)
-pixiv.net###root > header + div + div:not(:has(a)):not(:has(img))
-pixiv.net##.ads-top-info
-pixiv.net##.area_new:not(:has(.area_title))
-pixiv.net##.sc-AykKF:style(padding-bottom: 0px !important;)
-pixiv.net##a[href$="=work_detail_remove_ads"]:upward(3)
-pixiv.net##div:has(:scope > div > iframe:not([src])):matches-css(padding-bottom: 56px)
-pixiv.net##iframe[height="520"]:not([src]):upward(1)
-pixiv.net##iframe[height="60"]:upward(1)
-pixiv.net##iframe[height="90"]:upward(1)
-pixiv.net##li[size="2"][offset="1"]
+pixiv.net###js-mount-point-header.with-ad:style(min-height:0!important;)
+pixiv.net##._illust-upload + div[style^="display:block;width:968px;height:170px;"]
+pixiv.net##iframe[name="dashboard_home"]:upward(div[span="1"])
+pixiv.net##iframe[width="300"][height="250"]:upward(2)
+pixiv.net##iframe[width="468"][height="60"]:upward(1)
+pixiv.net##iframe[width="500"][height="520"]:upward(1)
+pixiv.net##iframe[width="728"][height="90"]:upward(2)
 
 ! https://forums.lanik.us/viewtopic.php?f=62&t=43033
 ||nextclick.pl^$3p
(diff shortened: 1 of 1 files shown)


DandelionSprout:
I am probably far too late to chip in on all this stuff, but these seem to be the most important entries I could find in 15min after a very long day of sleep:
```
!!! Assumes the absence of Pixiv Premium, or that "Ad display settings" at https://www.pixiv.net/setting_user.php is set to "Show".
!!! Also assumes that AdGuard Base is turned on.

! https://www.pixiv.net/en/ (Requires being logged on)
pixiv.net##div:not(.submit-area):has(:scope > iframe:not([src]))
pixiv.net##div[class^="sc-"]:has(:scope > div[style] > iframe[width="728"][height="90"])
pixiv.net##li[size]:has(:scope > div > iframe[name="topbranding_rectangle"])
! https://www.pixiv.net/en/tags/princess_dress
pixiv.net###root > div > div > div:has(iframe[name="footer"]):not(:has(img))
! https://www.pixiv.net/en/tags/%E4%BA%BA%E9%AD%9A%E3%82%B9%E3%83%BC%E3%83%84/artworks?mode=r18 (Requires being logged on and that R-18 is turned on in the user settings)
pixiv.net##li:has(:scope > iframe[name="infeed"])
```

I had an oddly hard time reproducing the rest of the Pixiv entries from Placeholder Buster.

Yuki2718:
@DandelionSprout You're not too late, the commit has not been merged. Just curious, have you added the committed rules? I think e.g. `pixiv.net##div[class^="sc-"]:has(:scope > div[style] > iframe[width="728"][height="90"])` should be covered by `pixiv.net##iframe[width="728"][height="90"]:upward(2)`, not?

DandelionSprout:
I didn't use the new rules at the time of testing, I just checked with uBO's logger which of the old entries I was even able to reproduce.

Yuki2718:
Okay, we don't need to hurry. I have no problem to edit or close this PR. My rationale for the new rules are here: https://github.com/uBlockOrigin/uAssets/issues/8041#issuecomment-711010387 It's no harm to create rules from scratch, I think. I haven't tested R18 mode BTW.

Yuki2718:
So thanks to @DandelionSprout 's help in the above comment I could find it:

pixiv

paperclip-dayo:
Also a bit late to the party but thanks for cleaning up those pixiv rules 👍 
Did a quick browse of random pages and nothing appears to be broken.
In the meantime I did find a couple more of other empty placeholders after the ad banner gets killed that might be worth cleaning up.

Between comments and related images on the illustration pages:
illust

Manga page `https://www.pixiv.net/manga` top header:
manga_top

Manga page `https://www.pixiv.net/manga` between editors recommended and pixivision section:
manga_bot

Homepage `https://www.pixiv.net/`, closer to the top of the page, above "popular tags":
mainpage

Yuki2718:
There's no too-late for a site like pixiv, it's near-impossible for a single person to make a perfect rule set without these helpful feedback. I'll check them later.

Yuki2718:
Okay, everything except for the second SS could be covered by `pixiv.net##div[class^="sc-"]:has(> div[style] > iframe[width="728"][height="90"])`:

<details>

pixiv1

pixiv3

pixiv4

</details>

About the second, on my end it's not blank screen:

pixiv2

paperclip-dayo:
Thanks, regarding the second ss, I checked with UBO off and it might be because I'm getting a different banner served from `pixon.ads-pixiv.net` 
banner

Makes sense to leave it untouched in this case I suppose since both are pixiv-related banners and not 3rd-party ads.

Yuki2718:
@paperclip-dayo Actually most placeholder rules added are related to `ads-pixiv.net`. It's possible they serve ads to you but not to me. Can you right-click on the ads, choose 'Inspect element' (on Firefox, on Chromium it's 'Inspect') and take SS so that I can see DOM tree of that staff?

paperclip-dayo:
I'm getting the same banner as you now, the one with free month of pixiv premium, so can't do that for now.
Gotta remember to do it next time I get a different banner from that server :)

PR #8160, opened 2020-11-05
Fix anti-Brave message on archive for uBO/Brave

ryanbr:

`https://archive.is/` is using an Anti-Brave check, this patch is for Brave/uBO (assuming the user is disable shields)

### Describe the issue

Site is using Anti-Brave checks, and user is unfairly targeting the browser. We worked around it in Brave with a similar fix, but for users of uBO and Brave, shouldn't be forgotten.

### Screenshot(s)
579ef1665583fd91853dfb84757566138c6f40ed

### Versions

- Browser/version: Brave Version 1.16.68 Chromium: 86.0.4240.111 (Official Build) (64-bit)

### Settings

### Notes

Breaking this check, targeting (US IP)

`try { if (Object.keys(Object.getPrototypeOf(navigator)).join(' ').match(/\bbrave\b/)) { document .cookie = 'us=1;path=/;max-age=86400'; document.location.href='/unsupported-browser'; } } catch(e) { document .cookie = 'us=1;path=/;max-age=86400'; document.location.href='/unsupported-browser'; }`

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -4034,3 +4034,7 @@ lebigdata.fr##.background-cover
 
 ! https://github.com/uBlockOrigin/uAssets/issues/8158
 @@||tags.tiqcdn.com/utag/usbank/global-sync/prod/utag.sync.js$script,domain=usbank.com
+
+! Brave specific (using Brave/uBO)
+archive.is,btdig.com,archive.today,archive.vn,archive.fo,archive.md,archive.li,archive.ph,archiveiya74codqgiixo33q62qlrqtkgmcitqx5u2oeqnmn5bpcbiyd.onion##+js(acis, document.cookie, document.location.href)
+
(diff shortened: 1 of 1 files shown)


gorhill:
What does this filter do to the page? Is the site broken by any filter in a default filter lists?

DandelionSprout:
My understanding of the discussion about it on Slack, is that `archive.is`'s owner really, really hates Brave Browser, to the point where the site runs a check(s) to detect Brave Browser and then deny access to all users of it.

uBlock-user:
They're blocking all Brave browser users from accessing the site via UA detection as per that script.

gorhill:
> `archive.is`'s owner really, really hates Brave Browser

I have nothing against Brave but this does not fit the purpose of any filter lists -- sites are free to break themselves if this is what they want but _Unbreak_ is to work around breakage caused by filters in default filter lists.

If we start doing this for Brave, then people will also ask to do this for website A which hates Firefox or website B which hates Google Chrome, and so on. I have my own preference but I don't want to push them in uBO's filter lists.

We need to stick to the purpose of lists otherwise such move will undermine the independence of uBO as a project.

Issue #7301, opened 2020-05-01
NSFW Ads adult sites


ghajini:
### URL(s) where the issue occurs

**NSFW**

`https://beegsexxx.com/` /fixed

`https://beeg.party/video/novice-dishonest-wifey` /fixed

`https://www.pornhd.com/` fixed

`https://m.spankbang.com/`  fixed

`https://spankbang.name` /fixed

`https://xnxxxvideos.co/` /fixed

`https://spankbang.cc/` /fixed

`https://www.porndig.com/` fixed

`https://www.vporn.com/` fixed

`https://www.pornflip.com` fixed

`https://www.tubxporn.com/` /fixed

`https://www.peekvids.com/`  - **clicking on video results in strict block page and not video** /fixed

`https://www.pornid.xxx` fixed

`https://www.pinflix.com/` fixed

`https://www.pornrox.com/` fixed

`https://www.youporn.com/watch/15890956/stepdad-woke-up-his-stepdaughter-and-then-she-started-fuck-with-him/` - **video not playing**

`Txxx.com`

`http://v.xtapes.to/` fixed

### Describe the issue

Popup ads, popunders ,crap xhr's 
Clicking on homepage videos results in ads

### Versions

- Browser/version: firefox A.
- uBlock Origin version: 1.26.2

### Settings

-  uBO's default settings

mapx-:
`youporn` + `txxx` are ok on my end

ghajini:
NSFW
`https://txxx.com/videos/1263243/black-guy-fucks-nicole-and-gives-her-facial/`

hamburger menu is inaccessible

mapx-:
it seems `txxx.com##+js(aeld, click, target)` is to keep only for desktop, can you confirm (and will have still the popunder) ?

ghajini:
Wait there are popunders on every click

mapx-:
first 2-3 on my end

ghajini:
Isn't that annoying popunders though blocked results in reload of page everytime?
What can be done to stop that?

mapx-:
We could stop popunders using `||txxx.com/covad/`

ghajini:
All looks good👍

ghajini:
**NSFW**
`https://txxx.tube/videos/2933210/megan-sage-tommy-gunn-in-shake-your-money-maker-brazzers/`

Popup ads

ghajini:
Popup ads

`https://www.pornid.name`

[...]

PR #8228, opened 2020-11-16
Update #7301 [NSFW]

Yuki2718:
`vporn.com` redirects to `pornone.com` so I removed filters for the former and added filters for the latter under #7301. If it's not appropriate tell me or fix on your side.


Yuki2718:
My bad, `||tds.pornvideotop.com/general/exoclick-` is actually on `beeg.party` but somehow I put under `beegsexxx.com`. Can you fix it? Sorry!

mapx-:
does not matter

Yuki2718:
Relieved :)

Issue #1614, opened 2018-02-24
stream2watch.org has redirects, focus stealing redirects in new tabs, pop-ups, pop-unders, ads, you name it!


MarsWanKenobi:
Replace the bracketed [...] placeholders with your own information.

### URL(s) where the issue occurs

`http://www.stream2watch.org/stream/68305`

### Describe the issue

When navigating to the above mentioned URL, I get all of the mentioned annoyances in the title.   

### Versions

- Browser/version: Firefox Nightly 60.0a1 (2-23-2018)
- uBlock Origin version: 1.15.10

### Settings

- Default settings

okiehsch:
That is quite a site, the stream is inside nested iframes, I counted 5.
Anyway
````
stream2watch.org##script:inject(abort-on-property-write.js, Fingerprint2)
wizhdsports.fi##script:inject(nowebrtc.js)
||d3q33rbmdkxzj.cloudfront.net^$image,domain=stream2watch.org
````
gets rid of all the popups and redirects on my end, but more could be needed for different streams.

**Comment by MarsWanKenobi** [comment]
Works! I tried out other streams this morning and I get no pop-ups (at least nothing permanent, just a white flash where the tab would want to open up then it would immediately close), no ads, no annoyances. Thank you for your speediness!

**Comment by ryanbr** [comment]
https://github.com/easylist/easylist/commit/4a5ae7aecaa6ff0ee6b3ffc5c9cecdd3f6f2ff8c

Should also clean it up.

**Comment by SkeletalDemise** [comment]
@okiehsch There's a redundancy. `||d3q33rbmdkxzj.cloudfront.net^$image,domain=kimcartoon.me` is redundant by `||d3q33rbmdkxzj.cloudfront.net^`

**Comment by okiehsch** [comment]
Thx, actually `||acloudimages.com^$script,domain=kimcartoon.me` is also redundant.
`||acloudimages.com^$third-party` in `EasyList`

**Comment by smed79** [comment]
add 
```
@@||sportzonline.xyz/channels/$subdocument,domain=stream2watch.org
```
https://forums.lanik.us/viewtopic.php?p=144859#p144859

PR #8236, opened 2020-11-17
Update #1614 (popup)

Yuki2718:
Still got popup (fixed by this filter) and I don't see ab or Fingerprint2 even with uBO turned off.


uBlock-user:
Is  `stream2watch.*##+js(set, ab, false)` no longer useful?

Yuki2718:
I don't see any problem with `stream2watch.*#@#+js(set, ab, false)`.

PR #8241, opened 2020-11-18
Update live-golf.stream

Yuki2718:
Not sure if it helps `https://old.reddit.com/r/uBlockOrigin/comments/jutds7/is_ublock_causing_this_problem_or_is_it_something/` as I can't reproduce the issue, but think no harm to add these.


mapx-:
only in FF some sound glitch, chrome ok.

Yuki2718:
Yeah, same as commented in Reddit.

PR #8270, opened 2020-11-24
Block bulbagarden custom ad server

shirt-dev:

`https://bulbapedia.bulbagarden.net/wiki/Main_Page`

### Describe the issue

Bulbagarden created their own ad servers to insert ads that are not present while logged in. This pull request mitigates the issues. It also removes the top menu which also is not present while logged in.

### Screenshot(s)

Without filter
image

With filter
image

### Versions

- Browser/version: Chrome 87
- uBlock Origin version: v1.30.6

### Notes

Without the cosmetic filter, the menu is displaced.

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -6273,3 +6273,9 @@ bollyholic.pw##div[class][style="float: none; margin:10px 0 10px 0; text-align:c
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/68391
 tecnomusic-evolution.com##+js(nosiif, visibility, 1000)
+
+! bulbagarden.net ads
+||ahost.bulbagarden.net
+||bulba-ad-host-website.azurewebsites.net
+bulbapedia.bulbagarden.net###navbarc > div
+


okiehsch:
Quoting our [readme](https://github.com/uBlockOrigin/uAssets/blob/master/README.md)

> If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable.

Your issue only needs normal syntax and should be fixed in EasyList.
You can open an issue here: https://github.com/easylist/easylist/issues

shirt-dev:
> Quoting our [readme](https://github.com/uBlockOrigin/uAssets/blob/master/README.md)
> 
> > If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable.
> 
> Your issue only needs normal syntax and should be fixed in EasyList.
> You can open an issue here: https://github.com/easylist/easylist/issues

I just discovered that the sidebar menu breaks without the following procedural cosmetic filter:
`bulbagarden.net###globalWrapper:style(position: initial !important)`

PR #8274, opened 2020-11-25
bulbagarden.net procedural cosmetic filters

shirt-dev:
```

https://bulbapedia.bulbagarden.net/wiki/Main_Page
https://bulbanews.bulbagarden.net/wiki/Main_Page
```

### Describe the issue

Bulbagarden created their own ad servers to insert ads that are not present while logged in. This pull request mitigates the issues. It also removes the top menu which also is not present while logged in.

Furthermore, without the procedural cosmetic filter added, the sidebar menu breaks and is unclickable.

### Screenshot(s)

image
image

### Versions

- Browser/version: Chrome 87

### Notes

Without the procedural cosmetic filter added, the menu breaks and is unclickable.

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -6273,3 +6273,9 @@ bollyholic.pw##div[class][style="float: none; margin:10px 0 10px 0; text-align:c
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/68391
 tecnomusic-evolution.com##+js(nosiif, visibility, 1000)
+
+! bulbagarden.net ads
+||ahost.bulbagarden.net
+||bulba-ad-host-website.azurewebsites.net
+bulbapedia.bulbagarden.net###navbarc > div
+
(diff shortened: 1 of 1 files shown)


okiehsch:
Your filters are breaking the intended layout of the site, which is incidentally the reason why you need a procedural filter to restore the ability to click the left sidebar menu, which **you** broke.

>  It also removes the top menu which also is not present while logged in.

So how do you create an account for that site after you blocked the "create account" button and how do you log in after you logged out seeing you also blocked the log in button.

In any case declined.

shirt-dev:
I mean the create account button is still there with the filters if you look closely.

The point of the ahost.bulbagarden.net is to inject custom styles for ads. Furthermore, the sidebar only broke because the HTML differs when logged out vs logged in.

When logged in, that top menu doesn't even exist because they aren't inserting ads.

okiehsch:
> I mean the create account button is still there with the filters if you look closely.


You are right, my bad I somehow missed them.

> The point of the ahost.bulbagarden.net is to inject custom styles for ads. Furthermore, the sidebar only broke because the HTML differs when logged out vs logged in.

There are no ads visible and blocking those requests breaks the intended layout, i.e. the site excpects the user to click the hamburger menu for a dropdown menu to appear and for the search function toe be in the top right corner.

Your filters restyle the whole site, now there is nothing wrong with that as a personal choice but I don't think it should be forced on all uBO users.

shirt-dev:
That's fair, I agree.

PR #8278, opened 2020-11-26
Fix bab issues on foxella.com

ryanbr:

BlockAdBlock used on `https://www.foxella.com/`  This commit should fix it

### Versions

- Browser/version: Brave


okiehsch:
Thx.

PR #8329, opened 2020-12-05
Fix Anti-adblock on livemint.com

ryanbr:

`https://www.livemint.com` updated the anti-adblock

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: Brave
- uBlock Origin version: 1.32.2

### Settings

### Notes


uBlock-user:
@ryanbr  try `livemint.com##+js(aeld, error)`

PR #8331, opened 2020-12-06
fix shop.forster.at being unclickable and blurred

haansn08:

* https://shop.forster.at/Verkehrszeichen/Verkehrszeichen-und-Schilder/Hinweiszeichen/53-17a-Ortstafel.html

### Describe the issue

Site is unclickable and blurred

### Screenshot(s)

20201206_03h16m18s_grim

### Versions

- Browser/version: firefox-83.0
- uBlock Origin version: 1.31.0


okiehsch:
We only fix issue that can be reproduced with a default setup in the unbreak list.
I assume you use` Fanboy's Annoyance List`, that list is not enabled by default.

PR #8481, opened 2021-01-18
Fixes GDPR warning on https://www.politico.eu/

ryanbr:

Fixes gdpr warning on `https://www.politico.eu/`

### Describe the issue

Due to reasons, `sp-prod.net` is exempt from various sites 

### Versions

- Browser/version: Brave
- uBlock Origin version: 1.32.4


okiehsch:
See the readme.

`GDPR messages will not be addressed in uAssets.`

PR #8488, opened 2021-01-22
Block sponsored results on DDG Lite

shreyasminocha:

`https://lite.duckduckgo.com/lite/?q=test`

### Describe the issue

Some "sponsored results" are visible.

### Versions

- Browser/version: Firefox ~84
- uBlock Origin version: 1.32.4


okiehsch:
https://github.com/uBlockOrigin/uAssets/blob/master/README.md

"If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable.

Issues involving

- anti-adblock
- ad-reinsertion
- popups/popunders
- site breakage
- video ads

will always be fixed in uAssets even if they don't require extended syntax."

Issue #8550, opened 2021-02-09
ads


eisamstiel:
url: https://www.elitepvpers.com/forum/
![Screenshot_2021-02-09 elitepvpers](https://user-images.githubusercontent.com/78649417/107389727-4f9ba980-6af7-11eb-8085-eb3e5755f6e7.png)

PIZZA90:
this can be "dangerous" for someone who does not have much idea?:

![Captura1](https://user-images.githubusercontent.com/33556647/107410027-10785300-6b0d-11eb-81b7-315f3865bc55.PNG)

**using uBO picker:**  `||*.* `

mapx-:
@gorhill

Alex4567890:
> 
> 
> this can be "dangerous" for someone who does not have much idea?:
> 
It can break the site, but you can go to the settings - > "My filters", delete the filter and save the changes.
Also, watch out if you're planning on using the element blocker while being in "My filters", because if you save any modification without refreshing, you will lose whatever you did with the content picker or maybe not anymore)
You can also see what the element picker highlights.

gorhill:
Yeah I should just drop filters which have at least one wildcard in the base domain part. I will transfer the issue to uBO.

Alex4567890:
> It can break the site, but you can go to the settings - > "My filters", delete the filter and save the changes.
If you think that your page may break, you can backup anything that you may have entered.

PR #8553, opened 2021-02-09
#8550 possible fix

Alex4567890:
If there is any better way to block the ads, please create a pull request with your filters.
The report of the ads: https://github.com/uBlockOrigin/uAssets/issues/8550


Alex4567890:
If there is any better way to block the ads, please create a pull request with your filters.

PR #8559, opened 2021-02-11
ad space

Alex4567890:


mapx-:
read here: https://github.com/uBlockOrigin/uAssets/blob/master/README.md

**If an issue can be fixed without using extended syntax, i.e. scriptlet injection, redirect, procedural cosmetic filters, etc. the issue will, as a rule of thumb, not be fixed by uAssets, because fixing it in EasyList if possible is preferable.**

Issues involving

anti-adblock
ad-reinsertion
popups/popunders
site breakage
video ads
will always be fixed in uAssets even if they don't require extended syntax.

This means you should report such issues at easylist repository:
https://github.com/easylist/easylist/issues

PR #8562, opened 2021-02-11
useless filter

Alex4567890:


mapx-:
still useful

image

Alex4567890:
> 
> 
> still useful
> 
> image

I'm pretty sure it's not about the ad blocker.

Alex4567890:
Actually, you're right. That thing does not appear with uBlock Origin.

uBlock-user:
When the fix for anti-adblock is added, everything else is also fixed together.

PR #8571, opened 2021-02-13
Block ads on scnlog.me

moriakaice:

`https://scnlog.me/`

### Describe the issue

Ads weren't filtered on the page, resulting with new window being opened.

### Versions

- Browser/version: Chrome 88.0.4324.150
- uBlock Origin version: 1.33.2

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1101,3 +1101,8 @@ tyla.com##div[class$="-margin-Advert"]:style(height:1px !important;width:1px !im
 clickscoin.com##+js(nosiif, visibility, 1000)
 clickscoin.com##+js(aopr, adBlockDetected)
 ||ybytexer.xyz^
+
+scnlog.me##^script:has-text('shift')
+scnlog.me##+js(aopw, _pop)
+*$3p,script,denyallow=google.com|gstatic.com|recaptcha.net|hcaptcha.com,domain=scnlog.me
+||gkirfipavwra.com^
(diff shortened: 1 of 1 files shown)


uBlock-user:
@moriakaice Does `scnlog.me##+js(acis, atob, 'shift')` work on Chrome ?

moriakaice:
> @moriakaice Does `scnlog.me##+js(acis, atob, 'shift')` work on Chrome ?
It does, at least for me.

moriakaice:
```
scnlog.me##+js(acis, atob, 'shift')
||gkirfipavwra.com^
```

Looks to be good enough for the page as well (the domain is yet another random characters domain used by the same ad network; that block doesn't do much for `scnlog.me` after the first line, but can save people elsewhere).

uBlock-user:
Yes, because `scnlog.me##^script:has-text('shift')` will not work for Chromium/Chrome, so having a scriptlet would rather be preferable.

moriakaice:
Cool, I had no idea. The original code did block it for me in Chrome, but maybe it was the 3rd party filter.

Do you want me to update the PR code?

uBlock-user:
Yes, these filters are adequate --

```
scnlog.me##+js(acis, atob, 'shift')
*$3p,script,denyallow=google.com|gstatic.com|recaptcha.net|hcaptcha.com,domain=scnlog.me
||gkirfipavwra.com^
```

Also add the mention of the domain, like this --

`! scnlog,me popups/ads`

at the top.

PR #8604, opened 2021-02-21
Obsolete filter (lcpdfr.com annoyance) (I think)

Alex4567890:
It does not work anymore. The new fix may only be temporary, I do not know.


okiehsch:
Thx.

PR #8711, opened 2021-03-13
Websites seem to be down

Nick754:


mapx-:
Which sites ?

Nick754:
I had them in a txt file but deleted it thinking that they would be highlighted in the commit.

mapx-:
```
layarperak.tk
sunnyfaucet.tk
statenaukari.in
```

PR #8730, opened 2021-03-17
added a filter for ovagames.com

IThinkIAmCool:


IThinkIAmCool:
It's for a popup or whatever that was being closed but I figured that it was annoying.

uBlock-user:
third-party crap, domain will be blocked instead.

PR #8770, opened 2021-03-23
fix https://github.com/AdguardTeam/AdguardFilters/issues/78227

IThinkIAmCool:


IThinkIAmCool:
and for a couple more domains/websites

mapx-:
What are you doing here ? => delete filters-2020.txt ?

PR #8826, opened 2021-04-01
Block counting copy-paste stackoverflow

paledega:
image
This dialog track my keyboard event so we must block **cdn.sstatic.net/Js/stub.en.js**

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -23306,3 +23306,6 @@ techdracula.com##.adsbygoogle
 ! As of 2020-04-09, new filters will be added to year-based sublists
 !#include filters-2020.txt
 !#include filters-2021.txt
+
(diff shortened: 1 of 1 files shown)


uBlock-user:
Filter -- `stackoverflow.com##+js(aeld, copy, HTMLElement)`

okiehsch:
The second logged "copy" event is sufficient enough on my end.

uBlock-user:
yeah mine only removed the second notice that appeared after you used all your "free copy" events, that will do.

lassekongo83:
This is an April fools joke. It will not be required tomorrow. `https://www.reddit.com/r/ProgrammerHumor/comments/mhjwyh/this_got_me_very_angry_for_a_few_seconds/`

Issue #5771, opened 2019-06-06
Arc uses bandwidth without opt-in


wh0:
### URL(s) where the issue occurs

`https://arc.io/product/`

This is a page about the script, which itself uses the script.

`https://hypem.com/popular`

### Describe the issue

Arc uses bandwidth and isn't opt-in. From the description (emphasis added):

> While on this website, Arc unites a little unused **bandwidth from your device** with millions of others to form a people-powered cloud network. Companies rent this community network and proceeds reward users and sustain sites, like Arc.io.
> 
> **Without doing** or giving up **anything** (including your privacy, personal data, or device performance), **users crowdfund** their favorite websites with spare **bandwidth**, not their wallets. And a portion goes to open Internet advocates, like Wikipedia.
> 
> **By visiting Arc-powered sites**, you help shape a better Internet… with less advertising, more privacy, and faster websites.
> 
> Arc.io welcomes you to the community and thanks you for your support! 

### Screenshot(s)

n/a

### Versions

- Browser/version: Firefox 67.0
- uBlock Origin version: v1.19.6

### Settings

n/a

### Notes

Their installation instructions say to add this script to a page:

```html
<script async src="https://arc.io/widget.js?[YOUR_WIDGET_ID]&cdn=False"></script>
```

or this script:

```html
<script async src="https://arc.io/widget.js?[YOUR_WIDGET_ID]&cdn=True"></script>
```

Also, I found a service worker in my browser with the script https://arc.io/arc-sw-loader.js

Their opt-out page: https://arc.io/optOut/

eplayerclick:
You do not want ads, you do not want anything, you just want people to work and give you free content.

PR #8874, opened 2021-04-13
Remove arc.io from the resource abuse list. They switched from default opt in to require explicit user permission before using bandwidth.

gruns:

`https://exploreatlas.co.uk/`
`https://arc.io/`

### Describe the issue

Hi guys! :wave:

I'm Ansgar. We (`https://arc.io/team`) build Arc (`https://arc.io/`), a
peer-to-peer bandwidth exchange and delivery network (CDN) that runs in
the browser.

Previously, Arc automatically connected to Arc's network and both
contributed/uploaded content to, and received/downloaded content from,
Arc's peer-to-peer network. No user action or permission was required.
And it was for this automatic, default behavior that Arc was added to
uBlock's resource abuse list.

However, six months months ago we changed Arc's behavior such that, in
the presence of adblock, Arc **requires user permission** and **doesn't
use any bandwidth** until the user explicitly opts in.

I created a page that showcases what this looks like here:

  * `https://arc.io/adblockDemo/`

So, what changed? Arc now

  1. Requests and requires explicit user opt in.

and

  2. No content is shared, or bandwidth used, until the user opts in.

And a user's opt in isn't permanent, either, of course. After opting in,
users can always later opt out through the widget, too. And Arc's widget
is always visible and interactible in the lower left corner of the page.

Since Arc no longer uses bandwidth without explicit user opt in, I
created this PR to remove Arc from uBlock's resource abuse list.

Thank you!

### Screenshot(s)

`http://i.imgur.com/8OI3XzA.png`

### Versions

- Browser/version: Chrome v87.0.4280.88
- uBlock Origin version: v1.34.0

### Settings

- Removed `arc.io` from uBlock's resource abuse list.

### Notes

Referencing Issue #5771.

### DIFF
--- a/filters/resource-abuse.txt
+++ b/filters/resource-abuse.txt
@@ -11,8 +11,6 @@
 
 ! https://github.com/uBlockOrigin/uAssets/issues/659
 /edgemesh.*.js$script,domain=~edgemesh.com|~edgeno.de
-! https://github.com/uBlockOrigin/uAssets/issues/5771
-||arc.io^$3p
 
 ! https://github.com/uBlockOrigin/uAssets/issues/803
 ||safelinkconverter.com^$script,3p


gruns:
Referencing Issue https://github.com/uBlockOrigin/uAssets/issues/5771.

uBlock-user:
> In the presence of adblock

How does Arc know about presense of adblock ?

gruns:
> How does Arc know about presense of adblock ?

With DOM bait. Similarly to how https://github.com/wmcmurray/just-detect-adblock/blob/master/src/helpers.js does it.

Full code:

```javascript
export async function isAdblockDetected () {
    let detected = false

    // Create the bait.
    const bait = document.createElement('div')
    bait.setAttribute(
        'class', `text-ad ad-text adSense adBlock adContent adBanner`)
    bait.setAttribute('style', `width: 1px ! important; height: 1px !important;
        position: absolute !important; left: -10000px !important;
        top: -1000px !important;`)
    document.body.appendChild(bait)

    // Give time for extensions to block.
    await sleep(100)

    // Check if the bait has been affected by an adblocker.
    if (window.document.body.getAttribute('abp') !== null ||
        bait.offsetParent === null ||
        bait.offsetHeight === 0 ||
        bait.offsetLeft === 0 ||
        bait.offsetTop === 0 ||
        bait.offsetWidth === 0 ||
        bait.clientHeight === 0 ||
        bait.clientWidth === 0) {
        detected = true
    } else if (window.getComputedStyle !== undefined) {
        const baitTemp = window.getComputedStyle(bait, null)
        if (baitTemp && (baitTemp.getPropertyValue('display') === 'none' ||
            baitTemp.getPropertyValue('visibility') === 'hidden')) {
            detected = true
        }
    }

    // Destroy the bait.
    document.body.removeChild(bait)

    return detected
}
```

uBlock-user:
> Requests and requires explicit user opt in.

I don't see any opt-in on that site, only opt-out.

uBlock-user:
> Requests and requires explicit user opt in.

Once opt-in and then opted out, I don't see any opt-in on that site again, only opt-out.

gruns:
> Once opt-in and then opted out, I don't see any opt-in on that site again, only opt-out.

Hrm. I can't reproduce this behavior.

Which browser and version are you using? And any other browser extensions of note?

Here's a video showing the alternating Opt In and Opt Out behavior I see: https://streamable.com/4ge1xg. Chrome v87.0.4280.88 with uBlock v1.34.0.

Steps enumerated in the video.
1) Load arc.io (or any site with Arc).
2) Open the widget.
3) Click `Enable Arc and support <site>` to opt in.
4) After the widget closes, re-open the widget.
5) Click `OPT OUT` + `Opt out` on the next screen to opt out.
6) After the widget closes, re-open the widget.
7) The widget remembers that you previously opted out: `You previously elected to opt out of Arc...` and presents an `Opt in to Arc` button.

Is this the behavior you see?

uBlock-user:
Firefox/Nightly, the only way it remembers in my case is if I create an exception filter to `tracker.arc.io`(blocked by EasyPrivacy)

gruns:
> The issue occurs after I refresh the page and open the widget

Attempting to track this down (terrible pun).

Just verified: I see the exact same behavior as in https://streamable.com/4ge1xg with `tracker.arc.io` blocked by EasyPrivacy (as part of uBlock). That is, even with `tracker.arc.io` blocked via the EasyPrivacy list, I still see the widget correctly alternating between opt in and opt out UI states.

Downloading Firefox/Nightly to attempt to reproduce the behavior what you're seeing.

gruns:
Strange. I can't reproduce in Firefox/Nightly, either.

I followed these steps:

1) Install Firefox/Nightly.
2) Install uBlock Origin.
3) Open uBlock Origin settings and disable `uBlock filters - Resource abuse` so Arc loads, as `arc.io` is currently blocked.

Then complete the prior steps from above:

1) Load `arc.io` (or any site with Arc).
2) Open the widget.
3) Click `Enable Arc and support <site>` to opt in.
4) After the widget closes, re-open the widget.
5) Click `OPT OUT` + `Opt out` on the next screen to opt out.
6) After the widget closes, re-open the widget.
7) The widget remembers that you previously opted out: `You previously elected to opt out of Arc...` and presents an `Opt in to Arc` button.

Here's a video of all of the above steps: open a fresh Firefox/Nightly, install uBlock Origin, and test Arc: https://streamable.com/2atfcj.

okiehsch:
@gruns it works as expected on my end. I asked `gorhill` what he thinks about it.

What I find problematic is that you made it opt in only if you detect adblock usage by the user.
Looking at your detection script that means if a site uses `arc` and we have a `generichide` filter for that site they will automatically connect to your peer-to-peer network of we remove the filter as you suggest.

gruns:
> @gruns it works as expected on my end. I asked gorhill what he thinks about it.

:tada: Great!

> Looking at your detection script that means if a site uses arc and we have a generichide filter for that site they will automatically connect to your peer-to-peer network if we remove the filter as you suggest.

Aha. Ya. I doubt the above detection is robust in the face of `generichide`. Excellent point.

What's a robust method to detect uBlock, even with a `generichide` filter?

I'm all ears here and happy to implement.

I'll also do some Googling and search GitHub to see what I can find.

uBlock-user:
> What's a robust method to detect uBlock, even with a generichide filter?

How about not detecting at all ? Make it unconditional like `authedmine.com` did. These baits are going to get removed and we fix anti-adblock issues everyday fyi.

gorhill:
> Looking at your detection script that means if a site uses arc and we have a `generichide` filter for that site they will automatically connect to your peer-to-peer network if we remove the filter as you suggest.

Yes, should definitely not use cosmetic-related filter, the same issue would arise if someone choose to disable cosmetic filtering using the no-cosmetic-filtering switch. To me it seems best to just detect whether `tracker.arc.io` is blocked or not.

gruns:
> To me it seems best to just detect whether tracker.arc.io is blocked or not.

Sick. This is now implemented and live!

The new adblock detection function that probes `tracker.arc.io`:

```javascript
async function detectAdblockViaRequestBait () {
    let detected = false

    // Ideally enough time for an adblock to block the request but not
    // enough time for the network connection to be established.
    const timeout = 15 // Milliseconds.
    const baitUrl = 'https://tracker.arc.io'
    try {
        const resp = await wfetch(baitUrl, { timeout })
    } catch (err) {
        if (err.name === 'TypeError') { // TypeError: Failed to fetch.
            detected = true // Blocked by adblock.
        } else { // AbortError, or other error, means not blocked.
            detected = false
        }
    }

    return detected
}
```

In action:

  * Blocked: http://i.imgur.com/7WL3a66.png

  * Unblocked: http://i.imgur.com/YO7OxRR.png

Arc now:

  1) Detects adblock with a probe for `tracker.arc.io`.

  2) Requires explicit user permission.

  3) After opting in, users can always easily opt out, too.

Anything left to resolve or change to merge this PR?

Thank you!

BottledSoda:
What is `tkr.arc.io` for ?

Yuki2718:
@gruns just FYI: https://github.com/easylist/easylist/commit/e80cdbd17653e36eec8242eb5e35d9c19fbb0e68

gruns:
@Yuki2718 thank you for the mention and link to the commit! 🙏

PR #8928, opened 2021-04-19
update adshares

ghajini:

many sites,
breakage recently observed at
`https://linkrit.com/view/D4cew1z7MM`

i will able to provide where filter is useful but its good blocking their domains instead of scriptlets filter

### Describe the issue

unlock links unclicakble
issue is better handled by easylist compatible blocking filters &routinely taken care by ublock filters

### Screenshot(s)

### Versions

- Browser/version: firefox a nightly
- uBlock Origin version: 1.34

### Settings

-  uBO's default settings


okiehsch:
Thx.

Issue #9001, opened 2021-05-02
scam website


IThinkIAmCool:
proof:

![Untitled](https://user-images.githubusercontent.com/83162792/116827333-23953f00-aba1-11eb-8f61-7ad3d706176b.png)
![Untitled2](https://user-images.githubusercontent.com/83162792/116827336-26902f80-aba1-11eb-93b9-f2be5f1cabab.png)

IThinkIAmCool:
It's already been added, sorry if I wasted your time.

PR #9002, opened 2021-05-02
address https://github.com/uBlockOrigin/uAssets/issues/9001

IThinkIAmCool:


IThinkIAmCool:
It has already been added.

PR #9061, opened 2021-05-09
update fix for netcine.biz and pop-up(s)

IThinkIAmCool:


uBlock-user:
I do not any experience anti-adblock acting with the existing fix

Secondly, where do you see the popups ?

uBlock-user:
I do not experience any anti-adblock acting with the existing fix

Secondly, where do you see the popups ?

IThinkIAmCool:
I didn't know it has been fixed since I hadn't updated uBO's 'uBlock filters'.
I saw the pop-ups before updating uBO's 'uBlock filters'.

IThinkIAmCool:
Could the timer be bypassed or whatever?

okiehsch:
No,  the timer is validated server side.
Add
`decorardicas.com.br,loucasporcabelos.com.br##+js(nano-stb)`
and you will see that the timer is boosted but the video will only appear after the timer has reached 0 on the server side.

PR #9148, opened 2021-05-22
obsolete (I think)

IThinkIAmCool:


mapx-:
test case like that 1 in the original issue

IThinkIAmCool:
@mapx- I didn't encounter any anti adb after adding `bitcolife.ru#@#+js(nobab)` and refreshing the page. Neither on `https://bitcolife.ru/`

PR #9178, opened 2021-05-24
obsolete

IThinkIAmCool:


IThinkIAmCool:
I did not encounter any anti adb mechanism without having ghide. I think the cosmetic filters should be applied, if needed, in EasyList.

okiehsch:
EasyList does not fix anti adblock issues.

IThinkIAmCool:
I know that. I mean that if by the removal of those filter there would be some issues (like ad placeholders) they should be fixed in EasyList.
Is there any anti adb mechanism in place after removing those filters?

PR #9194, opened 2021-05-25
Update ISSUE_TEMPLATE.md

IThinkIAmCool:


IThinkIAmCool:
I think I deleted an older branch of the fork(s) I made of uAssets.

llacb47:
It already has 
```
 I tried to reproduce the issue when...
 uBlock Origin is the only extension
```

PR #9223, opened 2021-05-27
update https://github.com/uBlockOrigin/uAssets/issues/1035

IThinkIAmCool:


IThinkIAmCool:
I meant to use "update" instead of "fix"

PR #9261, opened 2021-05-29
update https://github.com/jspenguin2017/uBlockProtector/issues/1048

IThinkIAmCool:


IThinkIAmCool:
Actually might be a preview.

PR #9317, opened 2021-05-30
obsolete filters

IThinkIAmCool:


okiehsch:
Site still uses blockadblock. I will remove the fingerprint filter.

PR #9337, opened 2021-05-31
blocks ad

IThinkIAmCool:


mapx-:
to be added in EL

however, maybe a better filter

`##[href*="/aff_c"]`

@ryanbr

PR #9474, opened 2021-06-29
games-database.com countdown speed-up

IThinkIAmCool:
Link I found on the internet: `https://games-database.com/7bt` (not mine - I am not responsible for your actions - I don't know if there is any malware)


uBlock-user:
link is 404

mapx-:
it's not 404 but the download wont start ever using the filter

PR #9526, opened 2021-07-02
addresses https://www.reddit.com/r/uBlockOrigin/comments/oagwfe/roblox_sponsored_ads_on_search_results/

IThinkIAmCool:
I already reported this to EasyList -- https://github.com/easylist/easylist/issues/8267

Example of an ad encountered after searching "tycon" on Roblox.com while being logged-in: 
image

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -4180,3 +4180,6 @@ igggames.unblocker.pro##.adstg
 
 ! https://github.com/uBlockOrigin/uAssets/pull/9517
 skidrow-games.com##+js(acis, JSON.parse, break;case $.)
+
+! https://www.reddit.com/r/uBlockOrigin/comments/oagwfe/roblox_sponsored_ads_on_search_results/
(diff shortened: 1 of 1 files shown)


uBlock-user:
Neither OP, nor you are posting the URL....

IThinkIAmCool:
@uBlock-user Do you have a Roblox account?

IThinkIAmCool:
If you have an account and you're logged in it:
image
^ You can search using this search bar.

IThinkIAmCool:
image

IThinkIAmCool:
image

IThinkIAmCool:
Actually, the filter is not a good fix since if some games were to have "Sponsored Ad" in their name, they would get hidden.

IThinkIAmCool:
The filter from `https://www.reddit.com/r/uBlockOrigin/comments/oagwfe/roblox_sponsored_ads_on_search_results/h3ihtei?utm_source=share&utm_medium=web2x&context=3` seems to be working.

PR #9561, opened 2021-07-07
Bug report form

Madis0:
As you liked the one for uBO, I made one for uAssets as well.

By the wording it seemed like not all fields are mandatory here, so I made some of them optional.

The PR also adds the links for support and uBO issue tracker in the issue creation view.

[How creation looks](https://github.com/Madis0/uAssets/issues/new/choose)

[How the result looks](https://github.com/Madis0/uAssets/issues/2)


uBlock-user:
> so I made some of them optional.

Which ones ?

Madis0:
> > so I made some of them optional.
> 
> Which ones ?

- Screenshot(s)
- Settings
- Notes

uBlock-user:
Settings must not be optional, lots of reporters do not use default settings and fail to provide that info.

uBlock-user:
Thanks!

krystian3w:
Form have backdoor:

https://github.com/uBlockOrigin/uAssets/issues/new?title=-2&template=bug_report.yml&url_s_where_the_issue_occurs=-1&describe_the_issue=0&screenshot_s=1&ublock_origin_version=2&browser_name_and_version=3&settings=4&notes=5

Maybe possible auto mark areas with prepare link.

Madis0:
Well, of course you can fill the fields with random data, how is that a backdoor? @krystian3w

krystian3w:
I suppose some of Internet found option to mark check list without read:

`ubo_is_the_only_extension=checked` ...

uBlock-user:
@krystian3w Report to Github

Madis0:
Well yeah, I suppose it was added to be used for useful applications, such as having a link in uBO settings that would fill the version number out automatically. 

But if someone starts to use it to spam, the managers could simply block the user, thats not very different from using a fully text-based form in a similar way.

krystian3w:
I would rather **madis0** report this to @Microsoft to add the option, "`no-allow-use-links-with-paramters-with-values`" in the yml file.

Madis0:
I'm sorry, but I just don't see any security issue here. What's the difference whether the user uses a specialized URL, a macro or some other method... If there is abuse, the user can be banned.

krystian3w:
I prefere block system methods before banish many users.

PR #9589, opened 2021-07-12
Update filters-2021.txt

Zerebokep:

```
https://www.elitepvpers.com/forum/sro-pserver-advertising/4919044-electus-poseidon-new-server-100-cap-eu-ch-grand-opening-25-06-2021-a.html
https://www.elitepvpers.com/forum/metin2-pserver-guides-strategies/4912938-release-new-armor-costume-auorakingdom-hairstyle-sash-dane.html
https://www.elitepvpers.com/forum/general-gaming-discussion/4904809-outriders-1-04-trainers-drop-hack-more-bypass.html
https://www.elitepvpers.com/forum/fortnite-trading/4902375-fortnite-hacks-guide-neu.html
```

Just a few of them, you'll find 100x more through the search.

### Describe the issue

Way too much thread owners use link shorteners to track their performance (mainly bit.ly), users want to see those links if they visit a private server presentation or a trading thread.

To reproduce the issue you need to be logged in and to have the link protection disabled (https://www.elitepvpers.com/forum/profile.php?do=editoptions).

### Screenshot(s)

https://imgur.com/a/peRc1Qt

### Versions

Browser/version: 91.0.4472.77
uBlock Origin version: v1.36.0

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1657,7 +1657,10 @@ elitepvpers.com##div[style="background: #ededed;margin:0"]
 elitepvpers.com##a[href][target="_blank"][rel="nofollow pass"]
 *$frame,3p,denyallow=facebook.com|google.com|hcaptcha.com|recaptcha.net|twitter.com|youtube.com|api.paymentwall.com|streamable.com|vimeo.com,domain=elitepvpers.com
 !elitepvpers.com##[target="_blank"] [src$=".gif"]
+elitepvpers.com##div[style="font-size:11px;margin-bottom: 5px;"]:has-text(/Advertise/i) + a[href]
 elitepvpers.com##[target="_blank"] [src^="https://aimex.xyz/"]
+elitepvpers.com##.cwcontent [src*=".gif"]
+elitepvpers.com###aboveheader [src*=".gif"]
 
 ! https://github.com/uBlockOrigin/uAssets/issues/8706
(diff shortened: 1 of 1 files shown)


uBlock-user:
Those filters block ads, removing them will make ads re-appear. What do you suggest for that ?

Zerebokep:
I suggest to be more specific like the existing rules for that site, e.g. `elitepvpers.com##div > .page > div > div[style]:has-text(/^Advertise/i)`, the one in question hurts sellers (myself too) because customers can't find our links and we always have to explain them to disable their adblockers, so this doesn't make any sense.

Maybe target the wrapper around the ads instead to filter any link shortener globally?

`<div align="center" class="cwcontent" style="padding:0">`

uBlock-user:
> to be more specific 

We used to be specific, but the site admin circumvents our filters, read my other post.

uBlock-user:
> e.g. elitepvpers.com##div > .page > div > div[style]:has-text(/^Advertise/i)

That's already present in the main list I'm asking for a replacement for the filters you suggested to remove.

Zerebokep:
Maybe you can apply the filters only to elements which are not inside the posts wrapper?

`<table id="post38785359" class="tborder" cellpadding="6" cellspacing="0" border="0" width="100%" align="center">`

I never created a ublock filter by myself, but with regex I would match that one like `id="post\d+"` with a negative lookahead.

PR #9609, opened 2021-07-21
Update filters-2021.txt

Zerebokep:
`
https://www.elitepvpers.com/forum/sro-pserver-advertising/4884475-asatrul-l-only-ch-l-right-now-cap-80-progressive-cap-l-low-rates-l-experience-labor.html`

### Describe the issue

The filter in question breaks private server and trade thread presentations, as most of them contain linked images.

### Screenshot(s)

https://imgur.com/a/CyxQM6x

### Versions

- Browser/version: 91.0.4472.77
- uBlock Origin version: 1.36.2

### Note

It also breaks site functionality as described in my last issues.

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1670,6 +1670,9 @@ elitepvpers.com##.cwcontent [src*=".gif"]
 elitepvpers.com###aboveheader [src*=".gif"]
 elitepvpers.com##[rel="nofollow bypass"]
 elitepvpers.com##.cwcontent [target="_blank"]
+elitepvpers.com##[target="_blank"] [src*=".png"]
+elitepvpers.com##a[target="_blank"][rel^="nofollow"]:has(img):not([href*="trello.com"])
+elitepvpers.com##p > [href^="https://playorigin.com/"]
 
 ! https://github.com/uBlockOrigin/uAssets/issues/8706
 send.cm##+js(aopw, Fingerprint2)
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1670,8 +1670,6 @@ elitepvpers.com##.cwcontent [src*=".gif"]
 elitepvpers.com###aboveheader [src*=".gif"]
 elitepvpers.com##[rel="nofollow bypass"]
 elitepvpers.com##.cwcontent [target="_blank"]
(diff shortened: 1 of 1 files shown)


okiehsch:
Can this be closed?

Zerebokep:
The filters which I removed are still causing the mentioned issues, the site is basically unusable with those filters and the last easylist updates.

okiehsch:
````
elitepvpers.com##[target="_blank"] [src*=".png"]
elitepvpers.com##a[target="_blank"][rel^="nofollow"]:has(img):not([href*="trello.com"])
````
Those two?

Zerebokep:
Yes 👍

okiehsch:
https://github.com/uBlockOrigin/uAssets/commit/e83ade614cd26c00e2a60890a8ff38b32e877758

JobcenterTycoon:
does `elitepvpers.com##a[target="_blank"][rel]:has(img:not([src^="https://www.elitepvpers.com/forum/attachments/"])):not([href*="trello.com"])` works?

Zerebokep:
I don't think so, TBM and elite gold icons didn't work with that rule too. Imo that rule is way too broad to start with.

JobcenterTycoon:
ok now more spezific (im disabled all bad easylist filter to test):

`elitepvpers.com##div > .page > div > div[style]:has-text(/^Advertise/i) + a` (based on a existing filter)
`elitepvpers.com##table > tbody > tr > td[id="aboveheader"] > div > p > a[target="_blank"][rel]:has(img)`

problem is its they will change again so this filter will be useless in a few days like all another one. When this not work without breakage im done with this  ...

Zerebokep:
I understand your point, but all rules which affect the threads will directly harm most sellers and private server administrators on elitepvpers. Because of that new easylist filter trading is currently impossible which puts customers with an active adblock at risk getting scammed and we traders aren't able to get positive ratings without telling them to disable their adblock. Most trading discords have now put up a warning to disable adblocks on the site.

It baffles me how such a filter as the easylist one gets approved, it basically removes every linked image.

JobcenterTycoon:
https://github.com/easylist/easylist/commit/f3c9228d31d95b9ab30abe27416917681a932ea3#commitcomment-53849955

Looks like they want to remove the ads for a longer time so they use too generic filter

PR #9636, opened 2021-07-26
Adblock nag on afterdawn.com and alternativeto.net

peace2000:

`https://www.afterdawn.com/`
`https://alternativeto.net/`

### Describe the issue

These sites have started to use adblock nag

### Screenshot(s)

kuva

kuva

### Versions

Firefox 90.0.2 
Ubo 1.37.1b1

### Settings

Default

### Notes

Nag seems to appear due to this is filter: `||fundingchoicesmessages.google.com^$3p`. I wonder if there were an universal solution to counter that for all domains that try to connect to that address? I hardly believe those two are the only ones using Google's funding choice.


okiehsch:
Can reproduce in Firefox, works fine with Chromium.

PR #9637, opened 2021-07-26
Adblock nag on high.fi, download.fi, muropaketti.com

peace2000:

`https://high.fi/`
`https://www.download.fi/`
`https://muropaketti.com/`

### Describe the issue

The same as: https://github.com/uBlockOrigin/uAssets/pull/9636

### Screenshot(s)

The same as: https://github.com/uBlockOrigin/uAssets/pull/9636

### Versions

Firefox 90.0.2
Ubo 1.37.1b1

### Settings

Default

### Notes

The same as https://github.com/uBlockOrigin/uAssets/pull/9636.

Could there be some kind of universal solution to counter this on all websites? I believe this is even more widely used.

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -4156,7 +4156,7 @@ hoodsite.com##+js(aopw, _pop)
 
 ! Brave and Firefox issue google funding
 ! https://github.com/uBlockOrigin/uAssets/pull/9636
-afterdawn.com,alternativeto.net,globo.com,high.fi,latimes.com,muropaketti.com##+js(nostif, f.parentNode.removeChild(f), 100)
+afterdawn.com,alternativeto.net,download.fi,globo.com,high.fi,latimes.com,muropaketti.com##+js(nostif, f.parentNode.removeChild(f), 100)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/7970
 avimobilemovies.net,hackiee.mobi##+js(nowebrtc)
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -4156,7 +4156,7 @@ hoodsite.com##+js(aopw, _pop)
 
 ! Brave and Firefox issue google funding
 ! https://github.com/uBlockOrigin/uAssets/pull/9636
-afterdawn.com,alternativeto.net,globo.com,latimes.com##+js(nostif, f.parentNode.removeChild(f), 100)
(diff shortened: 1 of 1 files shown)


okiehsch:
Sure, you can add 
`*##+js(nostif, f.parentNode.removeChild(f), 100)`
that injects that scriptlet globally, we don't add global injection filters to our lists.
Thanks for the report.

uBlock-user:
`##[class^="div-gpt-ad"]` is getting applied even though `#@#[class^="div-gpt-ad"]` is present, something broke in uBO.

uBlock-user:
@okiehsch can you reproduce on stable uBO ?

peace2000:
> 
> 
> Sure, you can add
> `*##+js(nostif, f.parentNode.removeChild(f), 100)`
> that injects that scriptlet globally, we don't add global injection filters to our lists.
> Thanks for the report.

Yeah, I can do that but I was thinking everyone else who uses Ubo as well. That nag is caused by blocking `fundingchoicesmessages.google.com` so I was just wondering if there were a way to fix this universally, without using a global +js filter. Just thinking, I don't know what's possible and what's not. I tried to use resource redirectors but they didn't help...

uBlock-user:
@peace2000 do you see anti-adblock on stable uBO ?

okiehsch:
@uBlock-user You are right It's a uBO bug, I can't reproduce with the stable version.

uBlock-user:
Broken by https://github.com/gorhill/uBlock/commit/c25938f5bcb519ee945c9ccab7d17474474dc825

uBlock-user:
Should revert both commits.

okiehsch:
@gorhill

uBlock-user:
`#@#[class^="div-gpt-ad"]` gets applied on **1.37.1b0**, thats the last working build

peace2000:
@uBlock-user Well in 1.37 that nag went away. So yeah seems to be a bug then.

uBlock-user:
@peace2000 happened with https://github.com/uBlockOrigin/uAssets/pull/9637#issuecomment-886595022

uBlock-user:
Filter count for uBlock badware list also went down from `2059` to `2019` with the latest dev build.

gorhill:
> Broken by gorhill/uBlock@c25938f

Sorry, will fix asap

gorhill:
Fixed in 1.37.1b2, and those suffering the regression will have to force an update of filter lists which were updated during 1.37.1b1.

peace2000:
@gorhill Tested, seems to work properly. Thanks!

PR #9652, opened 2021-07-27
Reduce filter expire time

peace2000:
I suggest, that this would be reduced. There could be occasions when something is broken and you want to send a fix for everyone asap.


uBlock-user:
We're not going to bombard GitHub, this was decided by gorhill to keep longer.

PR #9662, opened 2021-07-28
Redundant filter

peace2000:
kolumbus.fi is not in use anymore, and the new domain it redirects, has no issues.

Related report: https://github.com/NanoMeow/QuickReports/issues/1059


peace2000:
Sorry I was kinda blind....

PR #9663, opened 2021-07-28
Redundancies

peace2000:
These entries are not needed anymore, they are old and besides, adblocking for these domains are handled by the Finnish Easylist which is Ubo's stocklist these days.


okiehsch:
Thx.

PR #9719, opened 2021-08-06
Removed a bunch of redundancies

peace2000:
Removed filter entries that contain non-working domains.


peace2000:
It seems that Github doesn't handle comparison change properly on that one big section. This tool can compare properly:

https://countwordsfree.com/comparetexts

PR #9784, opened 2021-08-18
Badware

shenzhiming88:
The software does not work properly even after payment, cheating users of their property
See: `https://www.zhihu.com/question/21406655/answer/658774547`

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -902,3 +902,6 @@ thepiratebay3.com^$all
 
 ! https://www.huorong.cn/info/1617368984641.html
 ||qqfzn.com^$all
+
+! https://github.com/uBlockOrigin/uAssets/pull/9784
+||arctime.org^$all


okiehsch:
>  does not work properly even after payment

That is out of scope.

uBlock-user:
Site issue, not badware.

shenzhiming88:
> > does not work properly even after payment
> 
> That is out of scope.

So, only deal with malware?

okiehsch:
And with bundled Adware or fake sites etc.
Not with a software that does not work properly.

shenzhiming88:
> And with bundled Adware or fake sites etc.
> Not with a software that does not work properly.
> Allegedly.

However, the software requires users to pay to use the function, but the function is still not available after payment, I think this is fraud

uBlock-user:
@shenzhiming88 Contact with site admin and take it up with him ?

shenzhiming88:
> @shenzhiming88 Contact with site admin and take it up with him ?

According to the description of `https://www.zhihu.com/question/21406655/answer/658774547` and `https://www.zhihu.com/question/52157612/answer/658743763`, they refused to deal with this issue and abused paying users

shenzhiming88:
@uBlock-user

uBlock-user:
So the admin is acting scummy, still out of scope of this list.

PR #9869, opened 2021-08-31
update r-c & remove obsolete filters

IMayKnow:


IMayKnow:
`rong-chang.com##.ads-top` is for `https://www.rong-chang.com/`

PR #9906, opened 2021-09-02
Remove unnecessary whitespace

mjethani:
There's unnecessary whitespace in `badware.txt` which causes an error in the [linter](https://github.com/mjethani/flint).


mjethani:
@uBlock-user thank you!

I'm going to create a separate set of rules in the linter just for uBlock Origin-specific filters.

PR #9910, opened 2021-09-02
Punycode domains

mjethani:
The domain `österreich.at` can be encoded as `xn--sterreich-z7a.at`, which makes it a _little_ more efficient and, more importantly, consistent with other filters in the same list.


okiehsch:
Thx.

PR #9923, opened 2021-09-04
Add anchor in front of protocol

mjethani:
Unless the intention is to match _anywhere_ in the URL, it's better to have a `|` before `https://`.


uBlock-user:
That was a ref case url, we're not blocking `forums.lanik.us` anywhere.

PR #9925, opened 2021-09-04
Hide outlook.live.com anti adblock nag column

peace2000:

`https://outlook.live.com/mail/0/inbox`

### Describe the issue

There's a huge anti adblock nag column in Outlook. That column consists originally ads but when they are blocked, that column changes into a anti adblock nagging column, that takes a lot of space and prevents the full width usage of screen (it prevented it before as well though).

### Screenshot(s)

kuva

After hiding you have the whole screen width available:

kuva

Sorry, a lot of censorship :D...

### Versions

- Browser/version: [Firefox 91.0.2]
- uBlock Origin version: [1.37.3b21]

### Settings

The column will be visible with default uBO settings.

### Notes

That column is currently handled through uBO Annoaynce list, but I think that as that column prevents me from using the full width to read my emails, it should be considered rather as intrusive stuff, instead of a harmless Annoyance. But whatever way you decide to go with this, the current hiding rule in Annoyance list is very slow to hide it. My filter will hide it pretty quickly. It's safe to use, I have used a similar rule to hide it for a long time without any issues. I used a language specific version, but I made now a language neutral version of my rule.

My rule I have used for the Finnish UI Outlook:

`outlook.live.com##div[aria-label=Siirtymisruutu] ~ div[data-max-width="2400"] + div[class]`

English-language specific version of my rule original rule would be:

`outlook.live.com##div[aria-label="Navigation pane"] ~ div[data-max-width="2400"] + div[class]`

Language neutral, suggested rule:
`outlook.live.com##div[class][tabindex="-1"][role=region][aria-label][data-min-width][data-max-width] ~ div[class][data-max-width="2400"] + div[class]`

DOM tree if you don't have an account on MS services and you don't want to create a test account:

kuva

The selected element is the one that I target.

### DIFF
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -6239,3 +6239,6 @@ ijr.com##+js(acis, document.createElement, admiral)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/9895
 *$image,redirect-rule=1x1.gif,domain=guardian.gg
+
+! https://github.com/uBlockOrigin/uAssets/pull/9925
+outlook.live.com##div[class][tabindex="-1"][role=region][aria-label][data-min-width][data-max-width] ~ div[class][data-max-width="2400"] + div[class]
--- a/filters/annoyances.txt
+++ b/filters/annoyances.txt
@@ -377,10 +377,6 @@ digitalsynopsis.com##+js(aopr, document.oncontextmenu)
 ! https://forums.lanik.us/viewtopic.php?f=62&t=41301
 filefox.cc##+js(aeld, blur)
 
-! https://github.com/uBlockOrigin/uAssets/issues/3948
-! https://github.com/NanoMeow/QuickReports/issues/525
-outlook.live.com##a[href="https://windows.microsoft.com/outlook/ad-free-outlook"]:upward(4)
-
 ! https://github.com/uBlockOrigin/uAssets/issues/3096 right click
 visionias.net##+js(acis, $, contextmenu)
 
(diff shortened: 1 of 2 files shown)


okiehsch:
I can't reproduce.
It is already fixed by
`outlook.live.com##a[href="https://windows.microsoft.com/outlook/ad-free-outlook"]:upward(4)`
on my end.

okiehsch:
image

okiehsch:
I agree that your rule is "quicker".

peace2000:
`[data-max-width="2400"]` is - what I understand - the max width for email message previews, you can stretch you browser window as much as you want and still that value doesn't seem to change.

uBlock-user:
We have never considered **space obstruction** as a deciding factor as to whether this goes in main list or as an annoyance for other such anti-adblock/s which take place of blocked ads, so this too should go in Annoyances list.

peace2000:
All right, I edited my pull request so that my suggestion will go to the Annoyance list instead. It replaces the current rule that is slower.

uBlock-user:
If you do have another point that I missed, then do post here.

okiehsch:
This is only an annoyance in my opinion too.

peace2000:
Well, I also want to point out that if you don't block ads in that column (disable uBO to see what it looks like actually looks like), one would consider it just as a normal ad column. If someone just hides that element without blocking ads on it, he/she would be completely unaware that it eventually turns into anti adblock nag.

@uBlock-user @okiehsch

okiehsch:
> If someone just hides that element without blocking ads on it, he/she would be completely unaware that it eventually turns into anti adblock nag.

Our policy regarding if an anti adblock message is - as a rule of thumb - if the the message blocks site access, disables site functionality or hides content of the site we fix it in the main list.
If not it is usually fixed in the annoyance list.

peace2000:
OK, thanks anyway :)

PR #9957, opened 2021-09-12
Adding a tracker

shenzhiming88:
Address: `https://wenku.baidu.com/view/fab6afb40229bd64783e0912a216147917117e2a`
This is most likely a tracker that collects information from the user's browser
image

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -300,3 +300,6 @@ ppss.kr##.ppss-title-banner:remove()
 
 ! https://github.com/uBlockOrigin/uAssets/issues/9877
 ||instiengage.com^$3p
+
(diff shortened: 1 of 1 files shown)


uBlock-user:
```
(function fp(){
  var Days=1;var exp=new Date();exp.setTime(exp.getTime()+Days*24*60*60*1000);document.cookie="kunlunFlag=1;expires="+exp.toGMTString()
})()
```

That's the js code that is executed inside the script, so I don't see how this is a tracker.

shenzhiming88:
@uBlock-user Sorry, this was most likely a tracker before, but I added it again without verifying it, I'll pay more attention next time

PR #10017, opened 2021-09-16
Address Adware

shenzhiming88:
Adware link: `https://www.flash.cn/`(VPN to China may be required to prevent redirection)
The Chinese special version of Flash has an adware program bundled with it, if this program is removed, then Flash will not work
Report: `https://www.virustotal.com/gui/file/486c2ef133ea7fb96b8ac244ee3bb54b9a0536ac713608155070a622522fcbea`
图片

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1006,3 +1006,6 @@ thepiratebay3.com^$all
 ||serialkey89.com^$doc
 ||goharpc.com^$doc
 ||free4pc.org^$doc
+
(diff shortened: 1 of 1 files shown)


darktohka:
Check out https://github.com/cleanflash/installer for an adware-less version of this software.

iam-py-test:
> Check out https://github.com/cleanflash/installer for an adware-less version of this software.

@darktohka it looks like that repo was taken down:
image

PR #10056, opened 2021-09-22
Narrow Scope of Selector to intended Site

fwebdev:

`www.eharmony.com`

### Describe the issue

The Selector ##[href*="/redirect?tid="] is potentially affecting all Pages with that kind of Link.
(And at least does affect our Page eharmony.com where Customers can not be redirect to the Profile-Page after buying because the button is missing)

### Screenshot(s)

image

### Versions

- Browser/version: [Chrome Dev 92.0.4515.159]
- uBlock Origin version: [1.37.2]

### Notes

I encapsulated the Selector to the intended Site


okiehsch:
Thx.

PR #10057, opened 2021-09-22
ldjam 49 sponsor footer

ajayyy:

`https://ldjam.com/events/ludum-dare/49`

### Describe the issue

Sponsor footer at the bottom of the event page

### Screenshot(s)

image

### Reasoning

PR to uAssets due to the use of `:has()`. This is used in case text is added below the sponsor images.


uBlock-user:
Self promotions, we don't block these.

PR #10109, opened 2021-10-01
fix typo

peace2000:
@mapx- accept this to fix critical typo. Reddit's editor removed some characters... (I copied this rule from there)


uBlock-user:
@peace2000 Should have added to annoyances.txt, we don't fix these in the main list. I didn't notice before.

peace2000:
Ok, I'll make a new PR just a sec

PR #10143, opened 2021-10-05
Add a new loadout.tf filter

Tiagoquix:
Previously, a random gray square would randomly appear to me. I think it is a leftover from the ad that was removed. Anyway, I am using this filter for more than three months and it did not cause any problems at all, so I think it is a good idea to add my filter to the global filter list.

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -6495,6 +6495,7 @@ zilinak.sk##+js(nostif, offsetHeight)
 @@||loadout.tf^$ghide
 loadout.tf##.loadout-application-advertisement-header
 loadout.tf##.adsbygoogle:upward(2):style(left-3000px !important;position:absolute !important)
+loadout.tf###aswift_0
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/71225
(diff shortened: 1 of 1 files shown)


uBlock-user:
Not seeing gray area anywhere, besure to test on default settings with all other extensions disabled.

Tiagoquix:
> Not seeing gray area anywhere, besure to test on default settings with all other extensions disabled.

This square on the left side:

image

My filter removes it.

uBlock-user:
Doesn't exist due to presense of  `loadout.tf##.adsbygoogle:upward(2):style(left-3000px !important;position:absolute !important)`

You will have to investigate why you're seeing that square.

uBlock-user:
@Tiagoquix Do you see the square on Chrome ?

Tiagoquix:
> Doesn't exist due to presense of `loadout.tf##.adsbygoogle:upward(2):style(left-3000px !important;position:absolute !important)` in uBlock filters list.
> 
> You will have to investigate why you're seeing that square.

Weird.

> @Tiagoquix Do you see the square on Chrome ?

Yes.

PR #10145, opened 2021-10-06
Addresses site breakage caused by filter from ROad-block

ThinksHeIsCool:


ThinksHeIsCool:
The issue is already reported in ROad-block via [the issue I created](https://github.com/tcptomato/ROad-Block/issues/114)

uBlock-user:
what site breakage ?

PR #10149, opened 2021-10-06
Unblocking the criteo homepage

stm-x:

`https://www.criteo.com/`

### Describe the issue

Access to criteo.com is strictly-blocked by peter lowes list.

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -4273,3 +4273,7 @@ xunta.gal#@##anuncio
 @@||network-n.com/configs/sites/the-loadout.json$xhr,domain=theloadout.com
 @@||live.primis.tech/live/live*.php$script,domain=theloadout.com
 @@||video.primis.tech$media,domain=theloadout.com
+
+! criteo
+||criteo.com^$badfilter
+||criteo.com^$3p
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -4274,6 +4274,6 @@ xunta.gal#@##anuncio
 @@||live.primis.tech/live/live*.php$script,domain=theloadout.com
 @@||video.primis.tech$media,domain=theloadout.com
 
-! criteo
+! To counter `criteo.com` in Peter Lowe's
 ||criteo.com^$badfilter
 ||criteo.com^$3p


iam-py-test:
@mapx- 
 MVPS HOSTS has the same problem (after updating all filters):
image

okiehsch:
We only unbreak sites that are broken with a default setup.
MVPS Hosts is not part of the default setup.

`||www.criteo.com^$badfilter`

iam-py-test:
> 
> 
> We only unbreak sites that are broken with a default setup. MVPS Hosts is not part of the default setup.
> 
> `||www.criteo.com^$badfilter`

I don't need it unblocked, I just figured I would alert you about the problem

spirillen:
It is blocked for very good reasons by many, it is to be classified as Adware and Tracking

- https://medium.com/nextdns/cname-cloaking-the-dangerous-disguise-of-third-party-trackers-195205dc522a
- https://mypdns.org/my-privacy-dns/matrix/-/issues/3116
- https://mypdns.org/my-privacy-dns/matrix/-/issues/3117

If you for some reasons believes the ban should be eased a bit then block it at least as `^3p`

You can just try to scrape my collecting library: https://github.com/Import-External-Sources/hosts-sources

```shell
github/Import-External-Sources/hosts-sources$ git grep '\.criteo\.com$' | cut -d ':' -f 2 | sort -u
```

I will spare you the full list here as it is rather big (704 results)

ryanbr:
Disappointing. Looking from a webcompat perspective. Its an unnecessary first party block given its 3p blocked in Easyprivacy. To disable uBO or Brave shields just opens up more trackers to the user for the sake of policy. 

You want people not to need to disable their adblock software to get a workaround, the less interaction by the user the better. "Just make it work".

Yuki2718:
User actually don't need to disable uBO on the strict-blocked page, but some user will do.

gorhill:
> the less interaction by the user the better. "Just make it work".

Entirely agree of course.

However here we are not talking about "users", we are talking about sites used to push ads/trackers on users, from uBO's perspective the _clients_ of those sites_ are tracker/ad pushers, not users. They are welcome to eat their own food by disabling uBO if it gets in the way on these specific sites.

Yuki2718:
> the clients of those sites are tracker/ad pushers, not users.

I can say that's not always the case because subdomain also blocked. The latest article on `https://note.smartnews-ads.com/` (`https://github.com/AdguardTeam/AdguardFilters/issues/95110`) is "how COVID-19 changed consumer's sense of beauty".

ryanbr:
We take a more harder line in Easylist for revolving adservers/trackers. But each request is treated on its own merits if it comes to our attention. But having a blanket rule like this seems to is an overkill. General strict block should be reserved for specific servers imo.

Yuki2718:
In relation to https://github.com/uBlockOrigin/uAssets/commit/1142bef68ee287d27f6191c433ff8491570b6612 are all visitors of these pages (many of them include ordinary blogs about marketing) really customers/clients? We already unblocked `analytics.google.com` in https://github.com/uBlockOrigin/uAssets/issues/18576 and `pagefair.com` blocking has been bothering to me personally as I occaisonary visit and reference the site for their statistics (badfilterd in My filters). It seems most of these sites don't even have click-through tracker missed by the original rules. If it's click-through tracker that matters, we can use `$doc` instead of totally breaking sites (even then I prefer actually confirming existence of such trackers). Should a univ student interested in marketing and installed uBO who doesn't know how to badfilter rules should put up with all the annoynaces or disable uBO (and accept all the tracker)? To me it only looks like we want to break these sites.

gorhill:
> `pagefair.com` blocking has been bothering to me

If you think `pagefair.com` shouldn't be in Peter Lowe's, why not ask him to remove it? There must be a reason why he added it in his list.

Yuki2718:
Peter's is hosts file, not having `$3p` and 0 or 1. Based on publicwww it seems `asset.pagefair.com/measure.min.js` shoudl be blocked.

Yuki2718:
Don't know what quote that is, anyway what are blocked here are not adservers but ordinary and/or useful sites. Adservers are all blocked by EL.

gorhill:
So essentially you are arguing having Peter Lowe's as a default list.

Yuki2718:
I'm pro for having PL in the default lists as I have actually seen cases it catched something missed by other lists. Only questioning some aggressive blocking which we can had had been fixed with `$3p`.

gorhill:
We come back to the previous discussion: using `3p` removes the domain and all subdomains from being candidate for strict blocking. That's the issue.

Yuki2718:
Changes since the last discussion:
- `$doc` was not mentioned at the time
- We unblocked `t.co` and GA in https://github.com/uBlockOrigin/uAssets/issues/18576 (I believe it's right decision, any blogger may use GA)
(And still no single confirmed case that any of them actually turned to click-through tracker)

gorhill:
> We unblocked `t.co` and `analytics.google.com`

We used our judgment that blocking `t.co` was a problem due to widespread use of `t.co` to shorten links. As for `analytics.google.com` we know it's not used for tracking.

> no single confirmed case that any of them actually turned to click-through tracker

That's the problem, if they turn, strict-blocking won't be there to make it visible.

Yuki2718:
> As for `analytics.google.com` we know it's not used for tracking.

But what if I claim `analytics.google.com` may be used as click-through tracker in future? This is the very problem here. We all agreed that click-through tracker should be strict-blocked, but the story is different if it's not click-through tracker but harmless sites. The only reason we keep these rules is assumption without evidence. We *know*, say, taboola uses, and has been using, only `api.taboola.com` for click-through tracker, this is why the rule was there. What's the basis taboola may use parent domain or other subdomain in future? Even if that's the case, what is different from all other tracker which we and all the contributors seek for and add to filters once found?

gorhill:
> But what if I claim analytics.google.com may be used as click-through tracker in future?

Then we block it. People visiting that site are welcome to disable their blocker.

Yuki2718:
> Of course I assume that the claim would come with actual cases supporting it.

If so, we can revert at least some of https://github.com/uBlockOrigin/uAssets/commit/1142bef68ee287d27f6191c433ff8491570b6612 , right? I'm fine to block confirmed tracker, but am against blocking domain for unconfirmed theoretical possibility.

gorhill:
Let's wait for reports, than we can act on a per-cases basis?

Yuki2718:
All of them were reported once, but okay.

gorhill:
Yes, and if again, depending on the report, will just tell people they can just disable uBO if they want to visit the site, or email Peter Lowe if they disagree with the block. We can explain the reason we won't work around the block: we trust Peter Lowe and won't second guess him, but whoever else is free to do so.

Yuki2718:
Well, it's not on I don't trust Peter, but rahter `$3p` difference is significant in these domain. But TBH I don't wanna waste more of your and my time on this.

Issue #7897, opened 2020-09-11
Multiup.*


dhana-exe:
```
https://multiup.org/en/mirror/bfb44d260f4f3908e5cb057c2fec0fb6/Koikatu-All-In-One-Re-Pack.rar.rar
https://multiup.eu/en/mirror/bfb44d260f4f3908e5cb057c2fec0fb6/Koikatu-All-In-One-Re-Pack.rar.rar
```

if you click download... like on googledrive button... it redirect to ads website.... can u guys do that?
block that ads?

ghajini:
It will addressed as per @gorhill decision

This is discussed many times
https://github.com/uBlockOrigin/uAssets/issues/5435

Use this userscript in violentmonkey

`document.querySelectorAll("form>button[link]").forEach(b => {
  b.addEventListener("click", ev => {
    ev.preventDefault();
    location.href = b.getAttribute("link");
  }, true);
});`

Dhanamerdeka:
@ghajini amazing... this work like a charm....
but can i ask?
how do i can copy the link like googledrive with right click?
the link can't be copy... can you help me?

ghajini:
Wait for gorhill

Dhanamerdeka:
ok i will wait. :)

krystian3w:
No possible safe repair right click on `gdrive` (Google also for disk use html5 menu) / `google search` = possible break completly google maps.

Dhanamerdeka:
@krystian3w  ummm i mean... on multiup... if u click "download" they now can automatically go to googledrive without go to ads web... but what i ask... i want to copy those link without go to that link...
so i mean... i need something to make me onlu just do "right click > copy link"

mapx-:
You have to use `inspect` (right click on element then `inspect`) and you'll find the wanted link

krystian3w:
This would require another script which, on the basis of the button code ("html `<button>`"), will paste a link to the page ("html `<a>`").

uBlock-user:
This is a can't fix case as you need a dedicated [scriptlet](https://github.com/uBlockOrigin/uAssets/issues/5435#issuecomment-544169204) for this.

Dhanamerdeka:
@mapx- i know what inspect element... but i need something that direct... so i dont have to open that thing.... :(

Dhanamerdeka:
@krystian3w i not too good when it come to this type of things....
but how to use that?

Dhanamerdeka:
[...]

PR #10173, opened 2021-10-10
Usemet.nl fake buttons

krystian3w:
new name "Usemet.nl"

see #7897


krystian3w:
IMO better hide new name too, we can wait 10 years before back oldie domain after ICANN disable quarantine.

PR #10221, opened 2021-10-16
update ph(nsfw)

ThinksHeIsCool:
pornhub is nsfw
you may need a US IP if you try to encounter the ad I encountered


ThinksHeIsCool:
Also, before anyone tells me to report this to EasyList, I already have, but the filter seems to have been removed at some point.

S0dV:
Can someone prevent an ad on PH for `weedmaps` from appearing on the screen(s) of whoever it may or may not?
**You may need a US IP address in order to encounter it**
See https://github.com/easylist/easylist/issues/10840 if you can and want to.

PR #10242, opened 2021-10-17
Fix https://github.com/easylist/easylist/pull/9370

DeepChirp:
The domain used to track the user


uBlock-user:
A beacon is fired from that domain when navigating away from home page to a blog-post page.

Issue #9946, opened 2021-09-09
[Anti-adblock] pcbeta.com


shenzhiming88:
### Prerequisites

- [X] This is not a [support issue or a question](https://old.reddit.com/r/uBlockOrigin/)
- [X] I performed a [cursory search of the issue tracker](https://github.com/uBlockOrigin/uAssets/issues) to avoid opening a duplicate issue
- [X] I am running the latest version of uBlock Origin

### I tried to reproduce the issue when...

- [X] uBO is the only extension
- [X] uBO with default lists/settings
- [X] using a new, unmodified browser profile

### URL(s) where the issue occurs

`https://www.pcbeta.com/`

### Describe the issue

The page prompts the user to close the ad blocking software

### Screenshot(s)

![image](https://user-images.githubusercontent.com/66902050/132706294-d329bb48-47b6-41f3-9a3a-fcd269ff9933.png)

### uBlock Origin version

v1.37.0

### Browser name and version

Microsoft Edge 92.0.902.62

### Settings

_No need_

uBlock-user:
try `pcbeta.com###append_parent`

shenzhiming88:
@uBlock-user This does not seem to work, the ad placeholder is still there and the page uses a notification reminder to turn off the ad filter
![image](https://user-images.githubusercontent.com/66902050/132711322-0b46ae65-a6d5-4e14-9a41-1ce77014001c.png)

uBlock-user:
Updated

> the ad placeholder is still there

`pcbeta.com##body > div#wp div[class]:matches-css(height: 60px)`

helps ?

shenzhiming88:
@uBlock-user Yes, but the ad placeholder is still there
I think this rule can work
```
/\/data\/cache\/(?! (common|forum|home|portal))[a-zA-Z]+\.js/$script,domain=pcbeta.com
```
(Ref: `https://github.com/cjx82630/cjxlist/commit/26f7cd2ca61da81deef0dc09e96f80c53333120d`)

uBlock-user:
@gorhill Seems site's MutationObserver is overiding `pcbeta.com##body > div#wp div[class]:matches-css(height: 60px)` any way to defuse the MO ?

uBlock-user:
>  I think this rule can work

doesn't match any script there, so not working.

shenzhiming88:
@uBlock-user
[...]

PR #10257, opened 2021-10-19
update https://github.com/uBlockOrigin/uAssets/issues/9946

shenzhiming88:
Issue: https://github.com/uBlockOrigin/uAssets/issues/9946#issuecomment-946856745
I can't find a better way to address it.


uBlock-user:
That no longer works either,

`pcbeta.com##+js(aost, MutationObserver, observe)` ?

PR #10452, opened 2021-11-12
Fixes https://github.com/Crystal-RainSlide/AdditionalFiltersCN/issues/1

shenzhiming88:
> `ga.giuem.com` is a domain that hold google analysis proxy service, so users in china mainland can be tracked by google analysis. However, many users are not aware of being tracked by this.

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -332,3 +332,6 @@ reddit.com##+js(no-xhr-if, method:POST url:/^https:\/\/www\.reddit\.com$/)
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/qohky0/sfgatecom_images_blocked_when_using_the_ublock/
 ||cqrvwq.com^
+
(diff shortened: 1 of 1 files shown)


Yuki2718:
The domain is dead and no real example.

uBlock-user:
Used at `https://myip.cx/`

Yuki2718:
> Used at `https://myip.cx/`

I meant OP has to provide actual example, the one in the link is deal. In any case the name is not resolved on my end.

uBlock-user:
```
https://unpkg.com/@giuem/ga-proxy@1.1.2/dist/ga.min.js
https://github.com/giuem/ga-proxy
```

shenzhiming88:
Therefore, should this domain be blocked?

spirillen:
> unpkg.com

Hmm can't help notice that among all the records I have the only type of sites using `unpkg.com` is NSFW site :thought_balloon: :thinking:

PR #10647, opened 2021-11-28
Add domains

iam-py-test:

```
https://updsec.builtfromzero.com/minecraft-apk-download-hack-version/
https://riteupd.mimanduca.co/minecraft-hack-game-free-download/
https://momupd.enuguhomes.com/download-winrar-crack/
https://www.youtube.com/watch?v=8nY7SnvNxH4
```

### Describe the issue

Unblocked scam websites, and a password-stealing trojan

### Screenshot(s)
image
image
image

### Steps to reproduce (for the first three):
- Go to the url
- Click Download
- Wait as it counts down
- Once done, click the download button in the popup
### Steps to reproduce (3)
- Go to the YouTube video
- Click on the Bit.ly link in the description and download the file from Mediafire
- Unarchive the file using the password _11111_
- Upload the files inside to VirusTotal and note that they are password-stealing trojans

### Versions

- Browser/version: Firefox 94.0.2
- uBlock Origin version: uBlock Origin 1.39.1b1

### Settings

- Advanced user mode: Enabled
- Dynamic filtering mode: Hard
- uBlock filters: All enabled
- Ads: All enabled
- Filterlists:
image
image

### Notes

More info at https://github.com/iam-py-test/investigations/blob/main/2021/11/28/1.md and https://github.com/iam-py-test/investigations/blob/main/2021/11/28/2.md. 
Judging by the fact that these two have the same UI & use the same domains in their redirects, they probably are related

The third one is the same as the two above, even redirects to the same domains:
image

The trojan is from https://scammer.info/t/password-stealer/84348 and has been reported at https://bazaar.abuse.ch/sample/462a689d171f543c10efa08e963996d382585b67a6b298ec40d64f924adfb47a/. For verification, look at https://bazaar.abuse.ch/sample/7984602d945d527e03c32cab6c7471ebf34ac8c74c400e318245a3ef24e419af/#intel

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1208,3 +1208,8 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ||bit.ly/3p8kN5V^$document
 ||mediafire.com/file/at2tvnao5x6ivdo/EngineOwning.rar/$all
 ||download2264.mediafire.com/pnk44f5ci9rg/at2tvnao5x6ivdo/EngineOwning.rar^$all
+
+! same as above - hxxpx[:]//momupd[.]enuguhomes[.]com/download-winrar-crack/
+||momupd.enuguhomes.com^$all
+||fifxuq.granddidrepeat.top^$all
+||granddidrepeat.top^$document
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1201,3 +1201,10 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ||oaesso.info^$document
 ||abcwkv.naturaloffsteam.top^$document
 ||bestlifeoffer20.com^$document
+
+! https://scammer.info/t/password-stealer/84348
+! https://bazaar.abuse.ch/sample/462a689d171f543c10efa08e963996d382585b67a6b298ec40d64f924adfb47a/
+||youtube.com/watch?v=8nY7SnvNxH4^$all
+||bit.ly/3p8kN5V^$document
+||mediafire.com/file/at2tvnao5x6ivdo/EngineOwning.rar/$all
+||download2264.mediafire.com/pnk44f5ci9rg/at2tvnao5x6ivdo/EngineOwning.rar^$all
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1185,3 +1185,19 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ! https://github.com/uBlockOrigin/uAssets/pull/10620
 ||mcdupe.com^$all
 ||shadyclient.net^$all
+
+! https://github.com/iam-py-test/investigations/blob/main/2021/11/28/1.md
+||updsec.builtfromzero.com^$all
+||money-mod.com^$document
+||gospelchor.info^$document
+||smwgwa.brownbrothersilver.top^$document,popup
+||new.bestlifeoffer20.com^$document
+||reykijnoac.com^$document
+||totalnicefeed.com^$all
+||omnatuor.com^$all
+
+! https://github.com/iam-py-test/investigations/blob/main/2021/11/28/2.md
+||riteupd.mimanduca.co^$all
+||oaesso.info^$document
+||abcwkv.naturaloffsteam.top^$document
+||bestlifeoffer20.com^$document


Yuki2718:
On my end the layout of `https://updsec.builtfromzero.com/minecraft-apk-download-hack-version/` is too different, not something like mediafire:

bz

Yuki2718:
`https://www.youtube.com/watch?v=8nY7SnvNxH4` is suspended now.

Yuki2718:
We certainly do not want to add every single URLs of these, the opposite has been discussed internally - bloating lists with entries few user actually access is a major concern and uBO is never meant to protect all the bads on the Internet, we even discussed whether we should add traffic-limits for sites we'll address. you're free to keep lists for them on you side.

iam-py-test:
> We certainly do not want to add every single URLs of these, the opposite has been discussed internally - bloating lists with entries few user actually access is a major concern and uBO is never meant to protect from all the bads on the Internet, we even discussed whether we should add traffic-limits for sites we'll address. you're free to keep lists for them on you side.

Idea: 
I know you don't accept new filterlists, but I was thinking, how about adding another anti-malware blocklist, that way the uBo team can spend their time handling ads and trackers (and doing non-uBo stuff).

Yuki2718:
> Idea:
> I know you don't accept new filterlists, but I was thinking, how about adding another anti-malware blocklist, that way the uBo team can spend their time handling ads and trackers (and doing non-uBo stuff).

Better to comment on https://github.com/uBlockOrigin/uBlock-issues/issues/1116 or https://github.com/uBlockOrigin/uBlock-issues/issues/984. The problem is there's no well-maintained list with good quality. urlhaus was chosen because only this was acceptable and the maintainer cooperated well. In any case blocking bad sites by uBO lists updated at best every 24h is more of a futile effort in the era most of scam sites doesn't live longer than 3 days.

iam-py-test:
> no well-maintained list with good quality

What do you think of @DandelionSprout's list? It does do most of it's blocking through blocklisting entire TLDs, but is maintained. 
> uBO lists updated at best every 24h

Why did uBo never implement support for hours? Doesn't AdBlock Plus support that?

gwarser:
@iam-py-test expiration time in hours should work fine, It's the server problem.

iam-py-test:
> @iam-py-test expiration time in hours should work fine, It's the server problem.

Ah. For some reason I thought that that was not supported. Thank you!

Yuki2718:
@iam-py-test You have to convince @gorhill , not me.

gorhill:
> What do you think of @DandelionSprout's list?

I need a link to that list to be able to look at it and form an opinion about it.

Yuki2718:
IMO `||top^$doc,domain=~corriente.top|~gdtot.top` alone is enough to reject.

iam-py-test:
> IMO `||top^$doc,domain=~corriente.top|~gdtot.top` alone is enough to reject.

Examples of legit .top domains? I have only ever seen those two. Personally, I’d be more worried about blocking say, .tk (where I see many legit domains)

DandelionSprout:
I could make a uBO-inclusion list version without e.g. `||tk^(...)`, `||ml^(...)`, `||cf^(...)` and the "IQ test sites" section, if there's a demand for it.

But the `||top^(...)` entry would remain as a matter of life and death. It has saved me from malware redirections on at least 15 occasions.

Yuki2718:
```
http://xn--mnqp22j55ekji.top​
https://sobanomi-ikkanjin.top/
http://www.bubuka.top/
https://golfgear.top/
http://euzs.top/
https://www.tsuri.top/oita
https://ossme.top/
https://corriente.top/
https://reminder.top/
```

It doesn't take a minute finding a dozen of legit .top sites with `site:top`, tho some scam are blended into and YMMV as my Google search returns Japanese sites. Also there are many .top entries in popular lists e.g. `.top##` search in uBlock filters returns 64 hits (maybe about 20 sites). These are very well expected as those global TLDs are for everyone and not exclusively for bad actor.

> I could make a uBO-inclusion list version without e.g. `||tk^(...)`, `||ml^(...)`, `||cf^(...)` and the "IQ test sites" section, if there's a demand for it.

Search engine filters should be removed too, and whether to do so or not it's better to rewrite `www.google.*##.g:has()` with `:upward()`. With `:has()` every .g has to be evalutated and there are a dozen of such filters.

There seems to be quite redundancy with Badware list (e.g. `adblock.gjtech.net`) and dead or parked domain (e.g. `dutchmega.nl`). TBH I'm also partly responsilbe for redundant or obsolete regex filters, couldn't find time to continously open PRs. You can remove `/^https://www\.namejp0[1-9]\.xyz/$doc,popup,domain=xyz` and some other regex need update. I don't understand why `msn.com##.irisbanner` is in this list. `addons.mozilla.org` filters won't work as it's a privileged page.

DandelionSprout:
In that case I can only presume that Japan is happier about `.top` registrations than what Norway is, because the ones I get are less nicely promising:

<details><summary>Screenshot:</summary>

image

</details>

>Search engine filters should be removed too

I'll take note of that for when there's a need to make a uBO-inclusion list version.

>it's better to rewrite `www.google.*##.g:has()` with `:upward()`. With `:has()` every .g has to be evalutated and there are a dozen of such filters.

I'll put that on my to-do list for this weekend at earliest.

>dead or parked domain (e.g. `dutchmega.nl`)

In 2 or 3 of the list's sections, domains remain even after being parked, because they're hardcoded in e.g. user guides or user manuals and can end up being bought in the future.

>`You can remove /^https://www\.namejp0[1-9]\.xyz/$doc,popup,domain=xyz`

I'll take care of that pretty soon now.

>I don't understand why `msn.com##.irisbanner` is in this list.

Around 2019, MSN would occasionally have a top sticky banner that promoted an extension called "MSN New Tab". Those who remember various MSN IE toolbars, know that such an extension is a bad idea.

>`addons.mozilla.org` filters won't work as it's a privileged page.

The entries were designed with AdGuard for Windows in mind, the only known major adblocker who can filter on privileged pages. If I don't remember completely wrong, `addons.mozilla.org` entries technically do work in Chromium uBO.

gorhill:
AFter looking into it, I prefer to decline including it in the _Malware domains_ section.

Yuki2718:
> The entries were designed with AdGuard for Windows in mind, the only known major adblocker who can filter on privileged pages. If I don't remember completely wrong, addons.mozilla.org entries technically do work in Chromium uBO as well.

Yep, written in comment and it's not priviledged on Chromium.

DandelionSprout:
>After looking into it, I prefer to decline including it in the *Malware domains* section.

I feel I could've been given a chance to make a special list version that address all or almost all of the uBO team's concerns; as I've got experience with making special list versions for other adblockers' syntaxes and policies.

But I've had fatigue problems all autumn long and am unsure how quickly and consistently I'd be ready to initiate a "Feedback to me" process that'd likely last a week. 😅

iam-py-test:
On the topic of including lists, what about Legit URL Shortener?

DandelionSprout:
I can first-hand confirm that iam-py-test has been invaluable for *Legitimate URL Shortener* (even if I was hoping for this thread to focus more on Anti-Malware List), and that he is a full-time contributor with merge rights to it.

I could've got a 3rd contributor if needed, but the only other applicant focused too narrowly on PR-Chinese sites.

iam-py-test:
> > it is bigger then the AdGuard one already in uBo
> 
> Bigger is disadvantage too. There is a reason we (= AG maintainer) don't add some entries in LU to AGUTP. User don't always report problems.

True. AdGuard has the advantage of an online issue reporter. 
However, we do testing and try to handle issue reports as quickly as possible

Yuki2718:
Ofc should not be enabled, changing default-enable list is a big event and can not happen that easily.
> Those who have chosen to use this filter list are more computer literate and more likely to identify and report problems.

It has been repeteadly proven that this assumption is completely wrong. There's always people like `https://old.reddit.com/r/uBlockOrigin/comments/r5k3jt/why_it_doesnt_have_option_to_check_all_option_in/`, IDK how many times I saw this kind on Reddit, and reports from such user which often end up with "Cannot reporoduce" have long been annoyance on AG repo. Those computer literate are supposed to be able to find whatever lists they need on the Internet and can add it to Custom list, end of the story. The fact is even AGUTP which takes safer approach than LU was not instantly added as we had to evaluate the amount of breakage it causes. We removed Enhanced Tracking list and eager to remove than add.

PR #10660, opened 2021-11-29
Fix https://github.com/easylist/easylist/issues/8173#issuecomment-981723362 - breakage

shenzhiming88:
This rule results in not getting the little heart gift, please withdraw the PR
This rule is designed to track how long a user is on the air in order to obtain live gifts, not malicious tracking


Yuki2718:
You mean mobile app?

shenzhiming88:
@Yuki2718 The web side is also broken, except that many people should not notice this feature. And, the domain is indeed used for tracking.

PR #10662, opened 2021-11-29
Update badware.txt

shenzhiming88:
These are API domains that are used for malicious purposes. And, they have been intercepted by `Huorong` (An antivirus software in China).

Although this domain cannot be accessed, prompting 404, its subdomains can be accessed normally.

```
http://api.crm.duominuo.com/public/login
http://wnl.duominuo.com
```

> Recently, the Huorong security team found that a number of software under a commercial enterprise carries a backdoor virus "Backdoor/Jspider". The virus uses infected computers as "broilers" to pick up information about companies on platforms such as Alibaba and WeChat, as well as to brush up their rankings on search engines.
> According to the analysis of the Huorong security team, the backdoor virus "Backdoor/Jspider" spreads through the enterprise's software such as "Durian Ticket King", "Look Pretty" and "258 Business Defender", etc. Once the user's computer installs the above software, it will be infected by the virus, and even if the software is uninstalled, the virus remains in the computer to do evil.
> After the user's computer falls into the "broiler", it will receive remote instructions to visit Alibaba (`www.1688.com`), Qingbo big data (`www.gsdata.cn`) and major search engines (Baidu, 360, Sogou and China Search), not only picking up Alibaba's business registration information and transaction content (such as trade total demand information, etc.), but also picking up the various enterprises in the WeChat public number information, and in the search engine for some companies and products to brush ranking.
> The above-mentioned manipulation of "broilers" will take up a lot of CPU resources of the infected computer, resulting in slowing down and heating up of the computer.
> The Huorong security team found that this virus appeared as early as 2014. The creators of this virus are very cautious, when they detect the presence of "360 Security Guard" and "Tencent Computer Manager" in the computer, the virus will not be downloaded and installed.

image

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1187,6 +1187,4 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ||shadyclient.net^$all
 
 ! https://www.huorong.cn/info/151065569478.html
-||api.crm.duominuo.com^$all
-||static.duominuo.com^$all
-||wnl.duominuo.com^$all
+||duominuo.com^$all
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -1185,3 +1185,8 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ! https://github.com/uBlockOrigin/uAssets/pull/10620
 ||mcdupe.com^$all
 ||shadyclient.net^$all
+
+! https://www.huorong.cn/info/151065569478.html
+||api.crm.duominuo.com^$all
+||static.duominuo.com^$all
+||wnl.duominuo.com^$all


Yuki2718:
We certainly do not want to add every single URLs of these, the opposite has been discussed internally - bloating lists with entries few user actually access is a major concern and uBO is never meant to protect all the bads on the Internet, we even discussed whether we should add traffic-limits for sites we'll address. you're free to keep lists for them on you side.

Yuki2718:
Isn't this a C&C server? There's little sense to block C&C server by browser addon.

shenzhiming88:
Some pages may also use this API, so if I can find a site that uses this malicious URL, I will post it here

Yuki2718:
@shenzhiming88 Okay, thx.

shenzhiming88:
@Yuki2718 Uh, sorry, I can't seem to find a way to find which pages use this third-party domain.

Yuki2718:
Yep, publicwww says 0 hit: `https://publicwww.com/websites/%22duominuo.com+depth%3Aall/`. I'll close this PR for now.

Yuki2718:
You're right, `https://publicwww.com/websites/%22duominuo.com%22+depth%3Aall/` returns 100 hits. Still I can't find actual fire to the domain by these links.

PR #10691, opened 2021-12-01
Remove redundant filter

iam-py-test:
`https://github.com/uBlockOrigin/uAssets/blob/master/filters/annoyances.txt`

### Describe the issue

The filter `pelismart.com##*:style(-webkit-touch-callout: default !important; -webkit-user-select: text !important; -moz-user-select: text !important; -ms-user-select: text !important; user-select: text !important;)` appears twice without reason.
So I removed it

### Screenshot(s)

Not needed

### Versions

- Browser/version: Mozilla Firefox 94.0.2
- uBlock Origin version: uBlock Origin 1.39.1rc0

### Settings

Not needed. 

### Notes

I used a Python script to go through this, but verified this manually.


Yuki2718:
It's better to merge the two section.

PR #10704, opened 2021-12-02
Fix Mixpanel 1p issues on https://github.com/easylist/easylist/issues/9920

ryanbr:

`https://mixpanel.com/project/2138137/view/290551/app/dashboards#id=639383`

### Describe the issue

Due to Peter Lowes list, blocking `mixpanel.com` causing first-party issues.

### Screenshot(s)

144368125-596d2790-4214-4ba6-a78b-0c00700df87e
### Versions

- Browser/version: Brave
- uBlock Origin version: 1.38

### Settings

### Notes

Easyprivacy blocks mixpanel already, so just badfiltering is best option here

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -285,7 +285,7 @@ ppss.kr##.ppss-title-banner:remove()
 ||g.jwpsrv.com/g/gcid-*?notrack$frame
 
 ! https://github.com/uBlockOrigin/uAssets/issues/10012
-homedepot.com,tacobell.com##+js(set, bmak.js_post, false)
+tacobell.com##+js(set, bmak.js_post, false)
 
 ! https://github.com/easylist/easylist/pull/9136
 ||cloudflare.com/ajax/libs/fingerprintjs2/$script,redirect=fingerprint2.js,important,domain=gamebox.gesoten.com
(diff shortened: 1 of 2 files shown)


uBlock-user:
> Due to Peter Lowes list, blocking mixpanel.com causing first-party issues.

Regarding that, there has been a policy change - https://github.com/uBlockOrigin/uAssets/pull/10149#issuecomment-951814995

ryanbr:
Disappointing. Looking from a webcompat perspective. Its an unnecessary first party block given its blocked in Easyprivacy via a 3rd-party block. To disable uBO or Brave shields just opens up more trackers to the user for the sake of policy.

mapx-:
@gorhill

gorhill:
> opens up more trackers to the user for the sake of policy.

The pushers of tracking service mind trackers?

ryanbr:
> > opens up more trackers to the user for the sake of policy.
> 
> The pushers of trackers mind trackers?

I'm not a fan of them given the amount we block already.  The actual tracking mixpanel bits are outside of their 1p site, that's where it matters. 

Maybe not this site specifically but suggesting to turn off uBO to get something work seems a sledgehammer approach for fixing 1p pgl issues. I've said my piece, if it's ubo policy so be it.

uBlock-user:
>  Maybe not this site specifically but suggesting to turn off uBO to get something work seems a sledgehammer approach for fixing 1p pgl issues. I've said my piece, if it's ubo policy so be it.

Silent redirection tracking is not protected under `3p` which is gaining traction these days because trackers and ad-servers are blocked as `3p`. Blocking as `3p` != blocking the whole domain that is something one must understand. 

The only people who need to access the site are their customers/employees, so they can always whitelist the site which is only a click away, With that said, hardly a sledgehammer approach.

Issue #8706, opened 2021-03-12
send.cm


ghajini:
### Prerequisites

- [x] This is not a support issue or a question
    - Support issues and questions are handled at [/r/uBlockOrigin](https://old.reddit.com/r/uBlockOrigin/)
    
- [x] I performed a cursory search of the issue tracker to avoid opening a duplicate issue
    - Your issue may already be reported.
- [x] I tried to reproduce the issue when...
    - [x] uBlock Origin is the only extension
    - [x] uBlock Origin with default lists/settings
    - [x] using a new, unmodified browser profile
- [x] I am running the latest version of uBlock Origin

### URL(s) where this occurs

`https://send.cm/xm2n3sets3ot`

### Describe the issue

Popup ads

### Versions

- Browser/version: Firefox a nightly
- uBlock Origin version: 1.33.2

### Settings

 uBO's default settings
-

ghajini:
there's popup

`https://send.cm/fyorljhkqqt2`

Wassss:
`send.cm##^script:has-text(FingerprintJS)` break download link
how can i remove it ?

mapx-:
test link ?

ghajini:
**Test link=**

`https://send.cm/9r1esq0jf16o`

where's broken, i can download the file

**Screenshots=**

https://user-images.githubusercontent.com/20338483/142376088-c4bf1382-c08b-4923-ba60-7e9c675b90fe.png

https://user-images.githubusercontent.com/20338483/142376098-35c8f102-bd54-4b7c-8388-6e7eb50bb377.png

Wassss:
@mapx- it work now
thank you

Demanding0536:
Annoyance: On video file pages, right click is blocked and a keyboard shortcut to open dev tools is blocked (ctrl+shift+i). Opening dev tools/inspect causes the page to close itself, and loading the page with it enabled will cause it to close itself during load.

Test link: `https://send.cm/d/naW4`

Tested with latest Firefox and uBO, default lists and default + all annoyance filters.

PR #10720, opened 2021-12-03
Update https://github.com/uBlockOrigin/uAssets/issues/8706

ghajini:
```
https://send.cm/63b81dcgju9o
https://send.cm/lkq3rx1e9n3r
```
### Describe the issue

propeller ads

### Screenshot(s)

### Versions

- Browser/version: firefox stable
- uBlock Origin version: 1.39

### Settings

- 
- uBO's default settings

### Notes

Reviews for different browsers
@mapx-, @okiehsch
any breakages

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1740,6 +1740,7 @@ elitepvpers.com##div > .page > div > div[style]:has-text(/^Advertise/i) + a
 
 ! https://github.com/uBlockOrigin/uAssets/issues/8706
 send.cm##+js(aopw, decodeURIComponent)
+send.cm##+js(acis, Promise, JSON.parse)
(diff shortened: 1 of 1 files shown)


mapx-:
why not the classic 
`send.cm##+js(acis, JSON.parse, break;case $.)`

ghajini:
does not work always, comes after refresh

Yuki2718:
> does not work always, comes after refresh

I wonder if other cases we have added HTML filter can also be resolved by changing to Promise. AG also uses this.

ghajini:
> > does not work always, comes after refresh
> 
> I wonder if other cases we have added HTML filter can also be resolved by changing to Promise. AG also uses this. 

no, don't replace html filters, they are best

PR #10738, opened 2021-12-05
Move some filtering rules to Privacy

shenzhiming88:
These two filtering rules are not related to resource abuse, but are related to privacy. Therefore they should be moved to the privacy list.
At the same time, I use `||a.msstatic.com/huya/*/p2plib.js$script` instead of `||p2p.huya.com^$xhr` to fix https://github.com/uBlockOrigin/uAssets/pull/10610


LikesHisComputer:
@ryanbr

shenzhiming88:
Through the tests of @inoton and I, the current filtering rules will not cause any problems. So you can safely merge it.

PR #10744, opened 2021-12-05
Update filters.txt

shenzhiming88:
`https://cn.bing.com/search?q=Chrome`
This is the blank left after the removal of the ad
image


Yuki2718:
This was removed from EP for FP: https://github.com/easylist/easylist/commit/06efcf3cb8fdecb70eeedda9c8f32cb1106222b9

PR #10769, opened 2021-12-05
addresses Yavli ads

LikesHisComputer:


LikesHisComputer:
If anyone is trying to look for the URL - `https://www.go4games.ro/reviews/call-of-duty-vanguard-review-19295998/`

PR #10909, opened 2021-12-14
Better Target Klaviyo Analytics files

nshenkman:
Example URL : `https://omniarc-gaming-mouse.myshopify.com/`

### Describe the issue
The current url targets an API Klaviyo uses whereas this target prevents the tracking JS from ever being loaded. The API can be used for other purposes outside of tracking so this solution better targets Klaviyo Tracking.

### Screenshot(s)
Screen Shot 2021-12-10 at 3 08 09 PM

### Versions

- Browser/version: Chrome Stable
- uBlock Origin version: 1.32.5b10

### Notes

### DIFF
--- a/thirdparties/pgl.yoyo.org/as/serverlist
+++ b/thirdparties/pgl.yoyo.org/as/serverlist
@@ -51,7 +51,6 @@
 127.0.0.1 a-ads.com
 127.0.0.1 a.aproductmsg.com
 127.0.0.1 a.consumer.net
-127.0.0.1 a.klaviyo.com
 127.0.0.1 a.mktw.net
 127.0.0.1 a.muloqot.uz
 127.0.0.1 a.pub.network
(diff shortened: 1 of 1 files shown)


uBlock-user:
@pgl

pgl:
@nshenkman Can you give some examples of non-tracking uses this API is used for, please?

nshenkman:
Hi @pgl, that API is used for Forms subscriptions. So people who specifically fill out and submit a Form are currently unable to because the block on a.klaviyo.com exists. I've added a screenshot with an example
Screen Shot 2021-12-14 at 12 27 26 PM
.

pgl:
OK, thank you. I've updated the entry in my list. For future reference you can contact me directly at pgl@yoyo.org, it may be quicker to get a resolution.

nshenkman:
> OK, thank you. I've updated the entry in my list. For future reference you can contact me directly at [pgl@yoyo.org](mailto:pgl@yoyo.org), it may be quicker to get a resolution.

Awesome, Thank you! Am I good to Close out this PR then?

PR #10942, opened 2021-12-16
Update annoyances.txt

shenzhiming88:
`http://www.4399.com/`
This is the annoying background element and cannot be blocked by normal element filtering rules.
image


shenzhiming88:
I think it is better to add to Fanboy Annoyance.

PR #9649, opened 2021-07-27
[Anti-copy] 360doc.com

shenzhiming88:
Address:`http://www.360doc.com/content/21/0720/00/65661269_987374075.shtml`

PR #11042, opened 2021-12-23
Update https://github.com/uBlockOrigin/uAssets/pull/9649

shenzhiming88:


mapx-:
explain the issue

shenzhiming88:
@mapx- Provide longer term guarantee.
It is also possible to remove the above three lines of filters.
```txt
360doc.com##+js(aopr, document.oncontextmenu)
360doc.com##+js(aopw, document.body.oncopy)
360doc.com##+js(set, document.body.onmouseup, null)
```

mapx-:
still working for anti copy

shenzhiming88:
@mapx- Yes, but I think it's better to intercept iterative scripts so that the pages load faster. If you inject scripts, then access will be slowed down.

mapx-:
maybe 
`360doc.com##+js(noeval-if, copy)`

shenzhiming88:
> maybe `360doc.com##+js(noeval-if, copy)`

Why do we need to inject a script when we can intercept this script and be able to replicate it?

krystian3w:
May break page in the future, maybe Fanboy Annoyances block script if allow unlock copy by this simple network filter.

PR #11149, opened 2021-12-31
Update resource-abuse.txt

shenzhiming88:
Test link: `https://www.bilibili.com/kuawan/2021zmdy/pc/`
The web page uses P2P technology without permission and takes up a lot of upload broadband. After testing, this filter does not cause live streaming lag.
image


uBlock-user:
We have `##+js(nowebrtc)`, use that scriptlet instead.

shenzhiming88:
@uBlock-user Indeed, thank you for your reminder!

PR #11160, opened 2022-01-01
Update privacy.txt

shenzhiming88:
From AdGuard, this Cookie sets a unique user id for tracking how the user uses the site.
https://github.com/AdguardTeam/AdguardFilters/blob/32ed0d5ebce26a82e98de76186c64c07fff89a2c/SpywareFilter/sections/cookies.txt#L407
However, even though I added this line of filter, the cookie still exists and I don't know if it works.
image
image

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -358,3 +358,6 @@ reddit.com##+js(no-xhr-if, method:POST url:/^https:\/\/www\.reddit\.com$/)
 /cfga/jquery.js?$image
 /npm/sks@0.*/lazyload.js$script,3p
 ||skk.moe/*.js?$image
+
+! https://github.com/uBlockOrigin/uAssets/pull/11160
+bing.com##+js(cookie-remover, MUIDB)


shenzhiming88:
@gorhill How about adding a `$cookie` parameter, like AdGuard does? If there are many general rules for removing cookies (like this one), or a list of cookies that specifically remove privacy-related cookies (e.g. https://github.com/AdguardTeam/AdguardFilters/blob/master/SpywareFilter/sections/cookies.txt, which is part of the AdGuard privacy list), I don't think it's a wise choice to insert a script - it's not convenient to see the filter logs. When cookies are intercepted with this parameter, they can be marked red.
Thanks!

gwarser:
Does not work because it's http-only cookie

image

gwarser:
AdGuard doc about `$cookie` https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters#cookie-modifier

shenzhiming88:
@gwarser Okay, I'll try it now and see if AdGuard can intercept it.

gorhill:
> how the user uses the site

A site can know how it is being used by merely analyzing the fetching of resources from that site. Not that I am against supporting a new `cookie` modifier, but I really need to be convinced this really accomplishes something of value rather than just be a feature checkbox.

shenzhiming88:
> > how the user uses the site
> 
> A site can know how it is being used by merely analyzing the fetching of resources from that site -- no cookies needed. Not that I am against supporting a new `cookie` modifier, but I really need to be convinced this really accomplishes something of value rather than just be a feature checkbox.

I'm sorry I don't know much about this. But I think you can check the AdGuard privacy list for cookies - https://github.com/AdguardTeam/AdguardFilters/blob/master/SpywareFilter/sections/cookies.txt, the comments in the list seems to indicate that some cookies are related to tracking

krystian3w:
`$cookie` also block few Russian 🇷🇺  ads, may help reduce use `.user.js` for `dimisa ru-adlist`.

krystian3w (changes requested):

Issue #10602, opened 2021-11-25
douyu.com


shenzhiming88:
### Prerequisites

- [X] This is not a [support issue or a question](https://old.reddit.com/r/uBlockOrigin/)
- [X] I performed a [cursory search of the issue tracker](https://github.com/uBlockOrigin/uAssets/issues) to avoid opening a duplicate issue
- [X] I am running the latest version of uBlock Origin

### I tried to reproduce the issue when...

- [X] uBO is the only extension
- [X] uBO with default lists/settings
- [X] using a new, unmodified browser profile

### URL(s) where the issue occurs

`https://www.douyu.com/topic/jinchan?rid=8747682`

### Describe the issue

This web page uses P2P technology without the user's explicit consent and abuses the user's uplink broadband. I don't know if this falls under the category of `uBlock filters - Resource abuse`.

### Screenshot(s)

![image](https://user-images.githubusercontent.com/66902050/143473126-ed8532a4-86e3-41dc-94d9-1b78be2e72f1.png)

### uBlock Origin version

v1.38.4

### Browser name and version

Microsoft Edge 95.0.1020.44 (64-bit)

### Settings

_No need_

uBlock-user:
> abuses the user's uplink broadband

How exactly ? Provide more specific information.

shenzhiming88:
Simply put, is this the list to deal with for the use of P2P without asking the user?

okiehsch:
> , is this the list to deal with for the use of P2P without asking the user?

Yes.

shenzhiming88:
This page uses P2P without informing the user and takes up all of the user's upload bandwidth.
And this rule is helpful:
```
||sdkapi.douyucdn.cn/p2p?
```
Link: `https://www.douyu.com/19223`
Before:
![image](https://user-images.githubusercontent.com/66902050/143548320-5e1d594b-27ef-4f18-984d-8b18ab4d0502.png)
After:
![image](https://user-images.githubusercontent.com/66902050/143548716-f9eef414-cb6e-4d2f-b9cc-68af4cf9167b.png)

okiehsch:
Hm, I can't reproduce and there is no `sdkapi.douyucdn.cn/p2p?` request on my end.

shenzhiming88:
> Hm, I can't reproduce and there is no `sdkapi.douyucdn.cn/p2p?` request on my end.

I think you'd better use a VPN to China and I can record a video to show you if you want.

[...]

PR #11163, opened 2022-01-01
Update https://github.com/uBlockOrigin/uAssets/issues/10602

shenzhiming88:
Directly block p2p script to prevent lagging problems.


shenzhiming88:
@mapx- If it doesn't work, you can remove these rules.

PR #11192, opened 2022-01-04
Update specific_report_from_ubo.yml

JohnyP36:
### Describe the issue

L28 says `I did not answer truthfully to **all** the above checkpoints`, but I think the **not** is incorrect - and thus should be removed - as I guess the aim of this checkbox is to ensure people speak the truth in the above checkpoints.

### Versions

- Browser/version: MS Edge, v96
- uBlock Origin version: 1.39.2

#### Settings

*irrelevant*

### Notes

If I'm incorrect: sorry for your time.


uBlock-user:
It's a test to see who bothers to read and understand before hitting the check-mark as usual.

JohnyP36:
> It's a test to see who bothers to read and understand before hitting the check-mark as usual.

Aha, that's very smart. 👍🏼

PR #11958, opened 2022-02-28
update hackintoshshop. com

JobcenterTycoon:

`https://www.hackintoshshop.com/`

### Describe the issue

anti adblock alert but it can be closed.
### Screenshot(s)
Unbenannt

### Versions

- Browser/version: firefox developer
- uBlock Origin version: 1.41.8

### Settings

uBlock Origin default + uBlock Origin annoyances


mapx-:
tested in FF / chrome, after clicking ok will land in the page but after that whatever link you click you'll get am infinite-loading page. So you have to move the filters in the main list

Issue #12210, opened 2022-03-13
shortx.net: detection


iam23:
### Prerequisites

- [X] I read and understood the [policy as to what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets)
- [X] I verified that the issue has not already been reported (use [this button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out)
- [X] I forced an update of my filter lists (**how to do this:** [click _"Purge all caches"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches), then [click _"Update now"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now))
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/585534/91296476-ab9d8c00-e76a-11ea-9fb9-87d6861bdf73.png), or I verified that the issue is not caused by removing any of the default lists
- [X] I did not enable external filter lists, or I verified that the issue still occurs without enabling external filter lists
- [X] I do not have custom filters/rules, or I verified that the issue still occurs without custom filters/rules
- [X] I am not using [another content blocker along uBO](https://twitter.com/gorhill/status/1033706103782170625) (also mind browser built-in blockers)
- [ ] I did not answer truthfully to **all** the above checkpoints

### URL address of the web page

`https://shortx.net/yftu0`

### Category

detection

### Describe the issue

detect ublock

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

https://ibb.co/w7VWXnK
</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.41.8
Firefox: 97
filterset (summary): 
  network: 103682
  cosmetic: 48059
  scriptlet: 38819
  html: 630
listset (total-discarded, last updated): 
  added: 
    https://raw.github.com/reek/anti-adblock-killer/master/anti-adblock-killer-filters.txt: 0-0, 16h.30m
    https://raw.githubusercontent.com/bogachenko/fuckfuckadblock/master/fuckfuckadblock.txt: 26851-47, 16h.29m
    https://raw.githubusercontent.com/robonxt/CleanYourTwitter/master/CleanYourTwitter.txt: 24-0, 4d.13h.55m
    adguard-spyware: 26501-903, 3d.17h.4m
[...]

PR #12211, opened 2022-03-13
addresses #12210

PerfectSalad:


PerfectSalad:
https://github.com/uBlockOrigin/uAssets/issues/12210#issuecomment-1066026082

PR #12223, opened 2022-03-13
not needed in uAssets

PerfectSalad:
`faucetearn.com` seems to be for sale, so I removed filtering-directions for it


PerfectSalad:
removing filter for `g1novelas.org`

Issue #11503, opened 2022-01-29
bigbtc.win: detection


UblockAntiAdblock:
### Prerequisites

- [X] I read and understood the [policy as to what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets)
- [X] I verified that the issue has not already been reported (use [this button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out)
- [X] I forced an update of my filter lists (**how to do this:** [click _"Purge all caches"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches), then [click _"Update now"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now))
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/585534/91296476-ab9d8c00-e76a-11ea-9fb9-87d6861bdf73.png), or I verified that the issue is not caused by removing any of the default lists
- [X] I did not enable external filter lists, or I verified that the issue still occurs without enabling external filter lists
- [X] I do not have custom filters/rules, or I verified that the issue still occurs without custom filters/rules
- [X] I am not using [another content blocker along uBO](https://twitter.com/gorhill/status/1033706103782170625) (also mind browser built-in blockers)
- [ ] I did not answer truthfully to **all** the above checkpoints

### URL address of the web page

`https://bigbtc.win/`

### Category

detection

### Describe the issue

Anti-adblock

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.40.8
Firefox Mobile: 98
filterset (summary): 
  network: 99623
  cosmetic: 65252
  scriptlet: 27613
  html: 620
listset (total-discarded, last updated): 
  added: 
    https://www.i-dont-care-about-cookies.eu/abp/: 24288-13337, 3h.16m
    adguard-generic: 61399-5097, 3h.7m
  default: 
    ublock-filters: 30977-325, 3h.17m
    ublock-badware: 3835-1, 3h.17m
    ublock-privacy: 193-1, 3h.17m
    ublock-abuse: 77-0, 3h.17m
    ublock-unbreak: 1703-0, 3h.17m
    adguard-mobile: 7149-173, 3h.17m
    easylist: 62449-24476, 3h.17m
[...]

PR #12224, opened 2022-03-13
update #11503

PerfectSalad:


PerfectSalad:
`https://bigbtc.win/` => `mayalaze.xyz` 3p connections

@ryanbr

PR #12225, opened 2022-03-13
addresses https://github.com/AdguardTeam/AdguardFilters/issues/112769

PerfectSalad:


PerfectSalad:
@mapx- Should `gamingfactor.xyz##[href="https://zlink.tk/"]` be added as well?

mapx-:
if ad why not

PerfectSalad:
> if ad why not

@mapx- any idea for a different filter?

PR #12228, opened 2022-03-13
NSFW seed69. com popunder

PerfectSalad:


PerfectSalad:
> Better to open in EL as no extended syntax is required.

@Yuki2718 Do you think it would be a good idea to have `,redirect-rule=noopjs` to the filter I added?

PR #12231, opened 2022-03-14
fix(urlhaus-filter): replace download link

curbengh:
- https://gitlab.com/curben/urlhaus-filter/-/commit/8c94ddba40d794b194f614091a84f5fc1d00a4b0
- filter update is no longer pushed to repo


okiehsch:
Thx.

PR #12281, opened 2022-03-16
filter update

JobcenterTycoon:

`https://www.rjno1.com/`

### Describe the issue

self breakage

### Versions

- Browser/version: firefox developer
- uBlock Origin version: 1.41.8

### Settings

uBlock Origin default + uBlock Origin annoyances

### Notes

`https://github.com/422658476/banana-detect-adblock`


Yuki2718:
Maybe these should be moved to Quick. Interestingly they only counter uAssets, AG + Chinese has been fine for a week after last update.

JobcenterTycoon:
Because the admin invest all his time to counter my filters ...

See `https://github.com/422658476/banana-detect-adblock` the admin write exacly how to counter my filters because i blocked inline scripts and most inline CSS tags

PR #12351, opened 2022-03-21
Add Twitch.tv analytics

kasper93:

`https://www.twitch.tv/`

### Describe the issue

This blocks Twitch's usage tracking, analytics and bunch of other information sent to video delivery servers. None of this is needed for Twitch to work and only compromise user privacy.

You can see what data is sent by tracking post requests to `ttvnw.net`

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -377,3 +377,6 @@ endbasic.dev,jmmv.dev##+js(no-xhr-if, method:POST)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/11895
 ||serasaexperian.com.br/dist/scripts/fingerprint2.js^$redirect=fingerprint2.js,script,important
+
+! Twitch.tv analytics
+twitch.tv##+js(no-fetch-if, /video-edge-[^/]+ttvnw.net/ method:POST)


kasper93:
One important thing to mention as it cuts the user tracking, all rewards for watch time like channel points or other similar things are not tracked to your account. Same as if you watch twitch stream externally, with streamlink, youtube-dl or just not being logged in.

If you think virtual points are enough to sell your privacy, probably need to not use this rule.

kasper93:
> Hi, we want to add the rule, however, a concern is whether doing so will cause too much trouble for average user - uBO default setting is not solely for privacy-minded people but also for average Joe and Jane. 

I see that Privacy list is enabled by default, so indeed it will affect everyone.

> Can you clarify and demonstrate how the rule will actually affect that?

I wanted to show example of data that is send, but there huge json send with all sort of info and tracking tokens, it is hard to deanonymize without removing almost everything. Please just look at xhr post request to see what is sent.

As for the effect, after testing and to best of my knowledge the main side effect is that Twitch no longer have tracking information what do you watch and how long. Which is one of many data send, but this like have direct impact on user ability to collect rewards like channel points https://help.twitch.tv/s/article/channel-points-guide or other similar rewards which require tracking watch time like drops that sometimes are integrated into streams.

I sent this rule for PR, because as I see there are already rules for 1st-party tracking, like for reddit.com https://github.com/uBlockOrigin/uAssets/blob/c3121376f53f2ef197c339f1018a22dc27653ce5/filters/privacy.txt#L296 and other similar cases.

But I agree, it may be unwanted effect for some Twitch users, so maybe it is more fit for more strict privacy list which is not enabled by default. I still think Twitch tracks too much which is not essential for page to work correctly...

stephenhawk8054:
There was an issue about `Channel Points Acquisition` breakage with twitch in the past: [https://github.com/uBlockOrigin/uAssets/issues/7001](https://github.com/uBlockOrigin/uAssets/issues/7001). Is it the rewards system you mentioned?

kasper93:
> Example of client sided data sent every second via POST --

Yes this is one of the payloads, but there are more, depending on your activity.

PR #12752, opened 2022-04-14
address https://www.reddit.com/r/uBlockOrigin/comments/tzla0l/ad_show…

ghajini:
…ing_up_on_series9_dot_me/

```
https://series9.me
https://series9.me/film/the-penguins-of-madagascar-season-1-bhq/watching.html?ep=47
```

### Describe the issue

popups

### Screenshot(s)

### Versions

- Browser/version: opera

### Settings

-  uBO's default settings

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -11943,7 +11943,9 @@ culturemap.com##+js(aopr, CatapultTools)
 pipocas.tv##+js(nobab)
 
 ! https://adblockplus.org/forum/viewtopic.php?f=10&t=65201&p=187268#p187268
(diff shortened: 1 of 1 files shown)


okiehsch:
OK, finally a case where I can reproduce the new popup script.
Does
`series9.*##+js(acis, parseInt, break;case $.)`
work for you?

I think aborting the inline-script is the preferable fix if it works.

okiehsch:
`series9.*##+js(acis, Math, break;case $.)`
can be removed either way, it is obsolete.

ghajini:
> OK, finally a case where I can reproduce the new popup script.
> Does
> `series9.*##+js(acis, parseInt, break;case $.)`
> work for you?
> 
> I think aborting the inline-script is the preferable fix if it works.

yes it works

PR #12838, opened 2022-04-18
A: ylilauta.org

peace2000:

`ylilauta.org`

### Describe the issue

Occasionally when loading the page, there will be a big nag screen about ads that could not be loaded.

### Screenshot(s)

kuva

### Versions

- Firefox 99.0.1
- uBlock Origin version: 1.42.5b1

### Settings

- Defaults

### Notes

The whole iframe cannot be hid without causing the page to notice it but hiding that big screen of text is fine.

### DIFF
--- a/filters/quick-fixes.txt
+++ b/filters/quick-fixes.txt
@@ -149,6 +149,7 @@ search.seznam.cz##[class]:not([style]):matches-css(background-image: /url.*svg/)
 ||lauta.media^$media,redirect=noopmp4-1s,domain=ylilauta.org
 ||plausible.lauta.media/api/event^$xhr
 ||static.ylilauta.org/api/event^$xhr,1p,redirect=nooptext
+ylilauta.org##body.error
 
 ! https://github.com/easylist/easylist/commit/aa0d63b1b7e66b046f29236ee7a0cc9c90cb7de7#commitcomment-71441976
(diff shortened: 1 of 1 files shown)


JobcenterTycoon:
But the anti adblock is still here so the site notice. What about:

removing `||lauta.media^$media,redirect=noopmp4-1s,domain=ylilauta.org` because they check the vid and add

`ads.ylilauta.org##video`
`ads.ylilauta.org###adf-click:style(pointer-events: none !important;)`

To hide the ad. It works without anti adblock warning for me.

peace2000:
@JobcenterTycoon if that orange warning is removed, then it will display alert by the browser and I don't know how to handle that.

I think those suggestions are fine. I occasionally still see warnings with them but not so often.

JobcenterTycoon:
Ok added. Im on phone so i can't verify at the moment it worked on my pc a hour ago.

For the alert (after i removed the orange warning) i tried the uBO alert buster but it doesn't work i don't know why

peace2000:
@JobcenterTycoon What browser are you using? On Firefox, it can't work currently: https://github.com/uBlockOrigin/uBlock-issues/issues/2079

On Chrome it causes issues because the site has checks and says that browser functions have been tampered with.

PR #12960, opened 2022-04-26
Remove IF rules

bernd87:

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: [here]

### Settings

### Notes

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -624,12 +624,14 @@ amazon.*##.a-carousel-card > div > div[cel_widget_id^="adplacements:"]:upward(2)
 ! https://github.com/uBlockOrigin/uAssets/issues/12268
 amazon.*##.AdHolder
 
+! This needs refine. It currently blocks custom userscripts which is more annoying than what it should solve
+! The ads are still gone without that, so no idea what this rule is about, other than ruining greasemonkey scripts
 ! https://forums.lanik.us/viewtopic.php?p=131942#p131942
-imagefap.com##+js(aopr, Buu.serve)
-imagefap.com##+js(aeld, click, [native code])
-imagefap.com##+js(acis, $, open)
-imagefap.com##iframe[src="about:blank"]
-*$popunder,domain=imagefap.com,3p
+! imagefap.com##+js(aopr, Buu.serve)
+! imagefap.com##+js(aeld, click, [native code])
+! imagefap.com##+js(acis, $, open)
+! imagefap.com##iframe[src="about:blank"]
+! *$popunder,domain=imagefap.com,3p
 
 ! https://github.com/uBlockOrigin/uAssets/issues/1347
 motherless.com##+js(set, _ml_ads_ns, null)


stephenhawk8054:
Can you describe the issue in more details, and how to reproduce it?

The filter 

```
imagefap.com##+js(acis, $, open)
```

is still needed for desktop version `https://www.imagefap.com` to prevent popup.

For mobile version of `https://beta.imagefap.com` (which can only open in mobile environment), I can replace other filters with

```
!#if env_mobile
beta.imagefap.com##+js(aost, open, setPopUnder)
!#endif
```

to prevent the popup. But I'm not sure if it still causes issue for you or not, unless you give us more details about how to reproduce the issue.

bernd87:
hmm, which popups are we talking about? I am on the desktop version & I just tried with the this whole list disabled and i had no popups or ads at all, which i think is the case because they get blocked from some other list?!

`imagefap.com##+js(acis, $, open)` basically blocks any custom scripts it seems

okiehsch:
@stephenhawk8054 I can't reproduce but I can see the script in the source that you want to disable with the scriptlet.
Can you change it to 
`imagefap.com##+js(acis, $, popCookie)`
that should fix the greasemonkey issue I would assume.
@bernd87 
If you add 
````
imagefap.com#@#+js(acis, $, open)
imagefap.com##+js(acis, $, popCookie)
````
to your filter list are your custom scripts working or not?

bernd87:
@okiehsch 
with
`imagefap.com#@#+js(acis, $, open)
imagefap.com##+js(acis, $, popCookie)`
the scripts seem to work fine

bernd87:
I can post and explain the script if needed

stephenhawk8054:
I tested the site `https://www.imagefap.com` with `imagefap.com#@#+js(acis, $, open)` exception and opened in new Private Window. After that, clicking on the links at the `Our niches` table on the left triggered a popup for me.

But yeah, I think `$` and `open` can cause issues here with user scripts, `imagefap.com##+js(acis, $, popCookie)` would be better

okiehsch:
> I can post and explain the script if needed

Thanks, but no need really, the `acis, $, open` means that basically any custom script with a string "open" in it won't work, hence 
my suggestion to change it to `acis, $, popCookie`.
With that change only custom scripts that include the string `popCookie` won't work most of the time and that should be a rather rare occurenc.

okiehsch:
@stephenhawk8054 does the filter
`imagefap.com##+js(aopr, Buu.serve)`
serve any purpose on your end anymore?

stephenhawk8054:
@okiehsch Looks like nope, besides `imagefap.com##+js(acis, $, popCookie)`, I think others are obsolete now. Except maybe `*$popunder,domain=imagefap.com,3p` to automatically close the popup in mobile `beta` version, but it can be replaced by `beta.imagefap.com##+js(aost, open, setPopUnder)` on my side.

okiehsch:
Ok, let's remove them all and add your filter.

PR #13312, opened 2022-05-19
feat: adshield

seia-soto:
### URL(s) where the issue occurs

```
ppss.kr,ygosu.com,tgd.kr,inven.co.kr,loawa.com,feedclick.net
```

and all adshield applied sites.

### Describe the issue

ad, anti-adblocking

### Versions

uBlock Origin: 1.42.4
Waterfox (Firefox): 91

### Notes

https://github.com/uBlockOrigin/uAssets/issues/12163

PR #13313, opened 2022-05-19
fix: exclude loawa.com (breakage)

seia-soto:
From #13312 , `loawa.com` is causing breakage on sub-pages as I checked again and should be excluded for now.


JobcenterTycoon:
Sorry but its to problematic. I did a fast check and ygosu is broken too. `querySelectorAll` get extremly widely used so i removed completely for the first time (want to go to bed i have no time to do a deep check for all sites)

seia-soto:
Okay. I'll try to avoid it in future.

PR #13409, opened 2022-05-26
update https://github.com/uBlockOrigin/uAssets/issues/8706

ghajini:

NSFW `https://send.cm/haaoqngjjdtm`

### Describe the issue

update filter
### Screenshot(s)

### Versions

- Browser/version: firefox a

### Settings

-  uBO's default settings

### Notes
iam keeping `send.cm##+js(aeld, , break;case $.)` if site tries to load propeller requests from not inline

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -1681,7 +1681,7 @@ elitepvpers.com##div[style$="padding:0;"]:has-text(/Advertisement/i)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/8706
 send.cm##+js(aeld, , break;case $.)
-send.cm##+js(acis, document.createElement, 'script')
+send.cm##+js(acis, parseInt, break;case $.)
 send.cm##+js(acis, JSON, break;case)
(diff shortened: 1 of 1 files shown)


ghajini:
@mapx- 
Is this valid?
`send.cm##+js(acis, /parseInt|navigator/, break;case $.)`

mapx-:
no, regex accepted on the second parameter only

ghajini:
so how the filter will look, if i have to target `parseInt, navigator, break;case $.`

mapx-:
if matches

`acis, parseInt, /navigator|break;case/`

mapx-:
or 2 acis filters

ghajini:
or
`acis, parseInt, /navigator|parseFloat|break;case $./`

2 acis 👍

mapx-:
be careful with regex .. if you want to use a point `.`  as part of the matching stuff you have to use it `\.` otherwise has another meaning in regex

ghajini:
ok i will learn all this,ty👍

PR #13450, opened 2022-05-29
Addresses https://github.com/AdguardTeam/AdguardFilters/issues/120118

Branzeturi:


Branzeturi:
@mapx- https://github.com/AdguardTeam/AdguardFilters/issues/120118 -- should it be addressed in the same place?

Branzeturi:
> @mapx- [AdguardTeam/AdguardFilters#120118](https://github.com/AdguardTeam/AdguardFilters/issues/120118) -- should it be addressed in the same place?

@mapx- Could `/css/adblook.css$1p` be added?

PR #13502, opened 2022-06-01
Fix ad popup on discussingfilm.net

ryanbr:

### Describe the issue

Opening `discussingfilm.net` (and clearing cookies) and restore scroll.

### Screenshot(s)
discus

### Versions

- Browser/version: Brave

### Settings

### Notes


Branzeturi:
@ryanbr I have a question: Does Brave's ad blocker use uAssets filters by default?

ryanbr:
yes.

Branzeturi:
Ok, thanks!

PR #13596, opened 2022-06-06
Addressed in PersianBlocker and dead links

MasterKia:
`zarpop.com` (Persian popup site) and `adsima.net` (Persian ads site) are addressed in PersianBlocker.
`subtak.ir` and `fardasub.pw` seem to be dead Persian sites.


MasterKia:
If the policy is to keep _**general**_ ads/popup sites in the stock lists, then I could add [more](https://github.com/MasterKia/PersianBlocker/blob/main/PersianBlocker.txt#L31-L265).

PR #14119, opened 2022-07-24
Redundant entries

peace2000:
Removing redundant entries.

I'm currently handling that issue in the Finnish list temporarily and also reported it to the Easylist maintainers:

https://github.com/easylist/easylist/issues/12681

Better to fix the issue in it's original list. :)


mapx-:
this filter `@@||v.fwmrm.net^*ad/g/*_Supla_Audio_HTML5_Live$script` is the "same" with `Nelonen` one ?

peace2000:
@mapx- Yes, it does the same thing.

PR #14178, opened 2022-07-30
Update filters.txt

MasterKia:
This site has ad-reinsertion and heavy anti-adblock so _regular_ cosmetic filters don't work (They either trigger the anti-adb or the ads is just re-inserted).

The site admin has a long history of changing CSS names and hardening the anti-adb on a _**daily**_ basis so I believe it's beneficial to have these filters in uAssets.
`https://github.com/uBlockOrigin/uAssets/issues/10883`

Edit: The site is high-traffic in Iran.

This is happening **_everyday_** to make filters useless: 
b7d9x => b8d9x => b8d10x => b8d11x => ..
sideb5r => sideb6r => sideb7r => ...
آسیاتک آسـیـاتـک آســیــاتــک  ...

I'll keep the filters updated daily.

PR #14184, opened 2022-07-31
Update #14178

MasterKia:


MasterKia:
Added to uBlock filters because Brave doesn't have the quick fixes list yet.

https://github.com/brave/adblock-resources/issues/75
@ryanbr

PR #14216, opened 2022-08-02
Quick fixes: adjust expires grammar

teddybradford:
### Describe the issue

Fixes `Expires:` grammar: "1 days" → "1 day"


MasterKia:
The change doesn't seem to interfere with uBO's "! Expires: " regex.
https://github.com/gorhill/uBlock/blob/master/src/js/storage.js#L932

PR #14279, opened 2022-08-06
Update quick-fixes.txt

MasterKia:


krystian3w:
Add enforce inline block to avoid jump word to origin pop-up line.

https://www.w3schools.com/cssref/pr_class_display.asp

MasterKia:
Could elaborate using an example?

krystian3w:
Yes, for your [pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements).

MasterKia:
I tried but I can't spot the difference.

> to avoid jump word to origin pop-up line.

Could you send screenshots from A vs B?

A) `soft98.ir##body > :nth-of-type(1):before:style(content: "Some text" !important; color: #ff0000 !important; visibility: visible !important; font-size: large !important;)`

B) `soft98.ir##body > :nth-of-type(1):before:style(content: "Some text" !important; color: #ff0000 !important; visibility: visible !important; font-size: large !important; display: inline-block; !important;)`

krystian3w:
More mean toast pop-ups, but for now gone, so I can only draw glitch by "rectangles":

obraz

I try move 3-rd "rectangle" above white text:

obraz

PR #14474, opened 2022-08-21
Modify list titles: clarify Origin compat & brand

wavygecko:
- Clarifies that these filters are Origin


mapx-:
Let's leave them as they are now.
The group of those lists is called "Built-in" indicating lists managed by uBlock (Origin) itself.

quantumpacket:
This probably should have been _uBlock Origin_, it's more obvious to the end-user and doesn't introduce non-ASCII chars like the subscript and the em-dash.

My IDS was alerting me to some suspicious characters in the list file, and I narrowed it down to this commit adding the non-ASCII chars to the list title. Just wanted to make y'all aware of this. :man_shrugging:

PR #14572, opened 2022-08-28
Not needed in uAssets

Infrastructural:
most of the filters which were removed in this PR's commit were present in EasyList

I think `||eephaush.comhttps^` has a typo and `||eephaush.com` is included in EasyList


ghajini:
`||eephaush.comhttps^`
is not a typo

PR #14695, opened 2022-09-05
Create .gitattributes aka hightlighter (Linguist)

krystian3w:
Hightlighter based on https://github.com/github/linguist/pull/5968

Possible see on: https://github.com/ameshkov/VscodeAdblockSyntax/blob/master/test_rules.txt

https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes

PR #14699, opened 2022-09-06
GitHub Linguist support for adblock rules

scripthunter7:
GitHub has been supporting adblock syntax highlighting since September 5th. This PR will help you turn it on.

Further informations:
- Syntax highlighter repository: https://github.com/ameshkov/VscodeAdblockSyntax
- How Linguist works: https://github.com/github/linguist/blob/master/docs/how-linguist-works.md
- Linguist overrides: https://github.com/github/linguist/blob/master/docs/overrides.md
- Linguist commit: https://github.com/github/linguist/commit/e78ef71af3600f96b9a40a06511ebbe3797e7401

**Please note that the following rules will be displayed as invalid:**

```adblock
! Skipped arguments
example.com##+js(scriptlet, , arg2, arg3)
example.com##+js(scriptlet, arg2, , arg3)
```

This is a known issue and will be fixed in the next Linguist release (in about ~2 months). Related highlight PR: https://github.com/ameshkov/VscodeAdblockSyntax/pull/50

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.txt linguist-language=AdBlock linguist-detectable


scripthunter7:
Oh, sorry, duplicated

MasterKia:
Let's keep this open.

MasterKia:
Let's keep this open.
@gorhill Could you check this?

.gitattributes
```diff
@@ -0,0 +1 @@
+*.txt linguist-language=AdBlock linguist-detectable
```
MasterKia, line 1: @scripthunter7 I noticed that `[uBlock Origin]` can be added to the beginning of each list to get syntax highlighting there.

So can/should the `linguist-language` here be "uBlock Origin" as well?

MasterKia:
@gorhill Could you check this?

.gitattributes
krystian3w: Yes, these use Finnish list.

scripthunter7: @MasterKia If the `[uBlock Origin]` agent is present at the beginning of every file, it is not necessary. 

But if you want statistics like this, then you need `linguist-detectable`
image

```gitattributes
*.txt linguist-detectable
```

krystian3w:
PS. this down procentage shell syntax form 100% to 5,56% after cover all txt files.

IMG_20220906_104652


If we cover only few txt files then down less to 7,14%.

.gitattributes
MasterKia: I meant something like `*.txt linguist-language="uBlock Origin"`

scripthunter7: > I meant something like `*.txt linguist-language="uBlock Origin"`

This is simply not possible as Linguist doesn't know such a language or alias.
See https://github.com/github/linguist/blob/bf853f1c663903e3ee35935189760191f1c45e1c/lib/linguist/languages.yml#L230-L242

MasterKia: Also uBlock Origin is not an "**_ad_** blocker", so personally I don't like that statistics.

scripthunter7: > Also uBlock Origin [is not an "**_ad_** blocker"](https://github.com/gorhill/uBlock/#readme)

Neither is AdGuard, but their collective name is adblock

MasterKia: AdGuard literally calls itself "Ad Blocker":
https://github.com/AdguardTeam/AdguardBrowserExtension#ad-blocker-with-advanced-privacy-protection-features

MasterKia: > but I'm actually glad that we finally managed to add grammar to Linguist.

Thank you for the going through the heavy-lifting to add this feature.

MasterKia: > but I'm actually glad that we finally managed to add grammar to Linguist.

Thank you for going through the heavy-lifting to add this feature.

krystian3w: We can disable detect shell language in stats by same file.

Later you can same for Persian repo.

scripthunter7: Yes, but AdGuard can also block a lot of things, not just ads. Actually I don't want to do justice in this, I just wanted to help with the highlighter. 🙂 Perhaps you can try to [submit several aliases to Linguist](https://github.com/github/linguist/blob/bf853f1c663903e3ee35935189760191f1c45e1c/lib/linguist/languages.yml#L236-L240), eg. `adblock plus`, `ublock` / `ublock origin`, `adguard`, but the name of the language remains "Adblock Filter List" in that case as well.

krystian3w: So this must wait for alias and disable generate stats as long don't count as other section "Wide spectrum filter list".

MasterKia: > So this must wait

Or just add `[uBlock Origin]` to each file.

scripthunter7: As far as I know, the statistics always use the original names of the languages, i.e. in this case, even if you enter an alias in the .gitattributes file, the statistics will still show "Adblock Filter List"

krystian3w:
To wide see gorhill words.

https://github.com/uBlockOrigin/uAssets/pull/14695#issuecomment-1238058786

> > Possible see on: https://github.com/ameshkov/VscodeAdblockSyntax/blob/master/test_rules.txt
> 
> This filter below is valid in uBO, it will convert to ASCII at compile time:
> 
> ```
> пример.рф##banner
> ```
> 
> I think it's fair to let filter list maintainers use readable domain names.

MasterKia:
Also happens for Persian letters:
```adb
пример.рф##banner

درود.ir##.ads
```

MasterKia:
Why is it a deal-breaker?

scripthunter7:
> Why is it a deal-breaker?

The skipped scriptlet argument affects many more rules than the unicode domain. But both are known issues and they will hopefully be resolved in the next version.

scripthunter7:
@krystian3w @MasterKia @gorhill @mapx- 
The new Linguist release came out today, which already uses the improved highlighter:

```adblock
example.com##+js(scriptlet, , arg2, arg3)
пример.рф##banner
```

Please note that it may take some time for the new version to go live everywhere.

If you find any more highlighter issues, please report them here: https://github.com/ameshkov/VscodeAdblockSyntax/issues

GitHub Linguist [is published approximately every quarter](https://github.com/github/linguist/discussions/6154#discussioncomment-4028606), and the latest versions of the grammars are downloaded before each release.

krystian3w:
I think gray is bad into blockquote:

```adb
||example.org^$all,domain=com,denyallow=icu
/foo
! comment
```

> ```adb
> ||example.org^$all,domain=com,denyallow=icu
> /foo
> ! comment
> ```

In "CSS4" fix can be:

```css
blockquote:has(.highlight-text-adblock) { color: revert !important; }
```

But IDK about this think "Microsoft".

scripthunter7:
@krystian3w The issue you presented is also true for JS:

```js
// comment
let something = 2
```

> ```js
> // comment
> let something = 2
> ```

I think this should be fixed in the GitHub theme.

Issue #2764, opened 2018-07-04
serialomaniak.pl


lukjan32:
no icons are available for articles under the text, on the site serialomaniak.pl

![1](https://user-images.githubusercontent.com/17565773/42278614-b6819a1e-7f9b-11e8-8ff8-0628ebc18d8b.jpg)

as I add to the exceptions

![2](https://user-images.githubusercontent.com/17565773/42278790-29f870b2-7f9c-11e8-9f84-c9860a62ad13.jpg)

opera 54, ublock origins

zaphodbeebblebrox:
Polish sites should be reported to MajkiIT/polish-ads-filter. @MajkiIT

hawkeye116477:
@zaphodbeebblebrox It's issue with uBO Privacy.

mapx-:
not only that 1 (for me)

test
```
||addthis.com^$important,third-party,domain=~missingkids.com|~missingkids.org,badfilter
||addthis.com/*/addthis_widget.js$script,badfilter
||addthis.com^$important,third-party,domain=~missingkids.com|~missingkids.org|~serialomaniak.pl
||addthis.com/*/addthis_widget.js$script,domain=~serialomaniak.pl
@@||addthis.com/live/$script,domain=serialomaniak.pl
@@||addthisedge.com/live$script,domain=serialomaniak.pl
```

lukjan32:
please add

krystian3w:
No works for my side in Opera 54 and uBO 1.16.6.

Show when unlock `xmlhttprequest` from `m.addthisedge.com`:

```
||addthis.com^$important,third-party,domain=~missingkids.com|~missingkids.org,badfilter
||addthis.com/*/addthis_widget.js$script,badfilter
||addthis.com^$important,third-party,domain=~missingkids.com|~missingkids.org|~serialomaniak.pl
||addthis.com/*/addthis_widget.js$script,domain=~serialomaniak.pl
@@||addthis.com/live/$script,domain=serialomaniak.pl
@@||addthisedge.com/live$script,domain=serialomaniak.pl
@@||m.addthisedge.com/live/*$xmlhttprequest,domain=serialomaniak.pl
````

lukjan32:
as I added it works, please add to the filters

okiehsch:
This will not be fixed in `uAssets`.

Quote from gorhill:
[...]

PR #14789, opened 2022-09-12
Removed addthis.com, included in Easyprivacy

ryanbr:
Cleaned up unused filters, since its now included in Easyprivacy. any fixes can land in Easyprivacy

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -66,7 +66,6 @@
 ! https://www.eff.org/deeplinks/2014/07/white-house-website-includes-unique-non-cookie-tracker-despite-privacy-policy
 ! https://github.com/uBlockOrigin/uAssets/issues/1713
 ! https://github.com/uBlockOrigin/uAssets/issues/6319
-||addthis.com^$important,3p,domain=~amd.com|~missingkids.com|~missingkids.org|~sainsburys.jobs|~schonmagazine.com|~sitecore.com|~plotaroute.com
 ! https://github.com/gorhill/uBlock/issues/1384
 ! https://github.com/uBlockOrigin/uAssets/issues/11003
 ||addthis.com/*/addthis_widget.js$script,redirect=addthis.com/addthis_widget.js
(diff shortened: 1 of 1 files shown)


mapx-:
still needed (not only scripts but frame too)

```
@@||addthis.com/js/*/addthis_widget.js$script,domain=schonmagazine.com
@@||addthis.com/live/$script,domain=schonmagazine.com
@@||addthis.com/static/$subdocument,domain=schonmagazine.com
```

mapx-:
needing 
`@@||addthis.com/*-angularjs.min.js$script,domain=missingkids.*`

ryanbr:
Unless it breaks the site itself, then agreed it should be addressed with a fix. Because we'll have every site complaining "my addthis icons" aren't working, add to the Easyprivacy like x, y domain.

mapx-:
> addthis is a social icon tracking

probably you are right .. @gorhill @okiehsch  your opinion ?

okiehsch:
https://github.com/uBlockOrigin/uAssets/issues/2764#issuecomment-402854508

"If the breakage is only the Addthis widget failing to load, I rather not have the fix in uBO's filter lists -- or else the logical consequence is to remove addthis.com from "uBlock filters -- Privacy", which is not something I see as a good thing."

mapx-:
ok, so `schonmagazine` is not broken but `missingkids.com/.org` does

ryanbr:
Whats the issue with missingkids?

mapx-:
> Whats the issue with missingkids?

kill uBO filter

`||addthis.com^$important,3p,domain=~amd.com|~missingkids.com|~missingkids.org|~sainsburys.jobs|~schonmagazine.com|~sitecore.com|~plotaroute.com,badfilter`

and go to 
`https://www.missingkids.org/poster/USVA/VA22-2926/1/screen`

needing
`@@||addthis.com/*-angularjs.min.js$script,domain=missingkids.com|missingkids.org`

PR #14942, opened 2022-09-20
Added id.group-ib.com

Saiv46:
Some Group-IB customers integrates cross-site tracking in their websites, it works by creating invisible iframe with URL like `https://ru.id.group-ib.com/id.html`, which executes following:

```js
a();setInterval(a,500);function a(){try{window.parent.postMessage(JSON.stringify({action:"gcfids",data:{id:"IAYEOzT+eHm1f3gfPkOkuxSYI1jkVJvUMF6G7cIdW45wy8tKv8c49o3musEilBcxE2epx1B4rUSNhb7GM83dXjRUW7I4QTqzk8SVLIxB6yEtgEKhOh3cKs+WPbLm"}}),"*")}catch(b){}};
```

This rule prevents such cross-site tracking without breaking websites itself (it may affect "fraud" scope for user, but website mostly resolve that with harder captcha or other verification methods).


ryanbr:
From which sample website can this commit reference?

Issue #14688, opened 2022-09-05
titantv.com: breakage


MichaelEnge:
### Prerequisites

- [X] I read and understood the [policy as to what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets)
- [X] I verified that the issue has not already been reported (use [this button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out)
- [X] I forced an update of my filter lists (**how to do this:** [click _"Purge all caches"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches), then [click _"Update now"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now))
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/585534/91296476-ab9d8c00-e76a-11ea-9fb9-87d6861bdf73.png), or I verified that the issue is not caused by removing any of the default lists
- [X] I did not enable external filter lists, or I verified that the issue still occurs without enabling external filter lists
- [X] I do not have custom filters/rules, or I verified that the issue still occurs without custom filters/rules
- [X] I am not using [another content blocker along uBO](https://twitter.com/gorhill/status/1033706103782170625) (also mind browser built-in blockers)
- [ ] I did not answer truthfully to **all** the above checkpoints

### URL address of the web page

`https://www.titantv.com/Default.aspx?r=t`

### Category

breakage

### Describe the issue

Once again, TitanTv tv guide page loads fine when you first open it. Then, if you hit any of the ">" buttons to go to the next time frame it does so, but then has a "Loading, Please wait" popup box that stays on the screen, and seems to lock it so you can't click on anything after that point.

You can disable uBlock Origin, and the page loads fine, however it's so LOADED with ads and such, that alone makes it almost impossible to use.

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

![image](https://user-images.githubusercontent.com/80274754/188477474-8f2c67e5-6138-4645-a278-3e9c6df1485c.png)

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.44.0
Firefox: 104
filterset (summary): 
  network: 82839
  cosmetic: 42322
  scriptlet: 17748
  html: 734
listset (total-discarded, last updated): 
  default: 
    user-filters: 31-0, never
[...]

Issue #13595, opened 2022-06-06
Ad in TitanTv's listing grid has re-surfaced


afab21:
### Prerequisites

- [X] This is not a [support issue or a question](https://old.reddit.com/r/uBlockOrigin/)
- [X] I performed a [cursory search of the issue tracker](https://github.com/uBlockOrigin/uAssets/issues) to avoid opening a duplicate issue
- [X] I am running the latest version of uBlock Origin

### I tried to reproduce the issue when...

- [X] uBO is the only extension
- [X] uBO with default lists/settings
- [X] using a new, unmodified browser profile

### URL(s) where the issue occurs

https://www.titantv.com/default.aspx/

### Describe the issue

okiehsch provided a fix about a month ago that worked until last week in https://github.com/uBlockOrigin/uAssets/issues/13132.  There has apparently been a change in the ad that broke that fix.

### uBlock Origin version

1.424

### Browser name and version

Firefox 101.0

### Settings

Only adding the above okiehsch fix to My Filters.

afab21:
Sorry, that fix did not work.

PR #15008, opened 2022-09-23
update #13595 #14688

sugarfreesweetener:
@ryanbr @pgl `https://www.titantv.com/default.aspx/` `pointlesshour.com`


pgl:
> @ryanbr @pgl `https://www.titantv.com/default.aspx/` `pointlesshour.com`

Added as admiral domain, thanks.

Issue #14999, opened 2022-09-22
vtm.be: detection


hftjyg:
### Prerequisites

- [X] I read and understood the [policy as to what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets)
- [X] I verified that the issue has not already been reported (use [this button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out)
- [X] I forced an update of my filter lists (**how to do this:** [click _"Purge all caches"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches), then [click _"Update now"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now))
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/585534/91296476-ab9d8c00-e76a-11ea-9fb9-87d6861bdf73.png), or I verified that the issue is not caused by removing any of the default lists
- [X] I did not enable external filter lists, or I verified that the issue still occurs without enabling external filter lists
- [X] I do not have custom filters/rules, or I verified that the issue still occurs without custom filters/rules
- [X] I am not using [another content blocker along uBO](https://twitter.com/gorhill/status/1033706103782170625) (also mind browser built-in blockers)
- [ ] I did not answer truthfully to **all** the above checkpoints

### URL address of the web page

`https://vtm.be/vtmgo/afspelen/eaa4c11e3-26c8-4a7f-8877-9a456f40c319`

### Category

detection

### Describe the issue

detects ad block

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.44.4
Chromium: 105
filterset (summary): 
  network: 84323
  cosmetic: 41827
  scriptlet: 17884
  html: 0
listset (total-discarded, last updated): 
  default: 
    user-filters: 6-1, never
    NLD-0: 2437-14, 2d.5h.11m
    easylist: 63580-55, 4d.18h.18m
    easyprivacy: 30595-1172, 4d.18h.17m
    plowe-0: 3674-861, 6d.20h.32m
    ublock-abuse: 76-0, 6d.20h.35m
    ublock-badware: 4205-105, 8h.13m
    ublock-filters: 32865-152, 8h.12m
    ublock-privacy: 245-1, 6d.20h.34m
[...]

PR #15025, opened 2022-09-24
Remove unnecessary lines

BPower0036:
None / `vtm.be`

### Describe the issue
These lines can be removed, because they are already applied in the regional list. (see: https://github.com/EasyDutch-uBO/EasyDutch/commit/4f365269ca81d6e6e88376b19a78ab1adad2dc77 and https://github.com/EasyDutch-uBO/EasyDutch/commit/d2794926688f49a8bde74a1875d5c1948b36ce79)
Because `||fwmrm.net^$important,domain=vtm.be ` is also removed from uBO List (see https://github.com/uBlockOrigin/uAssets/commit/d49ff440a4e187dcde31de9f2fa69e076e330919), there is no need for those rules to stay in `quick-fixes.txt`

### Versions

- Browser/version: FireFox 105.0.1
- uBlock Origin version: 1.44.4

### Notes
See also https://github.com/uBlockOrigin/uAssets/issues/14999

### DIFF
--- a/filters/quick-fixes.txt
+++ b/filters/quick-fixes.txt
@@ -204,7 +204,3 @@ next-episode.net##+js(aost, document.createElement, inlineScript)
 
 ! https://github.com/easylist/easylist/commit/c4e35a0a60dc4a5a2a822f9c425afc358db19971#commitcomment-84395694
 bing.com#@##bepfo
-
-! hotfix https://github.com/uBlockOrigin/uAssets/issues/14999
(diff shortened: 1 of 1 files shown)


mapx-:
well well well well, vtm is an important site, quick fixes is ...quickly updated for all users, so I would have waited 3-4 days before removing filters

BPower0036:
Understand. My list updates only after 3 days

PR #15376, opened 2022-10-22
Add jewishhomelife-pdf.gq - Phishing against Adobe

actually-akac:
### URLs
```
http://jewishhomelife-pdf.gq/adobe/secure/document/
```

### Domains
```
jewishhomelife-pdf.gq
```

### Description

The domain _jewishhomelife-pdf.gq_ is running a phishing attack against **Adobe**.

Screenshot


ghajini:
how to reproduce? which website has this?

krystian3w:
Page banned in Google badware risk project so not really needed add to badware.txt.

I agree asking reproduce of opening - maybe EasyList can add new generic filter `##.spy-ad` and problem gone as virus ad mutate.

PR #15381, opened 2022-10-22
Add badware url

thegu5:
verifyhypixel.net : _Another_ 'verify' phishing page that maliciously takes advantage of Microsoft's oauth system to steal Minecraft accounts (at least for enough time to steal in-game items). Identical to #15197. The owners of this page seem to have switched TLDs from com (#15364) to net.

(There's a constant stream of new websites like this, that's why I don't condense everything into a single pr)


pandaninjas:
It's dead already 💀

Issue #15552, opened 2022-11-06
accounts.o2.co.uk: breakage


1Mark:
### Prerequisites

- [X] I read and understood the [policy as to what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets)
- [X] I verified that the issue has not already been reported (use [this button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out)
- [X] I forced an update of my filter lists (**how to do this:** [click _"Purge all caches"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches), then [click _"Update now"_](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now))
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/585534/91296476-ab9d8c00-e76a-11ea-9fb9-87d6861bdf73.png), or I verified that the issue is not caused by removing any of the default lists
- [X] I did not enable external filter lists, or I verified that the issue still occurs without enabling external filter lists
- [X] I do not have custom filters/rules, or I verified that the issue still occurs without custom filters/rules
- [X] I am not using [another content blocker along uBO](https://twitter.com/gorhill/status/1033706103782170625)
- [X] I verified the issue is not caused by the browser built-in blocker, or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through VPN)
- [X] I did not answer truthfully to **all** the above checkpoints

### URL address of the web page

`https://accounts.o2.co.uk/resetpassword/otac`

### Category

breakage

### Describe the issue

This text box is missing 
![image](https://user-images.githubusercontent.com/15360245/200183469-aa0cbce1-5456-4498-9b69-03a85eb51c08.png)

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

ublock Enabled 
![image](https://user-images.githubusercontent.com/15360245/200182837-b1d1704a-5d2b-4f5d-8016-455b5c2813aa.png)

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.44.4
Chromium: 107
filterset (summary): 
  network: 89686
  cosmetic: 38443
  scriptlet: 16629
  html: 0
listset (total-discarded, last updated): 
  default: 
    user-filters: 2-0, never
    easylist: 63276-56, 1d.18h.13m
[...]

PR #15553, opened 2022-11-06
Fix https://github.com/uBlockOrigin/uAssets/issues/15552

AntiAdvert:
Added the second rule in case there'll be an exception rule for blocking, and that would not be wanted for uBO

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -4764,3 +4764,7 @@ lemonde.fr#@#.gdpr-lmd-wall
 
 ! https://th4-app.taximail.com/campaign-detail/3#overview UI breakage
 @@||taximail.com/js/analytics.js$script,1p
+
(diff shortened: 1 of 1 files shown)


okiehsch:
Why should there be an exception?
 I don't have an account but if I understand correctly the issue is caused by the the redirection, which is unique to uBO, not the blocking of the request.

AntiAdvert:
In case it gets whitelisted in EasyList, for fixing the breakage.

mapx-:
I need the filters here: https://github.com/uBlockOrigin/uAssets/issues/15552#issuecomment-1304846599

AntiAdvert:
https://github.com/uBlockOrigin/uAssets/pull/15553#issuecomment-1304846992

AntiAdvert:
I thought the filters helped, but after deleting cookies + refreshing with shift + refresh afterwards I found out the filters weren't how I expected them to be.

PR #15676, opened 2022-11-14
shopify.com: ads 

Zweih:

`https://thescribes.co/XXXXX/orders/XXXXX`

### Describe the issue

Up-sell advertisements and trackers from appears on checkout pages for Shopify.  

### Screenshot(s)

chrome

### Versions

- Browser/version: Chrome 107.0.5304.110
- uBlock Origin version: 1.45.2

### Settings

- Added filters for Shopify up-sells advertisements and trackers.

### Notes

### DIFF
--- a/filters/filters-2022.txt
+++ b/filters/filters-2022.txt
@@ -4130,6 +4130,12 @@ pornhubed.com##+js(aopr, __Y)
 stream4free.live##+js(aopw, adcashMacros)
 stream4free.live##.closeButton
 
+! Upsell advertisements embedded into Shopify pages
+||launcher.disconetwork.com/
+||widget.disconetwork.com/
+||widget2.disconetwork.com/
+||consumer.discornetwork.com/
+
 ! https://github.com/AdguardTeam/AdguardFilters/issues/126628
 softcobra.*##+js(set, canRunAds, true)
 ! https://www.reddit.com/r/uBlockOrigin/comments/ya0wyc/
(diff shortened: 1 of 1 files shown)


Zweih:
> The filters should be added to Easylist.

Noted, I will move them over promptly.

> Could you provide a test account to access the site?

Unfortunately this screen is only visible after making a purchase, there is no secure way to make a test account for this.

mapx-:
@ryanbr

ryanbr:
I didn't see all the `disconetwork`, but this should help on the one I saw;  Doesn't seem to be an advert?

`https://widget2.disconetwork.com/static/coopcommerce-pixel.js?publisher_id=7185ff86-7454-4538-a289-2e2de5ba11b7&environment=production&&shop=the-scribes.myshopify.com`

Zweih:
> Doesn't seem to be an advert?They're targeted up-sells from a third party embed.

Zweih:
> Doesn't seem to be an advert?

They're targeted up-sells from a third party embed, it seems to specialize in tracking customers across [different Shopify storefronts and domains](https://www.disconetwork.com/resources/disco-now-integrates-with-woocommerce).  

> Having Disco integrated with your WooCommerce store allows you to leverage the power of Disco conversion tracking to understand your shoppers’ purchasing behavior. Disco uses email, phone number, device ID, WooCommerce Customer ID, among other customer properties to follow the user path and WooCommerce order data to account for purchases. This way, you’ll be able to understand which clicks and conversions are Disco-driven.

PR #15692, opened 2022-11-17
fix: skk.moe fetch method param

NeverBehave:

`https://blog.skk.moe/`

### Describe the issue

The params in the fetch is lower case while the filter is upper case.

### Notes

I believe the root cause is because `no-fetch-if` does not support case insensitive (require extra params if build regex from string), while in `fetch()` you could pass `PoSt` as method while still have correct behavior but bypass the filter. A separate issue for uBlock may be required.

Related https://github.com/gorhill/uBlock/blob/b6ed83bc5c840431ed03cddaed1daeb395db3b0e/assets/resources/scriptlets.js#L609

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -332,7 +332,7 @@ reddit.com##+js(no-xhr-if, method:POST url:/^https:\/\/www\.reddit\.com$/)
 ||zhihu.com^$removeparam=hybrid_search_extra
 
 ! https://github.com/easylist/easylist/issues/6724#issuecomment-1003172754
-skk.moe##+js(no-fetch-if, method:POST)
+skk.moe##+js(no-fetch-if, method:post)
 /cfga/jquery.js?$image
 /npm/sks@0.*/lazyload.js$script,3p
 


NeverBehave:
Currently, `method: Post` is used and this filter is thus invalidated. Just a head up for anyone who cares about this specific site.

mapx-:
does this help ?

`skk.moe##+js(no-fetch-if, body:/[\w\W]{700}/)`

NeverBehave:
I am not entirely sure what this filter is based on (a key in the body?). I would like to find a way to stop a kind (verb) of request (reading a blog imo doesn't need to post anything in the background), and right now I can't find any effective way to achieve this goal.

mapx-:
Test for breakage

`skk.moe##+js(aopr, btoa)`

or

`skk.moe##+js(aost, Number.isNaN)`

I see no POST anymore.

NeverBehave:
`skk.moe##+js(aopr, btoa)` this one seems work well as only tracking used it.

SukkaW:
@mapx- 

And you are just breaking following page:

- https://hv-monster.skk.moe/ (Charts stop working)
- https://lab.skk.moe/hentaiverse-expgraph

Those pages are using [ECharts](https://echarts.apache.org/en/index.html) library which uses `window.btoa`.

uBlock-user:
> I believe the root cause is because no-fetch-if does not support case insensitive (require extra params if build regex from string), while in fetch() you could pass PoSt as method while still have correct behavior but bypass the filter. 

@gorhill this needs to be addressed...

stephenhawk8054:
`##+js(no-fetch-if, method:/post/i)` will work.

uBlock-user:
> ##+js(no-fetch-if, method:/post/i) will work.

It doesn't.

gorhill:
What filter doesn't work where?

uBlock-user:
@gorhill Add `blog.skk.moe##+js(no-fetch-if, url:post)` and/or `blog.skk.moe##+js(no-fetch-if, method:/post/i)` and browse to `blog.skk.moe`, observe that fetch request of `method:post` is fired and with url having `post` string in it despite having the filter/s to defuse both. Both filters get bypassed.

gorhill:
The site uses an object with non-enumerable properties as the fetch's `options` argument, so the scriptlet can't "see" these options. Maybe one way to counter this is to at least explicitly lookup common properties instead of just using `Object.assign`.

SukkaW:
> The site uses an object with non-enumerable properties as the fetch's `options` argument, so the scriptlet can't "see" these options. Maybe one way to counter this is to at least explicitly lookup common properties instead of just using `Object.assign`.

Let's leveling the playing field a little bit: one of the proper ways to do this is to use `Object.prototype.hasOwnProperty.call` to look up the property first, then explicitly reading the value, which is what AdGuard's scriptlet does. `Object.getOwnPropertyNames` could also be a good alternative if you wish to stay with `for ... in ...`.

exzzzzzzz:
[1.67.1b5](https://github.com/gorhill/uBlock/releases/tag/1.67.1b5) This version broke the test. https://www.cloudflare.com/ru-ru/ssl/encrypted-sni/#results

PR #16502, opened 2023-01-26
Remove generic filter

NFSS10:

Any website that contains brand and model params, example:

`https://www.example.com?brand=a&model=b`
`https://www.google.com/?brand=a&model=a`

### Describe the issue

The filter `?brand=*&model=$doc` is too generic to be present in the filter list. It will flag every URL for just containing the `brand` and `model` params which shouldn't be a reason to flag them.

### Screenshot(s)

example_1
example_2

### Versions

- Browser/version: Firefox 109.0 
- uBlock Origin version: 1.46.0

### Settings

No changes

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -804,7 +804,6 @@ okashimag.com,tech4yougadgets.com##+js(aopr, Notification)
 ||adultonlineplay.com^$all
 ||fastandslut.com^$all
 &click_price=0.*&click_id=$doc
-?brand=*&model=$doc
 ||familialsimulation.com^$all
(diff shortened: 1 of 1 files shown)


Yuki2718:
No *real* example?

Yuki2718:
Although I admit the rule may be bit too generic, there's no other least common factor in those malicious URLs actually used in the wild. I'll add alternate rules.

NFSS10:
> No _real_ example?

Those examples show the generic nature of the issue, but a more specific one could be for example this URL: https://white.platforme.com/?brand=swear&model=vyner which was the one that caught my attention for the problem.

For context, this website is just a customization platform used by some stores to offer customization to their clients. The filter would just flag this website when I wanted to customize my shoes.

PR #16705, opened 2023-02-11
Replace many `:upward(selector)` rules with `selector:has()`

antonok-edm:
Procedural filters are relatively expensive to query, as they require running an in-page `MutationObserver`. The more recent native `:has()` selector allows many procedural filters to be rewritten as just native CSS selectors, which can be handled by browsers very efficiently. Chromium and Safari have support for native `:has()`; Firefox has it behind a flag and should have it relatively soon as well.

`a:upward(b)` is almost exactly equivalent to `b:has(a)`, with a couple of caveats:
- `a:upward(b)` will select the first `b` element above `a`; `b:has(a)` selects _all_ of them. `a:has(> b)` is a very useful tool.
- Lots of `:upward` filters were originally written as `a b:upward(a)` or similar, which should simplify to `a:has(b)` instead of `a:has(a b)`. This actually reduces duplication in the lists.

This PR replaces most of the `:upward` filters that use a selector rather than a number. The few remaining ones are more challenging, but could be handled as followups. I suspect that it might be possible to deprecate the `:upward(selector)` form, since native `:has` and others are quite expressive.

I've manually looked over the relevant sites and these filters seem to be either the same or improved in terms of blocking and breakage, as far as I can tell.

Apologies for the gigantic PR - I'm happy to split this into smaller changes if that makes it easier to review!

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -875,7 +875,7 @@ talkceltic.net##+js(set, samDetected, false)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/7365
 @@||splitshire.com^$ghide
-splitshire.com##.adsbygoogle:upward(.row-container)
+splitshire.com##.adtester-container
 
 ! https://github.com/uBlockOrigin/uAssets/issues/7366
 @@||kwamkidhen.com^$ghide
@@ -2522,7 +2522,7 @@ real-sky.com##+js(aopr, app_vars.force_disable_adblock)
 healthline.com##aside:has(div:matches-css-before(content:/ADVERTISEMENT/))
 healthline.com##div[data-empty^="true"]:matches-css-before(content:/ADVERTISEMENT/)
 healthline.com##[class*="css"]:matches-css-before(content:/ADVERTISEMENT/)
-healthline.com##[href*="redirect"]:upward(section)
+healthline.com##section:has([href*="redirect"])
 healthline.com##hl-adsense
 
 ! https://community.brave.com/t/website-kept-asking-leave-site-if-i-click-cancel-it-opened-new-window-i-couldnt-even-close-brave-without-end-process/
@@ -2921,10 +2921,9 @@ duplichecker.com##[href^="https://www.duplichecker.com/link"]
 duplichecker.com##.home_bg > .container > style + style + div
 duplichecker.com##.clearfix + div > style + style + div
 duplichecker.com##.clearfix + div:has(> style:has-text(min-height:302px))
-duplichecker.com###home_after_title > .adsbygoogle:upward(.main-content > div > *)
-duplichecker.com###home_before_buton > .adsbygoogle:upward(.data_inbox > div)
-duplichecker.com##div[style] > [class^="avds"]:upward(div[style])
-duplichecker.com##div > span[id^="side_bar"] > .adsbygoogle:upward(div)
+duplichecker.com##[class^="avds"]
+duplichecker.com##div[style]:has(> [class^="avds"])
+duplichecker.com##div:has(> span[id^="side_bar"] > .adsbygoogle)
 duplichecker.com##a[style="display: inline-block;"][target="_blank"][rel="nofollow"]
 
 ! https://github.com/AdguardTeam/AdguardFilters/issues/63002
@@ -2998,18 +2997,12 @@ ohentai.org##[class^="listleaderboardcontainer"]
 ohentai.org##.videobrick > .videoadintro:upward(1)
 ohentai.org##+js(nowoif)
 
-! youngpornvideos.com popup ads
-youngpornvideos.com##+js(acis, url)
-youngpornvideos.com##.ads
-youngpornvideos.com##.ads-mobile
-youngpornvideos.com##.ads-thumb-list:upward(.outer-item)
-
 ! https://github.com/uBlockOrigin/uAssets/issues/8988
 hentaicloud.com##+js(set, decodeURI, noopFunc)
 hentaicloud.com##+js(aopr, TotemToolsObject)
 hentaicloud.com##+js(aopr, AaDetector)
 hentaicloud.com##.vertical-ads-content
-hentaicloud.com##.ad:upward(.horizontal-ads-content)
+hentaicloud.com##.horizontal-ads-content:has(.ad)
 hentaicloud.com##section.videos-content:has(.thumbnail > a[href^="https://www.nutaku.net/signup/landing/"])
 
 ! javdoe.to/javsun.cc/javtc.fun popup ads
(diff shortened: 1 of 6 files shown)


gorhill:
`:has()` is still not native in Firefox.

gorhill:
> Procedural filters are relatively expensive to query, as they require running an in-page `MutationObserver`

The mutation observer is still required in uBO for other purpose than procedural filters, so I don't think there is much to gain in uBO and I am actually worried that such an amount of changes will cause unforeseen issues -- there is currently no profiling data showing this would be an measurable improvement.

krystian3w:
Also not all possible convert by newbie method:

https://github.com/AdguardTeam/ExtendedCss/issues/102#issuecomment-619898058

```adb
example.org##.element:upward(2)
! → 
example.org##*:has(> * > .element)
```

Yuki2718:
Native `:has()` doesn't support nested `:has-text()`. Anyway `f.tabtter.jp` is dead and I'll remove this.

antonok-edm:
> there is currently no profiling data showing this would be an measurable improvement.

Fair point. Here is a quick benchmark: adding 100000 elements to the page with random class names, nesting depths, and sibling counts.

## Benchmark setup

### Test page

```html
<!DOCTYPE html>
<html>
  <head>
    <style>
      .class1,.class2,.class3,.class4 {
        padding-top: 1px;
        width: 10px;
        padding-left: 2px;
      }
      .class1 {
        background: #f00;
      }
      .class2 {
        background: #0f0;
      }
      .class3 {
        background: #00f;
      }
      .class4 {
        background: #888;
      }
    </style>
  </head>
  <body>
    <script>
      const nElements = 100000;
      let targetParent = document.documentElement;
      for (let i = 0; i < nElements; i++) {
        const e = document.createElement('div');
        const r = Math.random();
        if (Math.random() < 0.25) {
          e.classList.add('class1');
        }
        if (Math.random() < 0.25) {
          e.classList.add('class2');
        }
        if (Math.random() < 0.25) {
          e.classList.add('class3');
        }
        if (Math.random() < 0.25) {
          e.classList.add('class4');
        }
        targetParent.appendChild(e);
        let t = Math.random();
        if (t < 0.2) {
          targetParent = document.documentElement;
        } else if (t < 0.6) {
          targetParent = e;
        }
      }
    </script>
  </body>
</html>
```

## Filters to test

I've compared the following rules, one at a time:
```adblock
localhost##.class2:has(> .class3)
localhost##.class3:upward(.class3 > .class2)
```

## Results

### Chromium
The difference is noticeable even before collecting any profiling data. With `:has`, there is no layout shift whatsoever, whereas with `:upward`, there's a flash of extra elements before the page load completes. In terms of profiling data, I see ~15% improvement in overall page load time using `:has`.

### Firefox (`layout.css.has-selector.enabled == true`)
I used a custom build of uBlock Origin from the `1.46.0` tag, with this patch to force native `:has` support:
```diff
diff --git a/platform/common/vapi-common.js b/platform/common/vapi-common.js
index 290837584..7d20763f2 100644
--- a/platform/common/vapi-common.js
+++ b/platform/common/vapi-common.js
@@ -55,6 +55,7 @@ vAPI.webextFlavor = {
 
     // This is always true.
     soup.add('ublock').add('webext');
+    soup.add('native_css_has');
 
     // Whether this is a dev build.
     if ( /^\d+\.\d+\.\d+\D/.test(browser.runtime.getManifest().version) ) {
@@ -91,9 +92,6 @@ vAPI.webextFlavor = {
         soup.add('chromium')
             .add('user_stylesheet');
         flavor.major = parseInt(match[1], 10) || 0;
-        if ( flavor.major >= 105 ) {
-            soup.add('native_css_has');
-        }
     }
 
     // Don't starve potential listeners
```

Results are similar to Chromium. No layout shift with `:has`. Noticeable flashing layout shift with `:upward`. ~15% page load time improvement with `:has`.

### Firefox (`layout.css.has-selector.enabled == false`)
`:has` is significantly slower; Firefox briefly brings up a warning about the slow page load before it completes.

## Conclusion

Native `:has` eliminates layout shift and improves performance, making it a definitive winner in this comparison. In a world where all browsers support `:has` natively, we should aim to use it as much as possible.

Of course, the challenge is that it's still only available in Firefox behind a flag. _In practice_, I think the performance difference wouldn't be an issue on real-world sites (this benchmark was slightly overkill). It's probably worth migrating to `:has` at least for lighter sites.

antonok-edm:
> Native `:has()` doesn't support nested `:has-text()`.

Thanks for catching, @Yuki2718. I tried to leave out filters that had other procedural selectors but that one slipped through somehow.

stephenhawk8054:
I guess what gorhill meant with `measurable improvement` is about the real-world changed sites in the commit? I also support the change to native `has`. However, since Firefox has not officially updated in stable release yet, many users who have not enabled that option in `about:config` might be at risk of inefficient current `has` with this amount of sites changed (for example `duplichecker.com##div:has(> span[id^="side_bar"] > .adsbygoogle)`). If the improvement in these real-world cases (not the test case) is not too much, I'm just afraid the gain would not outweigh the risk at the moment.

---

And actually for `duplichecker.com` site, the hiding result is different between 2 cases

```adb
duplichecker.com###home_after_title > .adsbygoogle:upward(.main-content > div > *)
duplichecker.com###home_before_buton > .adsbygoogle:upward(.data_inbox > div)
duplichecker.com##div[style] > [class^="avds"]:upward(div[style])
duplichecker.com##div > span[id^="side_bar"] > .adsbygoogle:upward(div)
```

```adb
duplichecker.com##[class^="avds"]
duplichecker.com##div[style]:has(> [class^="avds"])
duplichecker.com##div:has(> span[id^="side_bar"] > .adsbygoogle)
```

With this amount of sites, I don't know if there are other similar cases.

---

With this site, I think the meaning of these 2 filters are different

```adb
socialcounts.org##.container > div > .adsbygoogle:upward(div)
```

```adb
socialcounts.org##div:has(.container > div > .adsbygoogle)
```

gorhill:
> adding 100000 elements

What I had in mind is actual real world cases, and keeping in mind that Firefox does not have native `:has()` yet. I really do not want the filter lists to be crafted primarily for just Chromium-based browsers.

Yuki2718:
Maybe it should be done in case-by-case basis with one site per one PR so that we can check each case, tho I see this will be heavy burden for PR-er.

krystian3w:
Possible move this to tmp repository to avoid generate new ID here like from 16705 to ≈25xxx.

gorhill:
> Maybe it should be done in case-by-case basis with one site per one PR

No need, Firefox does not support native `:has()`, so our judgment of how to craft efficient filters needs to take this into account when dealing with filters with `:upward()` operator.

Given all what I said above, I prefer to decline this PR.

ryanbr:
When Firefox supports :has could it be reviewed again? @gorhill

krystian3w:
Only as ≈248 PR-s.

gorhill:
> could it be reviewed again?

Yes, but it should be done opportunistically and incrementally (Firefox ESR will take longer to get `:has()`), I don't believe that changing all at once without carefully testing every single cases being changed is a good idea. We have lived years with the current solution, there no emergency to change all for what might be immeasurable benefits in real world cases. I don't understand the urge to convert.

antonok-edm:
@MasterKia in `adblock-rust` specifically, I thought a reasonable first step towards implementation would be to automatically convert `:upward` to `:has` internally whenever possible - which has the nice bonus of only using native CSS syntax and reducing in-page JS footprint. Then I realized, if native `:has` is indeed better than `:upward`, might as well push to get the lists changed too so that everyone can benefit, not just Brave.

Of course, I understand uAssets is a uBlock Origin project and I don't want to push for changes here that don't have technical merit for uBlock Origin itself.

PR #16733, opened 2023-02-13
verkkokauppa.com: breakage

peace2000:

`https://www.verkkokauppa.com/fi/product/778669/Samsung-UE65BU8072-65-4K-LED-televisio/questions/write`

### Describe the issue

Sending a product question is blocked, due to "privacy settings".

### Screenshot(s)

kuva

```
Kysymystä ei voida lähettää!
Selaimellesi asetetut yksityisasetukset estävät kysymyksen lähettämisen. Muuta asetuksia, jotta pääset jakamaan mielipiteesi!
```

Translation:
```
Question could not be send!
Your browser's privacy settings prevent sending the question. Change settings, so that you can share your opinion!
```

### Versions

- Browser/version: FF 109.0.1
- uBlock Origin version: 1.46.1rc0

### Settings

Defaults

### Notes

You'll need an account. The issue is caused by Lowe's list. Also EP causes the issue (I'll make a PR for EP also).

When it's fixed, you'll see this button (it's a button to send a question):

kuva

### DIFF
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -761,6 +761,3 @@ twinfinite.net##+js(acs, document.createElement, admiral)
 
 ! ads.snapchat.com anti-adb
 ads.snapchat.com#@#.banner_ad
-
-! https://github.com/uBlockOrigin/uAssets/pull/16733
-@@||mpsnare.iesnare.com/snare.js$script,domain=verkkokauppa.com
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -761,3 +761,6 @@ twinfinite.net##+js(acs, document.createElement, admiral)
 
 ! ads.snapchat.com anti-adb
 ads.snapchat.com#@#.banner_ad
+
+!
+@@||mpsnare.iesnare.com/snare.js$script,domain=verkkokauppa.com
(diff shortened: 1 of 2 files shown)


peace2000:
Just a sec..

mapx-:
There is already a similar filter in EP. Did you try to catch @ryanbr attention on it ?

`@@||mpsnare.iesnare.com/snare.js$script,domain=power.fi`

peace2000:
@MasterKia I corrected the list now.

@mapx- Yeah I'm aware of that rule, it probably fixes similar issues.
https://github.com/easylist/easylist/pull/14966

krystian3w:
But @DandelionSprout recommend disable EP.

mapx-:
fixed by ep

krystian3w:
But Nordic countries may have disabled the list and can't rate TV bought in Finland shipped to Norway 🇳🇴  / Denmark 🇩🇰  / Iceland 🇩🇰   / Greenland 🇬🇱.

Lock limitation on this PR don't fix page for these group.

DandelionSprout:
Since the breaking entry appears to have been in EasyPrivacy in the first place, I can only presume that the low 3-digits of user in that situation would be entirely unaffected by it.

okiehsch:
If the issue is caused by both EP and Peter Lowe's we should add a fix in unbreak, to make sure that users who diable EP but keep Peter Lowe's list - both enabled by default - have that issue fixed.

PR #17820, opened 2023-04-27
Fix untitle.org right-click, drag

advertblocker:
`https://untitle.org/`

also interesting file: `https://tistory1.daumcdn.net/tistory_admin/userblog/userblog-a7bbf74de6acc4f84635e6cdaf5a8747e9e1e232/static/plugin/PreventCopyContents/script.js`


stephenhawk8054:
Does `untitle.org##+js(acs, document.oncontextmenu)` work for you? Since it's better to not abort write property completely to `document.oncontextmenu`

PR #17846, opened 2023-04-28
update nosteam

advertblocker:
ad blocker detection => in-page notice appears ("Welcome") => popunder after clicking on X


seedoils:
overlay appears even after closing the modal => having to click 2 times to access 1 link

PR #17860, opened 2023-04-29
Fix ideas0419.com right-click

advertblocker:


stephenhawk8054:
You need to remove `!` at beginning of the filter

advertblocker:
Oops, you're right. Could you remove it? (I'm not on my PC)

PR #17863, opened 2023-04-29
Fix geniusjw.com right-click

advertblocker:


stephenhawk8054:
I think your recent sites can be merged into 1 filter since they use the same scriptlet filter.

advertblocker:
> I think your recent sites can be merged into 1 filter since they use the same scriptlet filter.

why don't we make a section for such issues (like with `chp_ad` scriptlet)?

stephenhawk8054:
Ok, I'll create one for the sites with no additional references or no other filters.

PR #17886, opened 2023-05-01
Update specific_report_from_ubo.yml

GH-User999:
Updated default filter list image to reflect more recent options/filter list changes.


gorhill:
Your picture looks blurry compared to the one before.

PR #18167, opened 2023-05-21
babiato. tech anti-adb

bookarticle:


bookarticle:
actually, the user could "lie" to the site

JobcenterTycoon:
The anti adblock reappear after clicking on "I've Disabled AdBlock" so its a case for the main list.

PR #18204, opened 2023-05-23
agfy. io => agfy. to

bookarticle:


JobcenterTycoon:
Why you removing the filter instead of changing the domain to `agfy.to`? The filter are not obsolete.

bookarticle:
good question; i didn't investigate a lot at the time of the commit; and I reported the ads to easlist

PR #18456, opened 2023-06-11
remove autoplay from reddit embeds

jxyzn:

`https://embed.reddit.com/r/oddlyterrifying/comments/1463lck/what_is_happening_here_will_it_go_in_2_pieces_in/`

### Describe the issue

Reddit embedded videos auto-play, with no option to not do so even when creating a new embed from a post. This removes the 'autoplay' attribute from the 'shreddit-player' tag and keeps the player from starting playback. It's now click-to-play.

### Versions

- Browser/version: I tried in Chrome 112
- uBlock Origin version: 1.49.2

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -6800,3 +6800,6 @@ film4e.com,zamundatv.com##+js(acs, AudiosL10n)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/18421
 daysoftheyear.com##+js(aopr, googletag)
+
+! remove autoplay from reddit embeds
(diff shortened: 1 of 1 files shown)


jxyzn (commented):
I noticed that clicking the play button would also open the original reddit post in a new tab. Added another rule to neutralize that.

krystian3w:
On mother pages possible use:

```adb
||embed.reddit.com^$frame,object,redirect=click2load.html
```

If remove attr was started too late.

Direct opening with redirect to `click2load` was added only to AdGuard:

obraz

PR #18550, opened 2023-06-17
Add fextralife twitch stream blocker

AlpinDale:

`https://eldenring.wiki.fextralife.com/Elden+Ring+Wiki`
Any every other wiki page in Fextralife.

### Describe the issue

The fextralife wiki adds a sidebar to every wiki page to display their Twitch stream, if it's currently live. This is a massive waste of bandwidth for most people, as every new tab opens up another instance. This is a huge issue, because people tend to open multiple wiki pages at once. Also highly unethical to inflate twitch viewer count this way.

### Screenshot(s)

Twitch stream to the top left corner

### Versions

- Browser/version: FireFox/114.0.1

### Settings

- Default.

### Notes

This PR adds a filter for blocking the sidebar content and shouldn't affect all Twitch sidebar streams injected into other websites. If enough users complain about similar practices, perhaps this could be added as a filter rule later:
`||player.twitch.tv^$third-party `

### DIFF
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -3025,3 +3025,6 @@ online2pdf.com##div[id*="title_vertical"]
 
 ! https://github.com/uBlockOrigin/uAssets/issues/18534
 burbuja.info##+js(acs, RegExp, $)
+
+! https://github.com/uBlockOrigin/uAssets/issues/11883
+fextralife.com###sidebar-wrapper
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -3027,4 +3027,4 @@ online2pdf.com##div[id*="title_vertical"]
 burbuja.info##+js(acs, RegExp, $)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/11883
-fextralife.com###sidebar-wrapper
+||embed.twitch.tv^$frame,redirect=click2load.html,domain=wiki.fextralife.com


ahmeddots:
Seconding this. It is a massive annoyance.

And as said by @AlpinDale, it hurts my bandwidth and is unethical.

mapx-:
That filter does nothing, hiding only twitch but in background the player is going on

you need 
`||player.twitch.tv/js/embed/$script,domain=eldenring.wiki.fextralife.com`

on the other hand if the channel is fextralife's .. it could be considered as self-promotion
If it's about bandwidth, well could enter in the resource-abuse policy

@okiehsch thoughts ?

AlpinDale:
> That filter does nothing, hiding only twitch but in background the player is going on
> 
> you need `||player.twitch.tv/js/embed/$script,domain=eldenring.wiki.fextralife.com`
> 
> on the other hand if the channel is fextralife's .. it could be considered as self-promotion If it's about bandwidth, well could enter in the resource-abuse policy
> 
> @okiehsch thoughts ?

Thanks for pointing that out. Though for removal, we'd probably have to do the entire `wiki.fextralife.com` subdomain, since the stream embeds aren't restricted to the Elden Ring wiki only.

stephenhawk8054:
I think `Resouce abuse` should only be considered when users have no ways to pause/stop it easilly. Here it's a live video, and the network consuming stops when clicking on stop/pause button.

If we still want to stop the video, I think we should use `click2load` for it, so if others don't want to block the videos, they can click on it.

```adb
||embed.twitch.tv^$frame,redirect=click2load.html,domain=wiki.fextralife.com
```

AlpinDale:
Thanks, I made a new commit and added that.

stephenhawk8054:
Since it's a self-promo, I don't think it should be in the main list either. If we still want to stop the embed frame, the most we should consider is annoyance list.

okiehsch:
> @okiehsch thoughts ?

I agree with `stephenhawk8054` using bandwith unless it's done in secret is not "resource abuse".

It's self advertisement and I don't think that is should be added to the annoyance list.
We only add self advertisement issues to the annoyance list if they are "very" annoying and here I would be OK with adding it if the videos would be unmuted by default but they are muted, so I don't see how you can say that's especially annoying,

PR #18633, opened 2023-06-23
Fixes https://vadhander.hogakusten.com/

ryanbr:
Fixes cookie message showing. Prevents the overlay and fixes disable content.


stephenhawk8054:
@ryanbr Sorry, I just found that it can be solved by normal `set`. Can you test this?

```adb
vadhander.hogakusten.com##+js(set, settings.consent, true)
```

ryanbr:
Looks good, thanks

Issue #18932, opened 2023-07-10
deezer.com: detection


cesarnovo1971:
### Prerequisites

- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I forced an update of my filter lists. (Click the ["Purge all caches"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches) button while holding the 'Shift' key, then click the ["Update now"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now) button.)
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/124740436/235392297-c4fc6290-73a6-4a70-96a9-3f6a7683f21d.png), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable external filter lists, or I have verified that the issue still occurs without enabling external filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [ ] I did not answer truthfully to **ALL** the above checkpoints.

### URL address of the web page

`https://www.deezer.com/`

### Category

detection

### Description

![Imagem da Área de Transferência](https://github.com/uBlockOrigin/uAssets/assets/65466276/62b9ab6d-bc40-486e-ac0f-8c3f07e3df6c)

### Other extensions used

none

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin 1.50.0:
 since: 31d.48m
Chromium: 114
filterset (summary):
 network: 98243
 cosmetic: 43041
 scriptlet: 18282
 html: 0
listset (total-discarded, last-updated):
 default:
  user-filters: 0-0, never
  easylist: 68168-11, now
  easyprivacy: 33422-63, now
  plowe-0: 3740-1074, now
  spa-1: 5332-68, now
[...]

PR #18943, opened 2023-07-11
Fix anti adblock on deezer

Yann-P:
#18932

`https://www.deezer.com`

### Describe the issue

- Deezer uses SmartAds to detect DNS blocking of the sadcdn.com domain
- It pauses the playback and opens a popup
- To detect this blocking it creates a fetch request with an error handler that sets a global property `window.smartLoaded`

### Screenshot(s)

https://user-images.githubusercontent.com/65466276/252509427-62b9ab6d-bc40-486e-ac0f-8c3f07e3df6c.jpg

### Versions

- Browser/version: Firefox 114.0.2

### Notes

They [make a call to this domain and the `onerror` handler sets a flag on the `window` object](https://gist.github.com/Yann-P/a7947422a73399275b25d2914931d706#file-route-naboo-ads-js-L462-L465). This is what they use [later in this script](https://gist.github.com/Yann-P/a7947422a73399275b25d2914931d706#file-route-naboo-ads-js-L553-L556) to trigger the popup and pause playback

(Code retrieved from https://e-cdn-files.dzcdn.net/cache/js/route-naboo-ads.8e773f8740fa6a75b950.js)

~So probably blocking the `route-naboo-ads(.+).js` file in the uBlock list will restore functionality without having to disable DNS blockers.~ no, it causes "chunk loaded error"

`deezer.com##+js(aopr, smartLoaded)` seems to work.

### DIFF
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -524,6 +524,8 @@ compromath.com##+js(no-fetch-if, ads)
 ||cmodmedia*.live.streamtheworld.com/media/*.mp3$media,redirect=noopmp3-0.1s,domain=deezer.com
 ! https://github.com/uBlockOrigin/uAssets/issues/17392
 deezer.com##+js(nostif, Ads)
+! https://github.com/uBlockOrigin/uAssets/issues/18932
+deezer.com##+js(aopr, smartLoaded)
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/10n37i2/
 @@||wnacg.org/download$ghide


stephenhawk8054:
Does this work?

```adb
deezer.com##+js(set, smartLoaded, true)
```

Yann-P:
> Does this work?
> 
> ```
> deezer.com##+js(set, smartLoaded, true)
> ```

Just tested your snippet after removing the aopr, it does works

stephenhawk8054:
I think we should use `set` filter, since `aopr` can cause unexpected error.

Yann-P:
I updated the code with your suggestion

PR #19040, opened 2023-07-16
Block ad leftover on dll-files.com

falseuniversefacts:

`https://www.dll-files.com/api-ms-win-core-path-l1-1-0.dll.html`

### Describe the issue

Ad leftover that says "advertisement" above blocked ads

### Versions

- Browser/version: Tor Browser 12.5.1

### Settings

None, all settings are on default.


stephenhawk8054:
Does this work?

```adb
dll-files.com##center[style]:has-text(advertisement)
```

falseuniversefacts:
Yes, it does, and I changed it to that.

PR #19365, opened 2023-08-16
Add google.com to anti click trackers list

BPower0036:

`https://www.google.com`

### Describe the issue

When clicking on search results, google.com redirects it via their own server. Try `https://www.google.com/search?q=example.com%252Ftest%253Fq%253Dtest` as a test

### Screenshot(s)

### Versions

- Browser/version: 116.0.2
- uBlock Origin version: 1.51.0

### Settings

None of them affects this problem.

### Notes
Maybe, the rule should be narrowed down to only `https://www.google.com/search?` since their are more outgoing links and we don't want to break other links of google


stephenhawk8054:
There's `www.google.*##+js(set, rwt, noopFunc)` in annoyance list.

Does it work for you?

BPower0036:
Yes, it works.
So there is no need to merge this PR.

Yuki2718:
I don't.

stephenhawk8054:
Hmmm... I checked google activity settings and it just shows `search history` (aka search terms used), but not links clicked.

stephenhawk8054:
I mean without that filter. I don't see any settings in google settings that shows the clicked ones (or I haven't found it yet).

stephenhawk8054:
Ok, I'll move it to privacy list. Due to the update gap of both lists, we can keep in both until next week.

PR #19580, opened 2023-09-02
Update revanced.app rule in badware.txt

masterflitzer:

`https://madkarmaa.github.io/revanced/`

### Describe the issue

revanced api (`https://api.revanced.app`) or other official resources cannot be called from other sites because filter is too strict

### Versions

- Browser/version: Firefox Dev (118.0b3)
- uBlock Origin version: 1.51.0

- Browser/version: Edge (116.0.1938.69)
- uBlock Origin version: 1.51.0

- Browser/version: Firefox Android (117.0)
- uBlock Origin version: 1.51.0

### Settings

- ~~move revanced.* rule to the end of revanced block~~
- ~~add exception for revanced.app (the official domain which should be allowed/callable by others)~~
- update revanced.app filter (the official domain which should be allowed/callable by others):
`||revanced.*^$all,domain=~revanced.app` -> `||revanced.*^$all,from=~revanced.app,to=~revanced.app`

### Notes

revanced api is a public api which should be callable by other websites

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -3602,7 +3602,7 @@ torrdroidforpc.com##[href^="http://slugmefilehos.xyz/"]
 ! https://safeweb.norton.com/report/show?url=revanced.io
 ! https://sitecheck.sucuri.net/results/revanced.io
 ! https://github.com/uBlockOrigin/uAssets/pull/17469
-||revanced.*^$all,domain=~revanced.app
+||revanced.*^$all,from=~revanced.app,to=~revanced.app
 ||revancedapk.*^$all,domain=~revanced.app
 ||revancedapp.*^$all,domain=~revanced.app
 ||youtubevanced.*^$all,domain=~revanced.app


stephenhawk8054:
Don't use `@@`. Do these work for you?

```adb
||revanced.*^$all,domain=~revanced.app,badfilter
||revanced.*^$all,from=~revanced.app,to=~revanced.app
```

masterflitzer:
thx for the replies, I am not really familiar with the filter syntax and just tried the first that made sense

I will try it and reply back

masterflitzer:
> ```
> ||revanced.*^$all,domain=~revanced.app,badfilter
> ||revanced.*^$all,from=~revanced.app,to=~revanced.app
> ```

> ```
> @@||api.revanced.app^$xhr,3p
> ```

both of these work

which one would be better? i think the first one because then i avoid `@@` and also don't hardcode the api subdomain which theoretically could change

but can you explain to me what badfilter does in this case?
was that only for local testing or should it be comitted?

stephenhawk8054:
`badfilter` simply means disable the current filter. If you are replacing it, you don't need it.

masterflitzer:
okay thx got it, i updated the PR

PR #19628, opened 2023-09-06
Fix https://github.com/easylist/easylist/issues/17251 without exception

FanOfRootBeer:
- overrides gtm exception
- still unbreaks


mapx-:
I guess you'll break some functionality

FanOfRootBeer:
Should the blocking filter have been more specific?

mapx-:
I guess @dimisa-RUAdList could respond. If it's ok could be added in the regional list.

dimisa-RUAdList:
Ok, fixed in RU AdList: https://github.com/easylist/ruadlist/commit/5677317750757691541ddee827739da6ada5c932

PR #19638, opened 2023-09-06
obsolete filters

FanOfRootBeer:
`skidrowcodexgame.com` => parked
`adictosalatele.com` => parked
`netfile.cc` => parked
`toppnews4.net` => dead site (past domain expiry date)


ghajini:
`telefullenvivo.com` still accessible

FanOfRootBeer:
> `telefullenvivo.com`

oops, you're right, confused it with `televisionlibre.net`

stephenhawk8054:
`skidrowcodexreloaded.com` is still accessible, but looks like the filter is not needed any more.

PR #19696, opened 2023-09-11
Add TP-Link domains used for LAN configuration

gacelperfinian:
This adds five additional domains that is used by TP-Link for end-user configuration of their products. Four of them (tplinkap.net, tplinkmodem.net, tplinkplclogin.net and tplinkrepeater.net) have a self-demonstrating fallback webpage that is shown if not connected to their products.

The fifth one (tplinkeap.net) does not have a self-demonstrating fallback webpage but it is listed in their guide at https://static.tp-link.com/configurationguide/quick_start_guide_standalone_eap.pdf#page=5 with the likely explanation being that this domain is intended for enterprise users where administratiors would know how to access the equipment using IP addresses if this domain somehow won't work.


gwarser (approved):

stephenhawk8054:
@gwarser I guess we can merge?

PR #20135, opened 2023-10-18
fix: remove github copilot promotion button

anyone00:

https://github.com/uBlockOrigin/uAssets/blob/master/README.md or any other github file.

### Describe the issue

This removes the "Your organization can pay for GitHub Copilot" or "Code 55% faster with GitHub Copilot" promotion buttons when viewing a file in github web.

### Screenshot(s)

image

image

### Versions

- Browser/version: Firefox 118.0.2 / Chromium 118
- uBlock Origin version: 1.52.2

### Settings

defaults + 
- uBlock filters – Annoyances
- uBlock filters – Cookie Notices

### Notes

This is not always shown, especially if you've clicked on the "Don't show again" button before. I see these while I'm signed in with accounts I've not pressed the "Don't show again" button.

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -6920,3 +6920,7 @@ raenonx.cc##+js(nostif, show)
 
 ! https://deccanherald.com/india/delhi/people-detained-from-jantar-mantar-for-denouncing-israel-over-gaza-action-2728594 - Copy manipulation
 deccanherald.com##+js(aopr, document.oncopy)
+
+
+! GitHub copilot promotion when viewing files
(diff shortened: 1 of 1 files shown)


3kh0:
As a github user I support this

stephenhawk8054:
For the filters that are compatible with EasyList, please report to here: https://github.com/easylist/easylist/pulls

anyone00:
There is no "annoyance" easylist list afaik, and I don't like to use fanboy's annoyance lists in uBO. so if possible I would like to have it here :)

mapx-:
@ryanbr

PR #20138, opened 2023-10-18
A: haveibeenpwned.com

ajayyy:
https://haveibeenpwned.com/

Before:

image

After:

image

## Context

I originally added filterlist for the 1 Password ads in https://github.com/easylist/easylist/pull/10909. I was able to get most of it blocked, but not this last sentence.


ajayyy:
Stay is actually not required as it all exists in the html at the start

PR #20196, opened 2023-10-21
A: ocremix.org

MisterSheeple:

`https://ocremix.org/remix/OCR03969` for example

### Describe the issue

Before:

image

After:

image

### DIFF
--- a/filters/filters-2023.txt
+++ b/filters/filters-2023.txt
@@ -5055,3 +5055,7 @@ pmkisanlists.in,rojgargovernment.in,nregajobcard.in,cookar.net,cookoz.net,shrami
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/17c2b2e/ublock_detected/
 porninblack.com##+js(rmnt, script, deblocker)
+
+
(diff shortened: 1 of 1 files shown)


MisterSheeple:
I'm new to this, so if there's a better way this can be done, please let me know. I closed this because I was thinking of blocking the widget by checking for a "Promotion" string in there, so in case the widget were to ever disappear, it wouldn't cause breakage. But I can't figure out how to do that, so let me know if my current implementation is fine or if I can improve upon it, and if so, how.

BlazeFTL:
@MisterSheeple Its Always Better To Pinpoint Your Target Element
Ex:
`##div.widget-neutral.widget:has-text(Promotion)`
nth-of-child Is Not Ideal And Can Cause Breakage In Future
Take A Look Here `https://github.com/gorhill/uBlock/wiki/Procedural-cosmetic-filters`

MisterSheeple:
Thank you, should be better now.

mapx-:
report at easylist repo https://github.com/easylist/easylist/issues

PR #20247, opened 2023-10-23
Fix https://github.com/hagezi/dns-blocklists/issues/1744

iam-py-test:

```
magisk.info
magiskzip.net
magiskmanagerroot.com
```

### Describe the issue

Fake websites for https://github.com/topjohnwu/Magisk (https://github.com/topjohnwu/Magisk/issues/3435)
Thanks

### Versions
N/A

### Settings

N/A

### Notes

The first two websites are from https://github.com/hagezi/dns-blocklists/issues/1744
I added them and one other fake site I found as `$all` rules, plus a few broader rules which should cover all the known fake sites + any new ones. I am a bit worried they might cause false positives (ie fan sites, or the very unlikely chance something else is named the exact same thing), so `$doc` only. Kept the originals but can remove them if needed.


iam-py-test:
Thanks

PR #20387, opened 2023-10-31
Xpshort Up

BlazeFTL:
`https://xpshort.com/nOj2o`
@mapx- Is It Okay If I Move Those Rule From Line 4500 To Line 1 Or Start ?
Basically Any Site That Changes Domains Regularly ?


mapx-:
You mean positioning all those filters as the first filters in `filters-2023` list ?

PR #20652, opened 2023-11-13
DropGalaxy Up

BlazeFTL:
`https://dropgalaxy.vip/1067702uyv4s`
Hopefully This Works For Most People Out There
If The Previous Solution Seems Better Than Then Its Better To Reject This One...I Hope This Would Decrease The Chance Of Anti Adb On FireFox

### DIFF
--- a/filters/filters-2020.txt
+++ b/filters/filters-2020.txt
@@ -224,13 +224,10 @@ desktophut.com###loading-ad
 @@||challenges.cloudflare.com^$subdocument,domain=dropgalaxy.*
 @@||login.popmonetizer.com/ss/js/ads.js$script,domain=dropgalaxy.*
 ||assets.isavetube.com/images/dgpromo.jpg
-||youtube.com^$domain=dropgalaxy.*
 dropgalaxy.*##+js(acis, document.createElement, /break;case \$\.|popunder/)
 dropgalaxy.*##+js(trusted-set, document.visibilityState, visible)
-dropgalaxy.*##+js(no-fetch-if, /adtrue\.com|eningspon\.com|freychang\.fun|orquideassp\.com|popunder/)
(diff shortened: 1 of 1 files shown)


JobcenterTycoon:
I can confirm it works for Firefox but

```
dropgalaxy.*##+js(rmnt, script, /.onerror|document.write|async|Error|sbmt|isMobileBrowser|navigator.userAgent|InstallTrigger|userTimezoneOffset|badip/)
dropgalaxy.*##+js(rpnt, script, /.onerror|.onload|.offsetParent/)
```

breaking the cloudflare ddos protection screen. Also i don’t understand

`||youtube.com^$domain=dropgalaxy.*`

You can merge the `no-xhr-if` and `no-fetch-if` filter.

BlazeFTL:
@JobcenterTycoon I Tried Many Times But I Dont Get Any Cloudfare Screens Now At All...With Diff IP,Clear Everything..So Cant Test

YouTube Was Just Useless..For Clicking 3 Buttons No One Needs A Tutorial Also Reduced Bandwidth Usage..Can Be Removed If You Want

JobcenterTycoon:
Try a crappy free proxy browser addon, it will trigger cloudflare 99%.

BlazeFTL:
@JobcenterTycoon Can You Check Final Edit ?
I Still Think `youtube.com^$domain=dropgalaxy.* `
Bandwidth Waste🐸

JobcenterTycoon:
`badip` crashing the cloudflare check. I prefer more safe filter like

`dropgalaxy.*##+js(rmnt, script, /div-gpt-ad-dropgalaxycom|script1\.async|InstallTrigger|userTimezoneOffset=userDate\[/)`

To make rpnt more safe you can use [condition](https://github.com/gorhill/uBlock/wiki/Resources-Library#replace-node-textjs-) so the rpnt filter will only be applied when specific string or regex being found.

BlazeFTL:
I Had No Problem With Badip Tho...Anyway Removed...My Regex Capabilities Are Weak AF So I Dont Wanna Do Anything More🥱

PR #20768, opened 2023-11-19
Exclude MusicBrainz localhost 3P from LAN block filter

nicholastay:

MusicBrainz pages where Picard is being used, e.g. `https://musicbrainz.org/search/textsearch?limit=50&type=recording&query=iyowa&tport=8000` -> click 'TAGGER'

(Once visited with tport, it will remember the port and use it across the site, e.g. release info, etc)

### Describe the issue

On using MusicBrainz with Picard, to use the Tagger feature, clicking the button issues a request to `127.0.0.1:<tport>` in order to relay the result clicked to the local client. LAN block filter breaks this, so we add an exclusion here for it.

### DIFF
--- a/filters/lan-block.txt
+++ b/filters/lan-block.txt
@@ -102,4 +102,7 @@
 ! https://github.com/uBlockOrigin/uAssets/issues/19005
 @@||127.0.0.1^$xhr,domain=battlelog.battlefield.com
 !
+! https://github.com/uBlockOrigin/uAssets/pull/20768
+@@||127.0.0.1^$3p,domain=musicbrainz.org
+!
 ! ——— END


mapx-:
@gwarser

gwarser:
If you want to make holes everywhere, maybe don't use lan-block? 

There are already exceptions in the list, so if it looks sane, merge it.

I cannot look at it (literally - pink eyes).

nicholastay:
For reference, I’m happy to just have this as a personal rule on my side, feel free to close if it doesn’t feel appropriate to punch holes upstream. Just noticed there was an exception for Battlefield already there so thought I would raise a PR in case it helps anyone.

PR #20779, opened 2023-11-20
address pysznosci.pl (WP group) https://github.com/uBlockOrigin/uAssets/issues/12214#issuecomment-1817010374

garry-ut99:
@JobcenterTycoon


JobcenterTycoon:
Oh sorry i forgot about this issue.

The first filter can be merged with a existing filter to `wiadomosci.wp.pl,kobieta.wp.pl,pysznosci.pl##div > div:first-child > iframe[src^="https://v.wpimg.pl/"][frameborder="0"][scrolling="no"]:first-child:upward(2)`

For the second issue try `pysznosci.pl###bottom-sticky-container > div:last-child:matches-attr(class=/[a-zA-Z]{2}/):not([id])` because the elements are unstable and i can’t reproduce the timer redirect.

garry-ut99:
> The first issue can be fixed with a existing filter already. So it can be merged to 

Right, I just forgot to check the already existing filters.

> For the second issue try, because the elements are unstable

A class length & last-child handle is not guaranted to be stable either, the length didn't change do far, so maybe it's good choice at the moment, but `last-child` works only if GDPR is accepted in usual way, but doesn't work if an user accepts GDPR message a different way than usual, to reproduce try steps in this comment: https://github.com/uBlockOrigin/uAssets/issues/12214#issuecomment-1722520173, once `last-child` is removed from the filter, it works again.

Issue #20799, opened 2023-11-21
graphicuv.com: detection


americanape:
### Prerequisites

- [X] This is **NOT** a [YouTube](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [Facebook](https://www.github.com/uBlockOrigin/uAssets/issues/3367) or [Twitch](https://www.github.com/uBlockOrigin/uAssets/issues/5184) report.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I forced an update of my filter lists. (Click the ["Purge all caches"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches) button while holding the 'Shift' key, then click the ["Update now"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now) button.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [X] I did not answer truthfully to **ALL** the above checkpoints.

### URL address of the web page

`https://graphicuv.com/what-is-the-importance-of-branding-and-signature-style-for-company`

### Category

detection

### Description

Detecting ad block and element zapper/picker mode not working

### Other extensions used

no other extension

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.53.0
Chromium: 119
filterset (summary):
 network: 103496
 cosmetic: 43783
 scriptlet: 19066
 html: 0
listset (total-discarded, last-updated):
 default:
[...]

PR #20803, opened 2023-11-21
Fix Multi Shorteners 

BlazeFTL:
`https://lksfy.com/jlmfm0olam`
`https://go.birdurls.com/ntqy9qolam`
`vlshort.com/3k7hc9olam`
`https://cut.lc/w6c8qpolam`
https://github.com/uBlockOrigin/uAssets/issues/20799

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -5418,11 +5418,6 @@ fembed-hd.com##+js(aopr, __Y)
 ! https://github.com/uBlockOrigin/uAssets/issues/10779
 mconverter.eu##+js(nostif, offsetHeight)
 
-! https://github.com/uBlockOrigin/uAssets/issues/10777
-sevenjournals.com###yuidea-generate:style(display: block !important;)
-sevenjournals.com###yuidea-snp:style(display: block !important;)
-sevenjournals.com###yuidea-wait1
-
 ! https://github.com/uBlockOrigin/uAssets/issues/10793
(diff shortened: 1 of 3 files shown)


BlazeFTL:
It Would Be Cat And Mouse Game...This Site Owner More Active Than DG😂

JobcenterTycoon:
On the next update we can try `spoof-css`

BlazeFTL:
Currently spoof-css only returns the value that we set but if it could return values directly from css then it might perform better...
##a[href*=".click"] Was Already Updated With Different Styles,Spoof But All Got Bypassed By Site Owner...I Dont Know What MasterKia's rpnt Filter Was Maybe You Can Try On This Site In Future..

Cause Here `##a[href*=".click"]` = DropGalaxy's `iframe` Style Filters...I Had To Update Whenever Site Owner Updated

JobcenterTycoon:
MasterKia’s rpnt filter obtain the default real css and returning it to the site. With this MasterKia was able to bypass dragontea’s `spoof-css` detection. dragontea changed the style and then detect if the obtained `getComputedStyle` style also changed. This way to detect spoof-css is not uncommon.

See the full discussion: https://github.com/uBlockOrigin/uBlock-issues/issues/2618

BlazeFTL:
@JobcenterTycoon Recheck...Site Owner Updated🤣

JobcenterTycoon:
Done. Also blocked the linked badware domain and reported to different security vendors.

Issue #1775, opened 2018-03-23
lequipe.fr [Anti-Adblock]


KiLL4FuNK:
URL(s) where the issue occurs

`https://www.lequipe.fr/Cyclisme-sur-route/Actualites/Niki-terpstra-gagne-a-harelbeke-et-recompense-la-formation-quick-step-floors/886664`

Describe the issue

Anti-Adblock

Versions

Browser/version: Chrome 62.0.3202.62
uBlock Origin version: 1.15.18

Settings

Default

KiLL4FuNK:
Anti-Adblock @mapx-

mapx-:
try
```
lequipe.fr##.ab-full-wrapper
lequipe.fr##[rel="stylesheet"][type="text/css"] + * + DIV > DIV:first-child
```

KiLL4FuNK:
@mapx- It works perfect

mapx-:
the second filter could be 
`lequipe.fr##STYLE + * > :first-child:matches-css(z-index: 1)`

garry-ut99:
(no AA for me with `ghide` off) Many big ad placeholders caused by `ghide` from "uBlock filters": [SCREENSHOOT](https://user-images.githubusercontent.com/72945564/217843540-22139186-6e9e-4468-a35c-65ab91dc596f.png)

PR #20847, opened 2023-11-24
update #1775

ItsProfessional:

`lequipe.fr`

### Describe the issue

fix anti-adb

### Notes

anti-adb detection code (checks if fake `.ad-box` was hidden by uBO):
```js
window.$nuxt && !n && (n = (o = document.querySelector(".ad-box")) && ("none" === (null === (t = window.getComputedStyle(o)) || void 0 === t ? void 0 : t.display) || 0 === (null == o ? void 0 : o.offsetHeight)), null === (e = $nuxt.$store) || void 0 === e || e.dispatch("setIsAdsBlocked", n));
```


mapx-:
the filter is already neutralized in the uBO lists

https://github.com/uBlockOrigin/uAssets/blob/master/filters/filters-2023.txt#L3639

mapx-:
You provided a wrong page link.
The detection is active only when you open an article.

PR #20915, opened 2023-11-27
infinityscans.xyz Up

BlazeFTL:
`infinityscans.xyz`

### DIFF
--- a/filters/quick-fixes.txt
+++ b/filters/quick-fixes.txt
@@ -158,7 +158,8 @@ check-host.net##div[id]:matches-css(position: relative):has(a img:matches-css(wi
 !#if env_chromium
 @@*$script,xhr,domain=infinityscans.xyz
 !#endif
-*$frame,redirect-rule=noopframe,domain=infinityscans.xyz@@||infinityscans.xyz^$ghide
+*$frame,redirect-rule=noopframe,domain=infinityscans.xyz
+@@||infinityscans.xyz^$ghide
 @@||googlesyndication.com^$xhr,domain=infinityscans.xyz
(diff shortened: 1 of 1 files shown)


mapx-:
in chromium it's difficult with the race condition

BlazeFTL:
> such exceptions

On My Side FF,Chrome Both Works With script,xhr

mapx-:
maybe now we can remove 
`@@*$script,xhr,domain=infinityscans.xyz`
?

BlazeFTL:
@mapx- I Never Needed It

BlazeFTL:
> firefox android where html filter not worked

Mention Gorhill

mapx-:
I get detection in android

BlazeFTL:
Now For FF This Is Needed.. Chrome Works Bcz Of Script,xhr Allowed
```
@@||cdn.prplads.com^$script,domain=infinityscans.xyz

BlazeFTL:
I Say Let's Allow Script,xhr For FF Also...Let The Site Owner Waste His Time By Adding New Scripts,Xhr Checks

mapx-:
yes, I added `@@*$script,xhr,domain=infinityscans.xyz` in android and is working

However, we could add it only for mobile

BlazeFTL:
> only for mobile

There's No Saying That It Would Always Work For Mobile...

Allow For Everything

BlazeFTL:
> nothing left to block ads

No Ads Can Get Through...But The Same Goes For Every Problem... Defusing At Root Level For Less Exceptions Would Take Time...If Anyone Interested Go Ahead

PR #20959, opened 2023-11-29
Up Devup

BlazeFTL:
`http://devuploads.com/gp74hvcgow1q`


okiehsch:
On my end using chromium 
````
dev.miuiflash.com,djxmaza.in,thecubexguide.com##+js(rpnt, script, /none'|hidden'|undefined/)
````
breaks the "checking browser" button. If I disable that filter I can download without issue.

BlazeFTL:
Done

PR #20965, opened 2023-11-30
DG Up

BlazeFTL:
`https://dropgalaxy.com/drive/xx34excmo7jd`

Thinking About Dropping This Site...I Dont Think People Use It🤣
Tired With Site Owner Fights😆


BlazeFTL:
Hmm Typo `infinityscans.xyz##+js(no-fetch-if, platform.bidgear.com, true)`
Was Meant For extbuild
Fetch true isnt available in normal builds

PR #21004, opened 2023-12-01
Devup

BlazeFTL:
`https://devuploads.com/25wi943etnl3`


Omega-6:
not to be nagging, but it looks like there are google ads, could we find a more specific approach?

Omega-6:
yes, I know it's about anti-adblock

PR #21034, opened 2023-12-03
Fix https://github.com/easylist/easylist/issues/17915

Omega-6:
`https://www.thespruce.com/grow-bamboo-inside-1902611`

article title gets crammed


Omega-6:
yup, didn't even observe it since it was obsolete (or ejected via the temporary filter disabler in the logger)

PR #21039, opened 2023-12-03
Update filters.txt

Omega-6:
`https://igg-games.com/the-zone-stalker-stories-free-download.html`
ad under metadata (title, timestamp etc.), ad in right sidebar (last visible item there)


Omega-6:
@ItsProfessional I purposefully did not give a specific name to the pr since the admin is circumventing ad blockers

PR #21058, opened 2023-12-04
Fix https://github.com/easylist/easylist/issues/17937

Omega-6:
`https://edition.cnn.com/`

easylist cookie detection

use a EU IP address


peace2000:
Added this to the quick fixes list: https://github.com/uBlockOrigin/uAssets/commit/29ccd48dbed9b603547e00ec0d42ea856e8f6f81

PR #21059, opened 2023-12-04
Filters for alliance4creativity.com

Omega-6:
e.g.:

`zoro.to` => redirects to `https://www.alliance4creativity.com/watch-it-legally/?utm_source=zoro.to&utm_medium=Domain&utm_campaign=Redirect`

`$3p` to address frames

can we add:

```
||zoro.to^
||2embed.to^
```
(without `$all` for Brave compatibility)

?


JobcenterTycoon:
There is no reason to block this domain. The URL parameter are covered by `AdGuard URL Tracking Protection`

Omega-6:
@JobcenterTycoon not even the possibility of IP addresses getting logged when the user has no intention of visiting the site?

PR #21080, opened 2023-12-05
Address https://github.com/AdguardTeam/AdguardFilters/issues/167559

Omega-6:
`https://www.99corporates.com/`

detection


Omega-6:
@ItsProfessional have you tested in a chromium-based browser?

PR #21173, opened 2023-12-07
Parked domain (ytix.xyz)

Omega-6:


JobcenterTycoon:
Thanks for contributing.

One note: You can also remove multiply domains in one pull request.

Omega-6:
Sure, but I think mapx told me when I was using an account (probably on road-block) to open one / site

JobcenterTycoon:
When you only remove parked/obsolete filter its ok to process multiply domains with one pull request.

PR #21269, opened 2023-12-08
address "Fineshop Design" anti-adb sites

huyszn:
```
answerpython.com
gsm-solution.com
h-donghua.com
hindisubbedacademy.com
mydverse.com
mydverse.xyz
pkgovjobz.com
ripexbooster.xyz
serial4.com
serial412.blogspot.com
tutorgaming.com
```


Omega-6:
some maintainers probably prefer breaking the anti-adblock rather than dealing with issues caused by using multiple ad blockers

PR #21422, opened 2023-12-15
address `sportsupa.com` annoyance

huyszn:
`https://sportsupa.com/total/sports/2023/nfl-bb10/`

Anti right click


huyszn:
The anti right click occurs in stream links only listed on `https://sportsupa.com/schedule.php/` like `https://sportsupa.com/total/sports/2023/nfl-bb1/`. The stream pages have `<body oncontextmenu="return false">` in them. I tried using `sportsupa.com##body[oncontextmenu="return false"]:remove-attr(oncontextmenu)` but it only worked on firefox.

PR #21459, opened 2023-12-17
Privacy: add Private State Token permission policies

bershanskiy:

Launch Chromium with a test "private state token commitment"
```bash
chromium --args --additional-private-state-token-key-commitments='{ "https://private-state-token-issuer.glitch.me": { "PrivateStateTokenV1VOPRF": { "protocol_version": "PrivateStateTokenV1VOPRF", "id": 1, "batchsize": 1, "keys": { "1": { "Y": "AAAAAQQ7W5gOubJT3kTpzNGsekT9RZPXgXGrOMB2+QPw/ZzAuLrM3kc8eyHuTc1KmKjH4sh5+ev5GCI4HVVd46o6rWvNvk0iZQtVuUPhT8X54Ajebng8v5zUnpnPuTjGqlc7+MM=", "expiry": "1715356984440000" } } } } }'
```

Go to `https://private-state-token-issuer.glitch.me`

Click on "I'm a human" button and follow prompts. While doing it, in DevTools observe changes to Application > Private state tokens table and extra network requests with token data going back and forth.

### Describe the issue

Google Chrome (Chromium) already shipped "Private State Tokens" ([Google Docs](https://developers.google.com/privacy-sandbox/protections/private-state-tokens), [spec](https://wicg.github.io/trust-token-api/)). The proposal is somewhat related to IETF Privacy Pass protocol, but it is a distinct protocol developed by ["the improving web advertising business group"](https://www.w3.org/community/web-adv/) and is apparently focused specifically on advertisement reliability. The proposal was called Trust Tokens API before re-branding and shipping. This API extends Fetch API with new fields which allow websites to store tokens on users' devices (in "issuance" process) for later use by other websites (via "redemption" process). The API availability does not depend on user choice to block cookies or other privacy settings, does not have any user-visible prompts, and is designed to be completely invisible ("friction-less") to the user.

As of 2023, there does not appear to be any publicly available evidence that the API improves user experience at all or even that anyone ever attempted to use it to improve user experience by, e.g., reducing number of [CAPTCHA](https://uk.wikipedia.org/wiki/CAPTCHA)s displayed to the user (like [Apple did with Private Access Tokens](https://developer.apple.com/news/?id=huqjyh7k)).

### Screenshot(s)

### Versions

- Browser/version: Chromium Version 120.0.6099.71
- uBlock Origin version: 1.54.0

### Settings

None

### Notes


Westfargo0147 (approved):

Westfargo0147 (approved):

bershanskiy:
Mozilla updated their [position on "Private State Token API"](https://mozilla.github.io/standards-positions/#private-state-token) to "negative":
> Private State Tokens provides sites with the means to exchange information about visitors, using Privacy Pass to ensure that there are very tight bounds on the rate of information transfer. We conclude that the usage constraints in the design are insufficient to effectively safeguard privacy.

PR #21594, opened 2023-12-27
update gplinks

huyszn:
`disking.in`


huyszn:
@ItsProfessional I don't know when gplinks no longer uses domains but for the new ones I know when I use a link from gplinks and it redirects me to the new domains.

PR #21613, opened 2023-12-28
address `dollareuro.live` anti-adb

huyszn:
`https://dollareuro.live/`

There are also annoying page transition animations whenever you click on a link or wait for a page to load on the site but I don't know how to remove them.


D4niloMR:
you can remove the animation with

`||dollareuro.live/wp-content/plugins/elementor-pro/assets/js/page-transitions.min.js$script,1p`

Issue #7268, opened 2020-04-19
notube.net


JobcenterTycoon:
### URL(s) where the issue occurs

`https://notube.net`

### Describe the issue

Popup when you try to download a video

### Versions

- Browser/version: Firefox 75
- uBlock Origin version: v1.26.0

PR #21620, opened 2023-12-28
update #7268

huyszn:
Added domains:
```
notube.fi
notube.im
notube.lol
```


huyszn:
The reason I didn't do `notube.*` is because there is a domain called `notube.com` that is not related to the other domains.

krystian3w:
Maybe possible simplified to:
```adb
notube.*,~com##+js(ra, onClick)
notube.*,~com##+js(nowoif)
```

I doubt to create mirror places like:
`notube1.com` to undo simplification.

PR #21626, opened 2023-12-28
address `1primewire.com` ad placeholder

huyszn:
`https://ww2.1primewire.com/`
Ad placeholder at bottom right of site

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -358,8 +358,7 @@ forum.gigabyte.us###remove_ads_link
 
 ! 1primewire .com popups ads
 ||1primewire.com/addons/$image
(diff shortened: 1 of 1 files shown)


JobcenterTycoon:
Firefox supports native `:has()` now so `1primewire.com##[id]:has(> [class="close ico"])` should be faster.

The "BUY PREMIUM" button is not a case for the filterlist so `1primewire.com##[href="/user/profile/premium-membership"]` is not needed.

huyszn:
I saw other filters that had `##[href="/user/profile/premium-membership"]` so that was why I added it.

JobcenterTycoon:
Ok i asked the other contributer about `##[href="/user/profile/premium-membership"]` because i just see the button

PR #21644, opened 2023-12-28
address `sethphat.com` timer

huyszn:
`https://sethphat.com/download/?id=5497` from `https://toku.fun/post/kamen-rider-gotchard-episode-16-full-english-sub`


D4niloMR:
add `! https://sethphat.com/download/?id=5497` instead of `! sethphat .com timer` 

it's easier to check if the filter is still relevant later.

PR #21672, opened 2023-12-30
address `instantearn.in` shortener

huyszn:
`instantearn.in/lnwOySrW`


huyszn:
I tested `petrainer.in` with and without `##+js(nano-sib, timer, *, 0.001)` and it did not matter.

PR #21693, opened 2023-12-31
address `dfast.app` ad overlay and timer

huyszn:
`https://www.dfast.app/instagram/com-instagram-android/download.html`


huyszn:
It causes the ad overlay, not a detection script

PR #21967, opened 2024-01-14
Multi UP

BlazeFTL:


ItsProfessional:
You have been previously told by others to separate the cosmetic filters to different lines instead of comma concatenation, why do you still use commas?

BlazeFTL:
Old Habit Would Fix In Future : (

PR #21982, opened 2024-01-15
update (NSFW) `porninblack.com` anti-adb

huyszn:
`https://porninblack.com/`

I think this site also has popMagic popups because if you click on a video on the main page, it opens a new tab.


ItsProfessional:
why not add the chp filter in the chp section in 2023.txt?

ItsProfessional:
https://github.com/uBlockOrigin/uAssets/commit/3c35879b82df755d86b9f8b84f822da8a6111ad7

huyszn:
I saw that there was a section for `porninblack.com` filters and the filter for the popups would also be under it.

PR #22047, opened 2024-01-18
update `tii.la` shortener

huyszn:
`https://tii.la/qczmBQKd`

Make only "Get Link Here" button display. Also, the shortener sometimes goes to `host2loan.com` and not `blogtechh.com`


D4niloMR:
`:others()` is a procedural filter, should be used if the page is really bloated and not easy to deal with normal filters. 

This shortener doesn't really need cosmetic hiding, as you click in the "Continue" it goes to the "Get link here" without needing to scroll.

PR #22049, opened 2024-01-18
move `oko.sh` to shorteners

huyszn:
`https://oko.sh/ciz72yolam`


huyszn:
Is the link shorteners list also for shorteners that use different domains frequently?

Issue #20095, opened 2023-10-15
[nsfw] truyen-hentai.com: ads


adbl0cking:
### Prerequisites

- [X] This is not a support issue or a question. For support, questions, or help, visit [/r/uBlockOrigin](https://www.reddit.com/r/uBlockOrigin/).
- [X] I performed a [cursory search of the issue tracker](https://github.com/uBlockOrigin/uAssets/issues?q=is%3Aissue) to avoid opening a duplicate issue.
- [X] I am running the latest version of uBlock Origin (uBO).

### I tried to reproduce the issue when...

- [X] uBO is the only extension.
- [X] uBO uses default lists and settings.
- [X] using a new, unmodified browser profile.

### Description

exoclick `Neverblock`

### URL(s) where the issue occurs.

`https://www.truyen-hentai.com/` [redirects to local version]

### uBO version

uBO 1.52.2

### Browser name and version

chrome latest

### Settings

.

### Notes

```
||truyen-hentai.com/nb/
truyen-hentai.com##.async-reklam-placeholder
truyen-hentai.com##+js(rmnt, script, Exo)
```

PR #22109, opened 2024-01-20
upd #20095

ABetterWeb:
nsfw

detection => overlay


ABetterWeb:
also, it looks like the `rmnt` scriptlet for this site may be outdated (anyway no ad)

stephenhawk8054:
It still displays the overlay if I refresh without bypassing the cache. Does this work?

```adb
truyen-hentai.com###aabnode
```

ABetterWeb:
@stephenhawk8054 I don't get that on Firefox, but your filter works. How about still addressing `window.aab` though?

stephenhawk8054:
Ok I'll update it

PR #22115, opened 2024-01-21
address `teachmemicro.com` anti-adb

huyszn:
`https://www.teachmemicro.com/`


ItsProfessional:
By the way, when making PRs to add filters for new sites, add the link to your PR in the comment instead of `[domain] anti-adb`

PR #22266, opened 2024-01-30
Hide PornHub ad block notice

1366613:


mapx-:
already fixed in uBO annoyances by

`pornhub.*###js-abContainterMain`

PR #22271, opened 2024-01-30
add egun.de

skyracer2012:

`https://www.egun.de/market/index.php`

### Describe the issue

There is a banner advertisement which will now be removed.


mapx-:
already fixed in the german regional list, enable it, use uBO with default settings.

PR #22303, opened 2024-02-02
tgx up

BlazeFTL:


ItsProfessional:
Why did you move it to the top?

> New filter must be added at the end of the list.
The reason is to provide an easy way to check whether a filter is still relevant. The filters at the top of the file will be the oldest filters, and also the most likely to maybe be obsolete.
Old filters which are confirmed to still be required must be moved to the **end** of the list.

BlazeFTL:
@ItsProfessional It Isnt Any Issue...I Have Been Here Longer Than You So I Know What I Am Doing

BlazeFTL:
@huyszn `[href^="/hus.php"]`

BlazeFTL:
@daxy7 Tell Me Why You Laughing?
What Would Happen So Catastrophic If Those Few Lines Go At Top ?

ItsProfessional:
It's a policy that's been here forever, so it's better if its consistently followed.
I don't see any valid reason for them to be moved to the top.

PR #22596, opened 2024-02-21
https://github.com/uBlockOrigin/uAssets/issues/22586

ghajini:
https://github.com/uBlockOrigin/uAssets/issues/22586


stephenhawk8054:
We need breakage reports any way. Otherwise, there's no way to confirm. Based on the descriptions of the cookies, I think it would be fine.

PR #22787, opened 2024-03-05
Address revancedapps[.]com and other similar websites

iam-py-test:

`https://revancedapps.com/` (malicious)
`revancedapps.example` ([safe way to test rule coverage](https://en.wikipedia.org/wiki/.example))

### Describe the issue

While revancedapp.[TLD] is covered by existing rules, revancedapp**s**.[TLD] is not.
[As this pattern is being used in the wild](https://github.com/hagezi/dns-blocklists/issues/2288), I thought it pertinent to add a rule for it.
Thank you

### Versions

- Browser/version: Firefox 123.0 (64-bit)
- uBlock Origin version: 1.56.1b11

### Settings

Irrelevant, issue can be confirmed by rule examination.

### Notes

Full credit to @damianir750 for finding the original malicious website.


the-1Riddle (commented):
👀

PR #22811, opened 2024-03-07

Miipan:
Update verify button missing and block all element and text, so don't have to bother scrolling the screen

image


ItsProfessional:
> block all element and text, so don't have to bother scrolling the screen

Do this in "My filters"

Issue #22885, opened 2024-03-12
fruugo.de: nuisance


juarez-spa3:
### Prerequisites

- [X] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://github.com/uBlockOrigin/uAssets/discussions/17361) report. _These sites MUST be reported by clicking their **respective links**_.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [ ] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin along with other content blockers.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [X] I have verified that other extensions are not causing the issue.
- [X] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://www.fruugo.de/8tb-ssd-usb-c-usb-31-externe-solid-state-drive-ssd-betrouwbare-opslag-voor-gaming-studenten-professionals-8tb-zwart-40-korting/p-276197962-613988471?language=nl`

### Category

nuisance

### Description

cookie consents

### Other extensions used

none

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

![resim](https://github.com/uBlockOrigin/uAssets/assets/153795294/4551c99d-de6a-49b3-9c0d-696a5fc08528)

[...]

PR #22904, opened 2024-03-14
fix https://github.com/uBlockOrigin/uAssets/issues/22885

ghajini:
all under given in https://github.com/uBlockOrigin/uAssets/issues/22885


stephenhawk8054:
Looks like it works for me. But I can't test other further activities (buy, check-out...)

PR #22944, opened 2024-03-16
Add McGraw Hill copy-paste prevention

ericswpark:

`https://prod.reader-ui.prod.mheducation.com/epub/epub-id`

### Describe the issue

Attempts to highlight text and copy on a McGraw Hill e-textbook results in a popup. The first rule disables the popup completely. However, the site still prevents copy-pasting. The second and third rules remove the `oncut` and `oncopy` attributes from all elements, which are set to `return false` in a desperate attempt to sabotage students from actually using the material they paid for.

### Screenshot(s)

image

### Versions

- Browser/version: Firefox 123.0.1
- uBlock Origin version: 1.56.0

### Notes

I've tried removing the first rule in order to preserve the popup but also get copy-paste working. However, when the popup is dismissed, the text is unselected automatically, preventing copying.

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -7456,3 +7456,8 @@ helldivers.io###gwvideo
 ! https://bongdaplus.vn/ - anti right click, select
 bongdaplus.vn##+js(aeld, mousedown, undefined, elements, body)
 bongdaplus.vn##+js(set, document.oncontextmenu, trueFunc)
+
+! McGraw Hill e-textbooks - copy paste
+prod.reader-ui.prod.mheducation.com##+js(aeld, mouseup)
+prod.reader-ui.prod.mheducation.com##*:remove-attr(oncopy)
+prod.reader-ui.prod.mheducation.com##*:remove-attr(oncut)


ItsProfessional:
Disabling the whole popup with filters can be considered as breakage of the site.

> I've tried removing the first rule in order to preserve the popup but also get copy-paste working. However, when the popup is dismissed, the text is unselected automatically, preventing copying.

Add a filter to defuse the code unselects the text; if you can't, tell us how we can reproduce the issue. Does the site require an account?

ericswpark:
@ghajini you need a McGraw Hill account and an eBook purchased on that account, which does make reproducing it a bit annoying.

@ItsProfessional yes, I think some might find the highlighting function of the popup useful. I'll see if I can do that later. The site does require an account and it also requires purchasing an ebook license, unfortunately. :( If you do manage to get a MGH account, it should be reproducible in the main eBook interface, simply by highlighting any text portion of the eBook.

JobcenterTycoon:
So you select text, then the popup opens which also cover the whole text and when you close the popup the selection disappears?

Maybe you can craft a CSS rule which removes the backdrop but not the toolbar.

stephenhawk8054:
Since it's just an annoyance and we cannot investigate without the accounts, which would cause a lot troubles if some breakages happen in the future, I think it's better to use as custom filters.

ericswpark:
OK, so I had some time to test out exactly what they're doing. When I select the text, they show an overlay with a dark-gray background and the modal showing up on top of the text.

I tried the CSS rule trick that @JobcenterTycoon mentioned, and while it does get rid of the background, I think it can be construed as site breakage as it makes it hard to distinguish the popup from the text.

But the bigger issue here is that the reason the popup prevents copy-paste is because it steals focus from the text, and the user can only tab back and forth the items within the popup. When the popup exits, either via the Esc key, clicking on the dismiss button, or clicking on the dark transparent background, then the text is deselected. I've tried tracing where this happens in the JS using the debugger, but the JS is minified and I honestly can't figure out how it is deselecting the text.

Please feel free to close this -- I am satisfied with the workaround and don't care very much about the popup "functionality" being broken. I can live with the custom filters until the semester ends.

JobcenterTycoon:
Try `pointer-events: none` this will defuse all click events on the backdrop and the user can select and copy the text through the backdrop without hiding.

PR #23028, opened 2024-03-23

BlazeFTL:
`urls.cx/DVxgfufi?src=PSA` = Site Owner Doesn't Change Site Structure So others() Is Used
`fc-lc.xyz/zSCLOlIx` = Old Domains No Longer Used
`https://modijiurl.com/Niksindia` = Structure Changes So Only Ads Blocked Instead Of Others()


BlazeFTL:
A update is needed for fc-lc
`tmail.io##[id^="ad"]`

PR #23278, opened 2024-04-12
upd theporngod. com

ToIgnore:
nsfw `https://theporngod.com/`
popup when clicking on a video


JobcenterTycoon:
Better blocking the popMagic script directly instead of using a broad `nowoif` scriptlet.

PR #23284, opened 2024-04-13
DG Up

BlazeFTL:
`https://dropgalaxy.vip/hffg86dqmbjg`

I Think  These Can Be Removed If Anyone Care To Test
```
||adoto.net^$domain=dropgalaxy.*|financemonk.net,important
||rsc.cdn77.org^$domain=dropgalaxy.*|financemonk.net,important


JobcenterTycoon:
Works thanks.

PR #23353, opened 2024-04-18
Fix "importgant" typo in a filter

antonok-edm:
`display: block !importgant` is invalid CSS, so this filter has no effect at the moment.


mapx-:
thx

Issue #21501, opened 2023-12-20
baeldung.com: detection


kuflierl:
### Prerequisites

- [X] This is **NOT** a [YouTube](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [Facebook](https://www.github.com/uBlockOrigin/uAssets/issues/3367) or [Twitch](https://www.github.com/uBlockOrigin/uAssets/issues/5184) report.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Search [here](https://github.com/uBlockOrigin/uAssets/issues) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [ ] I did not answer truthfully to **ALL** the above checkpoints.

### URL(s) where the issue occurs.

`https://www.baeldung.com/cs/`

### Description

Soft Adblock Detection Popup  
This seems to fix the issue: 
`||www.baeldung.com/cs/wp-content/themes/baeldung/library/dist/js/ad-recovery.min.js$script,1p`

### Other extensions used

None

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

![Screenshot 2023-12-20 at 14-12-59 Baeldung on Computer Science](https://github.com/uBlockOrigin/uAssets/assets/41301536/2f7883bb-642e-4061-ace4-d367391790e3)

</details>

### Configuration

<details>
<summary>Details</summary>

```yaml
uBlock Origin: 1.54.0
Firefox: 121
filterset (summary):
 network: 106479
 cosmetic: 44167
 scriptlet: 19502
 html: 1143
listset (total-discarded, last-updated):
 default:
  user-filters: 0-0, never
  ublock-filters: 37318-232, 22m
[...]

PR #23503, opened 2024-04-30
Update block of baeldung.com recovery.min.js

kuflierl:
baeldung.com adblock-detection blocking was done in accordance with issue #21501.
`ad-recovery.min.js` was later renamed to `recovery.min.js`. This commit updates the filename.


kuflierl:
yeah, it works out

PR #23639, opened 2024-05-10
Update badware.txt

Polycarbohydrate:
added malicious domain


JobcenterTycoon:
`https://ryosx.cc^$doc` is not a good filter because it doesn’t block `http://ryosx.cc/`.

The site itself is malicious, blocked here: https://github.com/uBlockOrigin/uAssets/commit/64c95615fcbf5d1f4468bdf40b78b72a29bbe16f

PR #23664, opened 2024-05-11
Add kingexploits.com due to spread of malware

Polycarbohydrate:
Added kingexploits.com to the badware filter.


JobcenterTycoon:
Thanks.

PR #23763, opened 2024-05-19
address skuola.net anti copy

zWolfrost:

go to `https://www.skuola.net/` and click on any article (like `https://www.skuola.net/news/social-trend/quali-migliori-open-world-top-5.html`), then try to copy the text content of the article.

### Describe the issue

The text won't be copied, or saved to the clipboard.

### Versions

- Browser/version: Librewolf 125.0.3-1
- uBlock Origin version: 1.57.2

### Settings

uBlock's default


JobcenterTycoon:
Thanks works fine.

PR #23780, opened 2024-05-21
Modiji Up

BlazeFTL:
`modijiurl.com/ZTnw`


BlazeFTL:
@D4niloMR Can You Delete Comments That Are Useless Like This ?
Those 3 Comments Serve No Purpose
So If That Can Be Done For The Entire Thread Or As Much As Possible If You Are Busy Then It Would Be Great
IMG_20240521_213630

D4niloMR:
Sorry, I won't wipe the thread that way, but maybe other member want to do so.

BlazeFTL:
I Cant Keep Writing Paragraphs To Explain Why I Think My Decisions Are Better
Nevertheless I Wont Be Contributing Anymore
Had A Great Time Here

BlazeFTL:
Its Probably Better That A Idiot Like You Dont See Link The Link

PR #23797, opened 2024-05-22
quickfix https://forums.lanik.us/viewtopic.php?p=167211-olx-ua

dportvine:
This is not an advertisement, but a rating, although ads are added there on a paid basis. Top ads have only one copy.
`https://forums.lanik.us/viewtopic.php?p=167211-olx-ua`, https://github.com/easylist/easylist/commit/69ae45d63be4c34fc44ea7d7028765e33119193b


gorhill:
Should be fixed in EasyList?

cc @ryanbr

dportvine:
Yes, high traffic sites. I suggest to add, temporarily of course.

PR #23818, opened 2024-05-23
fixed in EasyList

dportvine:


JobcenterTycoon:
Too early, it would be better to remove after 5 days.

PR #23877, opened 2024-05-28
address https://github.com/easylist/easylist/issues/19226

dportvine:


stephenhawk8054:
@gorhill This site needs epoch time `1716874494` in the cookie value. I wonder if it's possible to have `$now$` for `trusted-set-cookie` ?

gorhill:
> I wonder if it's possible to have $now$ for trusted-set-cookie ?

It's already supported: <https://github.com/gorhill/uBlock/blob/1.58.0/assets/resources/scriptlets.js#L4231>

stephenhawk8054:
Hmm... Weird, when I use

```adb
abw.by##+js(trusted-set-cookie, cookiePolicy, %7B%22accepted%22%3Atrue%2C%22technical%22%3Atrue%2C%22statistics%22%3Afalse%2C%22marketing%22%3Afalse%2C%22expire%22%3A$now$%7D)
```

The cookie value still has `$now$` instead of the date

image

Or should I use unescape version?

stephenhawk8054:
Oh wait, it's because setcookie is using `value = Date.now();` instead of `value = value.replaceAll('$now$', Date.now());` like in local storage.

Yeah sorry I meant using it inside a string.

gorhill:
> I meant using it inside a string

https://github.com/gorhill/uBlock/commit/0e1e4b82c5

By the way, you don't need to pre-encode the JSON data, this also works:

    abw.by##+js(trusted-set-cookie, cookiePolicy, {"accepted":true,"technical":true,"statistics":false,"marketing":false,"expire":$now$})

Issue #22973, opened 2024-03-18
c.lazada.co.id: popups


rzgriz:
### Prerequisites

- [X] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://github.com/uBlockOrigin/uAssets/discussions/17361) report. _These sites MUST be reported by clicking their **respective links**_.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin along with other content blockers.
- [X] I have verified that the web browser's built-in content blocker/tracking protection or [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [X] I have verified that other extensions are not causing the issue.
- [ ] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://c.lazada.co.id/t/c.bTN9Uj`

### Category

popups

### Description

Annoying popup of online market app from south east asia

### Other extensions used

None

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.56.0
Firefox Mobile: 122
filterset (summary):
 network: 126730
 cosmetic: 37532
 scriptlet: 19048
 html: 1716
listset (total-discarded, last-updated):
 default:
[...]

Issue #17905, opened 2023-05-02
timesofindia.com: nuisance


Malgotra:
### Prerequisites

- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I forced an update of my filter lists. (Click the ["Purge all caches"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches) button while holding the 'Shift' key, then click the ["Update now"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now) button.)
- [X] I did not remove any of the [default filter lists](https://user-images.githubusercontent.com/124740436/235392297-c4fc6290-73a6-4a70-96a9-3f6a7683f21d.png), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable external filter lists, or I have verified that the issue still occurs without enabling external filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [ ] I did not answer truthfully to **ALL** the above checkpoints.

### URL address of the web page

`https://m.timesofindia.com/business/international-business/in-ai-fight-between-pichais-google-and-nadellas-microsoft-arvind-krishnas-ibm-and-many-others-will-bleed-jobs/articleshow/99941441.cms`

### Category

nuisance

### Description

App install annoyance in middle of the page

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

![Screenshot_20230502-220837~2](https://user-images.githubusercontent.com/122884037/235730270-93888e23-5858-4e18-b3be-5f53fdd623af.png)

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.49.0
Firefox Mobile: 112
filterset (summary): 
 network: 220066
 cosmetic: 223105
 scriptlet: 39075
 html: 1114
listset (total-discarded, last-updated): 
 added: 
[...]

PR #23901, opened 2024-05-29
address https://github.com/AdguardTeam/AdguardFilters/issues/179947

dportvine:


stephenhawk8054:
I think it's better be added to EL annoyance list

dportvine:
This banner appears every time, the user's choice is not stored in cookies. You can also download the APPS below and choose a platform.

<details>
<summary>Screen</summary>


</details>

Also, I see at least two rules in this filter that are clearly not ads
https://github.com/uBlockOrigin/uAssets/issues/17905
https://github.com/uBlockOrigin/uAssets/issues/22973

stephenhawk8054:
I can't check lazada page, but for timesofindia, there's no options for users to close the popup, unlike this case.

But in case the popup appears at every load, it can be addressed.

dportvine:
> But in case the popup appears at every load, it can be addressed.

Here's the video if you don't believe me.
<details>
<summary>Screen</summary>

https://github.com/uBlockOrigin/uAssets/assets/111344219/3b2c8124-b137-445a-b598-a9224b842e61
</details>

PR #24035, opened 2024-06-10
Update shorteners domain clk.asia

akn19:

`https://clk.wiki/kTv89`

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: Chrome v125
- uBlock Origin version: 1.58.0

### Settings

### Notes


D4niloMR:
Add `apekite.com` too.

Issue #24130, opened 2024-06-17
modrinth.com: ads


Zephiii:
### Prerequisites

- [X] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://github.com/uBlockOrigin/uAssets/discussions/17361) report. _These sites MUST be reported by clicking their **respective links**_.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin along with other content blockers.
- [X] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [X] I have verified that other extensions are not causing the issue.
- [X] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://modrinth.com/mod/more-furnace-variants-lieonlion/version/1.0.2+1.20.4-Fabric`

### Category

ads

### Description

Numerous mods hosted on the site have ads for server hosting companies injected into the descriptions by the mod authors. It's difficult to reliably block these.

### Other extensions used

none

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

[...]

PR #24165, opened 2024-06-19
 Add curseforge and additional ad domains for fix of #24130 

raqean:

`https://modrinth.com/resourcepack/dramatic-skys`
`https://modrinth.com/mod/tips`
`https://modrinth.com/mod/the-filler-update`
`https://modrinth.com/modpack/quantum-freedom`
`https://www.curseforge.com/minecraft/modpacks/technical-electrical`

### Describe the issue

Block additional domains that are used by modrinth and curseforge in-content banner ads. Added curseforge as an additional domain where this format of banner ad is common, but with a different element layout.

### Screenshot(s)

Screenshot 2024-06-19 at 04 23 03
Screenshot 2024-06-19 at 04 23 11

### Versions

- Browser/version: Firefox 127
- uBlock Origin version: 1.58.0
- Basing my changes off of the latest commit related to these sites at the time.

### Settings

no changes

### Notes

none

### DIFF
--- a/filters/filters-2024.txt
+++ b/filters/filters-2024.txt
@@ -2149,6 +2149,5 @@ game8.co##.js-side-ads-movie-container:remove()
 
 ! https://github.com/uBlockOrigin/uAssets/issues/24130
 modrinth.com##section.normal-page__content :is([href^="https://shockbyte.com/partner/"],[href^="https://billing.kinetichosting.net/"],[href^="https://bisecthosting.com/"],[href^="https://meloncube.net/"],[href^="https://scalacube.com/"],[href^="https://billing.ember.host/"],[href^="https://nodecraft.com/"]) > img
-curseforge.com##section.project-description :is([href*="shockbyte.com"],[href*="billing.kinetichosting.net"],[href*="bisecthosting.com"],[href*="meloncube.net"],[href*="scalacube.com"],[href*="billing.ember.host"],[href*="nodecraft.com"]) > img
-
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Can you convert to `:is()` format, so it's easier to add domains without adding more lines?

Also, can you change `href*` to `href^`? We want to minimize false positives and target correct elements on the page.

raqean:
I used `href*` instead of `href^` because on curseforge the target domain is buried within a url-encoded redirection link. I've applied the `:is` syntax.

stephenhawk8054:
From what I see, that site has `/linkout?remoteUrl=` at the start. Does this filter work?

```adb
curseforge.com##section.project-description [href^="/linkout?remoteUrl=http"]:is([href*="shockbyte.com"],[href*="billing.kinetichosting.net"],[href*="bisecthosting.com"],[href*="meloncube.net"],[href*="scalacube.com"],[href*="billing.ember.host"],[href*="nodecraft.com"]) > img
```

leo60228:
It's likely desirable to also support the legacy.curseforge.com subdomain (which serves a role roughly equivalent to old.reddit.com).

raqean:
Yes that seems to work. I'll update the PR.

raqean:
> It's likely desirable to also support the legacy.curseforge.com subdomain (which serves a role roughly equivalent to old.reddit.com).

That's a good point, I'll update the filter.

stephenhawk8054:
`curseforge.com##` filter will also cover all subdomains of the page as well, so you don't need to add `,legacy.curseforge.com`

stephenhawk8054:
Thanks. Please comment when you think the PR is complete.

raqean:
I think it's complete, I couldn't find any other sponsor domains. There might be more, but they're hard to discover automatically. People would need to report them to get added individually.

mrshmllow:
This is extremely upsetting to see

CDAGaming:
It is, but this'll end up just being another cat and mouse game. People already have bypasses up and running, so it'll be interesting to see how this all pans out.

It's important to remember who cast the first stone in all this.

PR #24178, opened 2024-06-19
aha-music.com - hide Chromium extensions for Firefox users

anewuser:

`https://www.aha-music.com/`

`https://www.aha-music.com/8bc80319a6da5d39a9647a1271552461`

### Describe the issue

Big buttons that add no value for Firefox users.

buttons

buttons 2


stephenhawk8054:
We don't actually hide buttons for specific platforms, unless they are too intrusive.

Anyone else's opinions?

PR #24221, opened 2024-06-22
Unblock the sponsorships/ads that are unobtrusive.

PieyIsAPie:
This just allows the unobtrusive Modrinth ads. Maybe block some ads that are actually obtrusive instead of things that don't annoy anyone?


stephenhawk8054:
uBO follows [EasyList's policy](https://easylist.to/pages/policy.html), including affiliate images. Some of the filters are already in EasyList, removing the filters from uBO does not change the results. If you want to discuss about the policy, please discuss in another more proper places including EasyList and other content blockers using EasyList as well.

PR #24285, opened 2024-06-28
feat: Add extended polyfilio domains

SISheogorath:
### Describe the issue

According to the news article by bleepingcomputers and in line with the announcement from Google on the topic, the number of impacted CDNs by this malware campaign is even bigger.

See: https://www.bleepingcomputer.com/news/security/polyfillio-bootcdn-bootcss-staticfile-attack-traced-to-1-operator/

This will probably break a lot of stuff. But I guess it has to break.

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -4201,8 +4201,18 @@ oltacidergisi.com##+js(acs, atob, minimalDrainValue)
 ! https://github.com/uBlockOrigin/uAssets/pull/24255
 ! https://sansec.io/research/polyfill-supply-chain-attack
 ! https://www.bleepingcomputer.com/news/security/polyfillio-javascript-supply-chain-attack-impacts-over-100k-sites/
+! https://www.bleepingcomputer.com/news/security/polyfillio-bootcdn-bootcss-staticfile-attack-traced-to-1-operator/
 ||polyfill.io^$all
 ||googie-anaiytics.com^$all
+||bootcdn.net^$all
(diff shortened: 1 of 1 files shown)


vicevolf:
Is this an accurate approach? Although these CDNs have been compromised by security risks, completely blocking these commonly used CDNs in China could directly lead to abnormal or even broken loading of many websites, severely affecting user experience and forcing them to disable plugins and leave the security protection. Are there more precise filtering rules? It is necessary to balance security risks with the introduction of new problems.

`||bootcdn.net^$all
||bootcss.com^$all
||staticfile.net^$all
||staticfile.org^$all`

图片

SISheogorath:
Given that we have to assume that all these CDNs are operated by the same entity according to the leaked Cloudflare account credentials, and have been observed serving malware, there is probably no more precise rule to use here.

When I made the Pull Request I had similar concerns. Hence:

> This will probably break a lot of stuff. But I guess it has to break.

Is the cure worse than the disease? I don't think so, uBlock origin already allows you to manage individual blocks per page, if you really want to unblock something. Otherwise its up to the broken websites to update their dependencies or use a plugin that rewrites them.

Given the amount of Chinese language involved with the operator one could even argue that their targets are probably even in China and Chinese speaking communities.

qingtingxianzhang:
You made only two commit on this project, yet caused countless Chinese developers to have to redo a lot of work. Great job, you're really something!

fuck you

2024-10-30

PR #24391, opened 2024-07-06
monitoruldevrancea. ro contextmenu select

SeamlessRelaxation:
`https://www.monitoruldevrancea.ro/2024/07/05/jandarmii-vranceni-la-datorie-pentru-siguranta-cetatenilor-la-zilelelor-focsaniului-si-pe-traseele-montane/`


JobcenterTycoon:
Looks a bit unsafe, i would use safer selections like `stopPrntScr` instead of `tooltip`

SeamlessRelaxation:
feel free to edit

PR #24761, opened 2024-08-04
Fix unescaped comma in replace modifier

chrisduerr:
I was just looking at the files trying to understand the rules format and I'm pretty sure that commas need to be escaped since they're used as separators for modifiers?

Maybe uBlock handles this just fine, but it is wrong according to adguard's docs (which uBlock references).

If this is not an issue or if this is something that is more widespread and should be fixed for all incorrect replace regexes, feel free to just close this issue.


stephenhawk8054:
I don't think it's an issue for uBO.

@gorhill Sorry, what do you think?

chrisduerr:
> I don't think it's an issue for uBO.

I'd assume that the parser is basically just looking for the next two `/` to terminate the replace? So the only thing that **must** be escaped would be `/` then?

gorhill:
There is no need to escape the comma if there is no error reported by the parser.

PR #24956, opened 2024-08-19
strip GitHub `notification_referrer_id` query param

brian6932:

`https://github.com/notifications`

### Describe the issue

For some reason this query param can cause a near infinite redirect, as well as the base64 encoded click tracking. Removing it doesn't effect notifications being marked as read or not.

It does seem that the param being set's required for the following shelf to show up on notifications, so I could see this being an unwanted change: 

### Versions

- Browser/version: Firefox 129
- uBlock Origin version: 1.59.0


stephenhawk8054:
I can't reproduce the breakage. For param removal issue, please create issue in `AdGuard URL Tracking Protection` list first: https://github.com/AdguardTeam/AdguardFilters/issues

The `removeparam` section in uBO is only for common params and low chances of breakages. If the filter is stable in that list for a while, it can be adapted to uBO's default list.

PR #25100, opened 2024-08-29
[youtube] Improve ads skip when video loop enabled

Khang-NT:
Fix - ublock jumps into the middle of the playing video when it starts looping

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: [here]

### Settings

### Notes


stephenhawk8054:
Can you test if this works for loop issue?

```adb
www.youtube.com##+js(trusted-rpnt, script, (function serverContract(), 'console.log("uBO - injected");(()=>{if(ytInitialData.topbar.desktopTopbarRenderer.logo.topbarLogoRenderer.iconImage.iconType==="YOUTUBE_PREMIUM_LOGO")return;if(location.href.startsWith("https://www.youtube.com/tv#/"))return;document.addEventListener("DOMContentLoaded",(function(){const skipAd=()=>{const videoPlayer=document.getElementById("movie_player");if(!videoPlayer)return;console.log(videoPlayer.getStatsForNerds()?.debug_info,"uBO - debug");if(!videoPlayer.getStatsForNerds()?.debug_info?.startsWith("SSAP, AD"))return;const progress=videoPlayer.getProgressState();console.log(JSON.stringify(progress),"uBO - progress");if(progress&&progress.duration>0){videoPlayer.seekTo(progress.duration)}};skipAd();const observer=new MutationObserver((()=>{skipAd()}));observer.observe(document,{childList:true,subtree:true})}))})();(function serverContract()', sedCount, 1)
```

Khang-NT:
Yes it works, because the check `if (!videoPlayer.getStatsForNerds()?.debug_info?.startsWith('SSAP, AD')) return;` is perfectly.
But if there are N ads (N > 1), uBO will seek n times, it glitches and wastes data. You can copy the old logic where uBO calculates total of n ads duration, then seek will be more effective.

stephenhawk8054:
> You can copy the old logic where uBO calculates total of n ads duration, then seek will be more effective.

We know about multiple ads case. However, we did test it with a user with an edge case of `html5_enable_ssap_entity_id` returning `undefined` and seeking towards the end (of ads) in cases of multiple ads caused a lot of instabilities where the player becomes stuck and cannot continue playing.

Also the old logic just automatic assumes that the ads time is at the nearest end of the `ssapConfig` data which only deals with the information about the ads at the start of the video. If YT starts to do SSAP ads in the middle or at the end, the conditions will become more complicated are more prone to errors. Hence I think it's better to skip at each ads to easily debug and adjust in edge / future cases.

stephenhawk8054:
It's already added.

PR #25177, opened 2024-09-02
add roblox.*

SomeAspy:

```yml
roblox.com.kg
roblox.tz
```

### Describe the issue

Account stealers use other TLDs of the roblox site to steal accounts, by mimicking the roblox login page.
see:
[roblox.tz/login?returnUrl=4614319469131406](https://roblox.tz/login?returnUrl=4614319469131406)
[roblox.com.kg/users/5903770680/profile](https://roblox.com.kg/users/5903770680/profile)
I have a reasonable suspicion that there are many more like this.

### Notes

Only official domain: https://roblox.com (source: https://devforum.roblox.com/t/list-of-roblox-apis-subdomains-availability-information/2212126)


krystian3w (approved):

ryanbr:
This seems to quite an aggressive commit, maybe just block the specific domains

SomeAspy:
I am pretty sure there are more TLDs similar to this than would make sense blocking individually - besides, the only official Roblox domain is Roblox.com, anything else should already be deemed suspicious.
If there are issues, I think it would be worth a case-by-case inspection.
Roblox is already a bit of a target for scams, since its primary audience is younger children - how many of these younger children have uBlock installed, I'm unsure, but I feel the real amount of false-positives this rule cause are minimal.

ryanbr:
Just keep in mind anyone with a  legit "roblox" in the subdomain url, will probably break here.

SomeAspy:
Oh, that's true. I based it off some other wildcard filter I found. I will have to look sometime tomorrow.

stephenhawk8054:
You can use this

```adb
||roblox.*^$all,to=roblox.*|~roblox.com
```

It will block `roblox.kg` but not `roblox.example.com`

stephenhawk8054:
Since `github.io`, `gitlab.io`, `pages.dev`... are registered as [public suffixes](https://publicsuffix.org/list/public_suffix_list.dat), we cannot use `to=roblox.*` here: https://github.com/uBlockOrigin/uAssets/issues/25221

So it's better to specify the exact TLD / public suffix that are malicious.

SomeAspy:
Am I missing someting, or is there no way to effectively block anything but the safe TLD, without capturing subdomains?

stephenhawk8054:
`roblox.github.io` is not subdomain. `github.io` is an eTLD

SomeAspy:
I see. That's quite interesting. I suppose I should have done more research before blanket banning roblox.*
My apologies

PR #25382, opened 2024-09-20
remove uploady.io fake download button

Rust1667:

`https://uploady.io/thpn66davgb9`

### Describe the issue

fake download button

### Screenshot(s)

image

### Versions

- Browser/version: [here]

### Settings

### Notes


stephenhawk8054:
Does `uploady.io###dlnk` work?

Rust1667:
yes it works

Issue #25387, opened 2024-09-21
jetpunk.com: detection


LeopoldTal:
### Prerequisites

- [X] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://github.com/uBlockOrigin/uAssets/discussions/17361) report. _These sites MUST be reported by clicking their **respective links**_.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin along with other content blockers.
- [X] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [X] I have verified that other extensions are not causing the issue.
- [X] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://www.jetpunk.com/quizzes/genius-level-general-knowledge-4`

### Category

detection

### Description

Regression: #21314 is back

**To reproduce**: While logged in, attempt to rate or comment on any quiz. A popup modal asking the user to disable their adblocker appears.

**Cause**: The website attempts to detect ad blockers by checking that Google Tag Manager is present: `'object' == typeof window.google_tag_manager`. But GTM is blocked by:

[...]

Issue #21314, opened 2023-12-09
jetpunk.com: detection


snakeoildeluxe:
### Prerequisites

- [X] This is **NOT** a [YouTube](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [Facebook](https://www.github.com/uBlockOrigin/uAssets/issues/3367) or [Twitch](https://www.github.com/uBlockOrigin/uAssets/issues/5184) report.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.)
- [X] I forced an update of my filter lists. (Click the ["Purge all caches"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#purge-all-caches) button while holding the 'Shift' key, then click the ["Update now"](https://github.com/gorhill/uBlock/wiki/Dashboard:-Filter-lists#update-now) button.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [ ] I did not answer truthfully to **ALL** the above checkpoints.

### URL address of the web page

`https://www.jetpunk.com/`

### Category

detection

### Description

Page detects the use of both EasyList and ublock-filters.

**To reproduce**: While logged in and using uBO with default filter lists, attempt to rate a quiz. A popup window asking the user to disable their adblocker is presented.

**Detection vector, EasyList**: "canary" script, `ad-engine.js`, globally blocked by EasyList. Script only creates an empty, hidden `div`, the absence of which is used for detection.

**Detection vector, uBO-f**: `!window.asc || window.asc == 1`

[...]

PR #25575, opened 2024-10-08
Fix jetpunk.com detection again

LeopoldTal:
jetpunk.com used to detect adblockers by checking whether `google_tag_manager` exists, but now they also check an attribute of the object. Add a filter to provide this attribute as well.

`https://www.jetpunk.com/user-quizzes/117765/country-by-emojis`

### Describe the issue

Regression: #25387 is back, itself a regression of #21314.

To reproduce: While logged in, attempt to rate or comment on any quiz. A popup modal asking the user to disable their adblocker appears.

### Screenshot(s)

See linked issues

### Versions

- Browser/version: Firefox 131.0
- uBlock Origin version: 1.60.0

### Settings

- Default

### Notes

The relevant jetpunk.com code is:

```javascript
      isAdblocker() {
        return !(
          this.user &&
          (
            this.user.isPremium ||
            l.isQuizmaster(this) ||
            1 == this.user.isjetpunk
          ) ||
          'object' == typeof window.google_tag_manager &&
          'object' == typeof window.google_tag_manager['G-Z8CH48V654'] && // <- this causes the regression
          document.getElementById('CxkDwmaFMXvB')
        )
      }
```

This could be a bit of a cat-and-mouse game… happy to work on a more general solution if you can think of one.

### DIFF
--- a/filters/filters-2021.txt
+++ b/filters/filters-2021.txt
@@ -303,6 +303,7 @@ jetpunk.com##+js(set-local-storage-item, PageCount, $remove$)
 jetpunk.com##+js(set, asc, 2)
 ! https://github.com/uBlockOrigin/uAssets/issues/25387
 jetpunk.com##+js(set, google_tag_manager, {})
+jetpunk.com##+js(set, google_tag_manager.G-Z8CH48V654, {})
 @@||jetpunk.com^$script,1p
 @@||jetpunk.com^$ghide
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
I think it's fine. There are other ways that can defuse it without specifically set that key, but any defusing method can still be countered anyways.

PR #25625, opened 2024-10-11
Recommend a way to address paywalls

Happypig375:
See if you endorse this recommendation.


Happypig375:
@uBlock-user I updated the text to mention a specific DMCA takedown instead of recommending a specific browser extension. Is this acceptable now?

PR #25811, opened 2024-10-27
Rewrite readme and fix workflows

permanentdaylightt:
Fixed workflows so theyre getting run as described


garry-ut99:
By the way, maybe would be good to include also changes from this one as well, as it adds missing things that are handled in uAssets but were not ever mentioned so far (I would also add: redirections):
- https://github.com/uBlockOrigin/uAssets/pull/17008

garry-ut99:
Yes, means a group of: [ `href-sanitizer` & `urlskip` & `removeparam` & `redirect` & `redirect-rule` ] filters. You may name the group whatever you like / fell, maybe: "Redirections/URLs cleaning" or "Redirections/Links cleaning".

PR #26297, opened 2024-12-04
update dead link in filters/lan-blick.txt

mrd83:
Since there's no "uBlock-discussions" repo neither in https://github.com/uBlockOrigin/ nor https://github.com/gorhill/ and ...uBlockOrigin/uBlock-issues/discussions/910 links to an issue anyway - I decided to put the destination link here. Note that I'm not sure is it the issue you wanted linked here - but the original link was dead anyway!


mrd83:
I see. But still - the link is present in **publicly available** file (also in ublock asset viewer, obviously) - you may (or may not ;]) want to change it or remove - to avoid confusion from other minions like me 🙇🏻‍♂️

stephenhawk8054:
Some issues stemmed from internal discussions, and we need references to re-check when necessary too.

PR #26402, opened 2024-12-10
Update eltiempo.es & otempo.pt rules

ryanbr:
Consent walls on `eltiempo.es` & `otempo.pt`


krystian3w:
Second filter seems ate `)` character, so won't help close cookie wall.

mapx-:
fixed

PR #26754, opened 2025-01-06
Update instagram,facebook consents

ryanbr:
Instagram is using changing uuid's for declining consents. Causing issues for users logging in via email

Better option is to use a trusted click here;

ref: https://github.com/brave/brave-browser/issues/42993


ryanbr:
Added removal of old consent value. (safe to remove after a while)

PR #27014, opened 2025-01-29
Fix existing tweakers.net cookie dialog handling

kaoneko:

https://tweakers.net/

### Describe the issue

The existing rule has no effect (anymore), the cookie consent dialog is not dismissed. I have updated the rule to target the proper CSS selectors.

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -1865,10 +1865,6 @@ eltiempo.es##+js(trusted-click-element, [target="_self"][type="button"][class="_
 eltiempo.es##+js(trusted-click-element, button[type="button"][class="_button_15feu_3"], , 1000)
 otempo.pt##+js(trusted-click-element, [target="_self"][type="button"][class="_10qqh8uq"], , 1000)
 
-! https://github.com/uBlockOrigin/uAssets/issues/25952
-tweakers.net##+js(trusted-click-element, #pg-shadow-root-host >>> button#pg-reject-btn)
-tweakers.net##+js(trusted-click-element, #pg-shadow-root-host >>> button#pg-configure-btn)
-
 ! accepting only required cookies
 gardengirls.de##+js(trusted-set-cookie, Cookie, accept_cookies\,\,, , , reload, 1)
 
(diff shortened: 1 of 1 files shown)


kaoneko:
Looking at this a bit longer, I see the real problem is the outdated rules for `myprivacy.dpgmedia.nl` in `annoyances-cookies.txt`, which fail to handle the DPG Media cookie wall correctly because of the same outdated CSS selectors and maybe conflicting rules too.

When visiting a DPG Media site like `tweakers.net` for the first time, one gets redirected to e.g. `https://myprivacy.dpgmedia.nl/consent?callbackUrl=https%3A%2F%2Fwww.tweakers.net%2F&siteKey=ehnsj7n39x9wm5sy` which handles the privacy choices for _all_ DPG Media sites. When the user succesfully sets their preferences there once, visiting any DPG Media site doesn't show a cookie consent dialog anymore. However, currently the `annoyances-cookies.txt` doesn't handle this correctly, resulting in a cookie consent dialog on each individual DPG Media host (tweakers.net, nu.nl, reclamefolder.nl etc.)

While this pull request is an improvement for now, it doesn't address the real problem. If the problem can be addressed on the level of dpgmedia.nl, we don't need a separate rule for tweakers.net anymore. So far I got this:
> myprivacy.dpgmedia.nl##+js(trusted-click-element, '#pg-host-shadow-root >>> button#pg-configure-btn, #pg-host-shadow-root >>> #purpose-row-SOCIAL_MEDIA input[type="checkbox"], #pg-host-shadow-root >>> button#pg-save-preferences-btn')

Which is a corrected version of one of the existing dpgmedia.nl rules (which allows embedded social media). It handles all DPG Media sites at once for the ones I tested (tweakers.net, nu.nl, reclamefolder.nl) when I disable the _uBlock Filters - Cookie notices_ list in UBO (which has one or more conflicting rules).

This needs more looking into and shouldn't block the merging of this pull request, but may in the end make this pull request obsolete, if I or anyone has the time to really look into properly handling all DPG Media sites.

kaoneko:
I'm on it, expect a new pull request soon; closing this one.

PR #27695, opened 2025-03-22

BlazeFTL:
Removed Rules Are No Longer Used


BlazeFTL:
Missed(Line 207)
```
/nt.js$domain=fc-lc.xyz
```
I Get Captcha Error With Ublock In FF Android Though...But Without UBO It Works 
`fc-lc.xyz/ltio`

mapx-:
> Missed(Line 207)
> /nt.js$domain=fc-lc.*

what do you mean ?

BlazeFTL:
Needs To Be Added

PR #28132, opened 2025-04-30
Fix max.com trusted-click not working

ryanbr:
`https://auth.max.com/login` (Spanish IP tested) 

Wasn't working consistently, since in some instances the loading of the max webpage would be delayed (took 5-7secs to load here). rather than extend the time out, lets just set the cookie. Cookie value is set ="Reject".


ryanbr:
Speaking to a user, the consent took 12 seconds to load on max.com

PR #28192, opened 2025-05-06
add urlskip filter for physicsandmathstutor.com to access pdf files d…

mylesbartlett72:
…irectly

Any URL of the form
`https://www.physicsandmathstutor.com/pdf-pages/?pdf=[some URL-escaped URL]`
For example:
`https://www.physicsandmathstutor.com/pdf-pages/?pdf=https%3A%2F%2Fpmt.physicsandmathstutor.com%2Fdownload%2FPhysics%2FA-level%2FPast-Papers%2FCAIE%2FPaper-5%2FQP%2FJune%202008%20QP.pdf`

### Describe the issue

The website uses a wrapper page to display PDF files alongside advertisements, by iframing the PDF based on a url parameter.  The PDF file can be accessed directly from the URL in said parameter, which removes the sidebar.

uBO already removes the adverts from the sidebar, but the sidebar itself remains.  This pull request adds a filter to remove the sidebar by redirecting requests to the pdf iframe page to the referenced pdf.

### Screenshot(s)

Note that the issue most likely does not occur on mobile or safari (see note 3)

Without the filter, a (blank with uBO, advertising-filled without uBO) sidebar appears, and the file referenced in the `pdf` parameter is loaded in an iframe:
2025-05-06T11:28:22,701862875+01:00

With the filter, the pdf file is loaded directly by the browser (and thus no sidebar) (in Firefox, this is equivalent to right-clicking the PDF, hovering over "This Frame" in the context menu, and clicking "Show Only This Frame"
2025-05-06T11:32:05,548153923+01:00

### Versions

- Browser/version: Firefox 135.0.1 (64-bit), from Fedora 41 repositories
- uBlock Origin version: 1.63.2

### Settings

- Custom filters for specific websites, including the filter added by this PR
  - I have tested without these filters and the issue still occurs
- All "Annoyances" filters enabled (issue still occurs regardless of whether or not they are enabled)

### Notes

1)The issue can also be corrected using a cosmetic filter (namely `www.physicsandmathstutor.com##.pdf-sidebar`), but this is less optimal since using `urlskip` reduces the total volume of data transferred (by not loading the HTML for the page as well as the PDF) - therefore I decided not to propose that as a solution.
2)The pages in question are not tracking redirects or similar (as would typically merit a `urlskip` filter as far as I can tell), since they generally don't perform any redirect (see note 3 for the exception), but instead exist to display advertising in a sidebar alongside a PDF.
3)The pages contain inline javascript (in the DOM inspector it is contained in a `<script>` block at the very end of the `<body>`) that serves to determine if the user is using Safari (via user-agent) or a mobile browser (via some heuristics involving determining if the browser has touch support in JS as well as the display width), in which case it sets `window.location.href` to the value of the decoded url parameter `pdf`.  If none of those conditions were met, it instead sets the `src` attribute of an `<iframe>` to this url parameter value (so that the PDF is displayed with the browsers native viewer but the page can still display advertising).
4)You can actually put an arbitrary URL in the `pdf` parameter and it will load in the iframe (and as far as I can tell it will also get redirected to on mobile) since this is not prevented by the pages CSP (or any validation in the inline JS).  Therefore, the urlskip filter does not seem like it would meaningfully weaken security (as arbitrary redirect is already possible).

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -7914,6 +7914,3 @@ simptown.su##.juice_ban
 
 ! https://labreakfastburrito.com/ - pie adblock
 labreakfastburrito.com##+js(aopr, PieScriptConfig)
-
-! https://github.com/uBlockOrigin/uAssets/pull/28192
-||physicsandmathstutor.com/pdf-pages$urlskip=?pdf
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -7915,5 +7915,5 @@ simptown.su##.juice_ban
 ! https://labreakfastburrito.com/ - pie adblock
 labreakfastburrito.com##+js(aopr, PieScriptConfig)
 
-! https://github.com/uBlockOrigin/uAssets/pulls/[PLACEHOLDER SO I CAN GET A PR NUMBER]
+! https://github.com/uBlockOrigin/uAssets/pull/28192
 ||physicsandmathstutor.com/pdf-pages$urlskip=?pdf
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -7914,3 +7914,6 @@ simptown.su##.juice_ban
 
 ! https://labreakfastburrito.com/ - pie adblock
 labreakfastburrito.com##+js(aopr, PieScriptConfig)
+
+! https://github.com/uBlockOrigin/uAssets/pulls/[PLACEHOLDER SO I CAN GET A PR NUMBER]
+||physicsandmathstutor.com/pdf-pages$urlskip=?pdf
--- a/filters/filters-2025.txt
+++ b/filters/filters-2025.txt
@@ -1576,3 +1576,6 @@ asianboy.fans##+js(trusted-suppress-native-method, DOMTokenList.prototype.add, '
 
 ! https://github.com/uBlockOrigin/uAssets/issues/28189
 @@||seelen.io^$script,1p
+
+! https://github.com/uBlockOrigin/uAssets/pull/28192
+||physicsandmathstutor.com/pdf-pages$urlskip=?pdf


mylesbartlett72:
woops i hit enter early let me fill in the fields manually

mylesbartlett72:
Since my proposed solution uses the advanced filter syntax (namely `urlskip`), I believe it is therefore appropriate to contribute it here as opposed to easylist.

My proposed solution also uses a privileged filter (again, `urlskip`), so therefore it is impossible to contribute it elsewhere (if it is to function).

Finally, although this is a purely cosmetic issue, I consider it appropriate to add a filter for since it hides leftover artefacts from advertising that was blocked (and thus improves user experience).

mylesbartlett72:
> I think this can be addressed in main list. Can you move the filter to `filters-2025.txt` file?

of course!

I wasn't sure which list to put it in, and the annoyances list seemed most appropriate.  Would it be best to create a new PR for that and close this one, or use this existing PR and revert the previous commits?

stephenhawk8054:
You can edit the files in this commit.

stephenhawk8054:
Thank you.

krystian3w:
Works too:

```adb
||physicsandmathstutor.com/pdf-pages/?pdf=http$doc,urlskip=?pdf,badfilter
physicsandmathstutor.com##.pdf-sidebar
```

So compatible with ELC.

mylesbartlett72:
> Works too:
> 
> ```
> ||physicsandmathstutor.com/pdf-pages/?pdf=http$doc,urlskip=?pdf,badfilter
> physicsandmathstutor.com##.pdf-sidebar
> ```
> 
> So compatible with ELC.

That does work (as I did point out in my original comment on this PR under note 1), but it results in avoidable network traffic from loading the HTML associated with the iframe, sidebar, and JS to go to the PDF directly (and a side bonus of urlskip is that if someone shares the link to the PDF by copying it from their address bar then someone using the shared link will get the benefits of the filter without needing any adblocker)

If you want to contribute your alternative filter to some upstream filter list, you could do so without the `badfilter` definition, since I can see no reason why each method would be mutually exclusive (with the urlskip filter, the page where the cosmetic filter would take effect is never loaded)

You could also do
```
!#if !ext_ublock
physicsandmathstutor.com##.pdf-sidebar
!#endif
```
alongside my existing filter, so that the cosmetic filter is never parsed in ublock origin (and thus doesn't result in any extra load, and still gives ublock origin users the benefit of the filter even if they disable cosmetic filtering, for example if they are using a low-end device that cannot handle the overhead of cosmetic filtering)

PR #28444, opened 2025-05-21
Update lan-block.txt

Gamer-friend:
Fixed additional Modified code to fit with proxy & port processions.

The exceptions section was verified to ensure that the domains listed were correctly formatted and that the $ modifiers were appropriately applied. This includes checking that the exceptions do not inadvertently allow unwanted traffic.

`All file.`
### Describe the issue
The regex patterns used for matching URLs were checked for correctness. They appeared to be valid, but ensuring they are properly formatted and escaped is crucial. For example, ensuring that special characters like . and [ are correctly escaped.

### Screenshot(s)
None

### Versions

- Browser/version: [Chrome/Edge]
- uBlock Origin version: [Any]

### Settings
None- did not change.

### Notes

Some comments were redundant or overly verbose. I kept the essential comments that provide context or links to issues while removing unnecessary ones. I also Ensured that the structure of the filter rules was clear and that there were no missing or extra characters that could lead to parsing errors.

I noticed an error at when you are doing the Domain verification, I Verified that the domain matching patterns were consistent and correctly formatted. For instance, ensuring that the domain parameters were consistently using the `~ `operator for negation where appropriate.

### DIFF
--- a/filters/lan-block.txt
+++ b/filters/lan-block.txt
@@ -1,4 +1,4 @@
-﻿[uBlock Origin]
+[uBlock Origin]
 ! Title: Block Outsider Intrusion into LAN
 ! Last modified: %timestamp%
 ! Expires: 29 days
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Why do you replace `\[` and `\]` with `$`? `$` means end of line in those regex. And why do you remove the comment lines? Those are for explanation of the regex structure.

cc @gorhill This looks strange. What do you think?

gorhill:
> This looks strange. What do you think?

That's just wrong.

PR #28891, opened 2025-06-18
Update Gtlinks

BlazeFTL:
`https://gtlinks.me/LBZnC`
Other edits are minor domain update and duplicate removal


stephenhawk8054:
You can use `json:{"tagName": "IFRAME"}` instead of `{ "value": {"tagName": "IFRAME" }}`. It's shorter and the former might be deprecated in the future.

BlazeFTL:
Done !

PR #29055, opened 2025-07-02
Remove dead domains filters/filters-2022.txt

ryanbr:
Pyfuncible removal of dead domains

### DIFF
--- a/filters/filters-2022.txt
+++ b/filters/filters-2022.txt
@@ -46,7 +46,7 @@ ac-illust.com,photo-ac.com##body.modal-open *:style(filter: none!important;)
 freebie-ac.jp##+js(trusted-rpnt, script, currentTime = 1500 * 2, currentTime = 0)
 
 ! D4zz-sites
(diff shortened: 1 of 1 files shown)


Yuki2718:
Generic rules like `*#*/ad/$image,redirect-rule=1x1.gif` should not be removed. I can recover after merge, but nice if you can recover.

ryanbr:
If its just related from a specific commit/website, probably not needed anymore?

ryanbr:
`*#*/ad/$image,redirect-rule=1x1.gif` is more specific than generic, looking at the `#`. Just remove it, and if its seen again just re-add it (unlikely tbh). Unless you can show where else its applicable?

Yuki2718:
The rule is for antiblock .org g207, probably is still useful. OTOH I think `||grow.gab.com/galahad/$media` can be removed.

Yuki2718:
Hmm, `insurancevela.com` is living - but maybe no longer used, as the original shortner domain is parked. Unsure about `||viacom18.moloco.com/adserver/v1$xhr,domain=jiocinema.com`, may have subdomain.

PR #29378, opened 2025-07-28
Remove SoundCloud Banner Ad Space

lexi65:
SoundCloud has started using banner ads. The ads themselves are blocked, but there is a huge block of space which is wasteful, so I have added it to the filter list.

### URL(s) where the ads occur

https://soundcloud.com

### Screenshot(s)

image

### What I did
- Filter out the class that is being used, which is .bannerAd.


stephenhawk8054:
Please report to https://github.com/easylist/easylist

PR #29714, opened 2025-08-19
Fix missed banner and text ads in modrinth descriptions and changelogs

raqean:

1: `https://modrinth.com/datapack/clifftree`
2: `https://modrinth.com/mod/time-on-display`
3: `https://modrinth.com/mod/resourcify`
4: `https://modrinth.com/mod/euphoria-patches`
5: `https://modrinth.com/mod/loot-refill`

### Describe the issue

Banner and text ads in descriptions and changelogs.

### Screenshot(s)

<details>

1: 

2: 

3: 

4: 

5: 

</details>

### Versions

- Browser/version: firefox 141.0 (64-bit)
- uBlock Origin version: 1.65.0

### Settings

-

### Notes

Note that the filter needs to be come more general to cover these cases. `section` needed to be removed since in the changelogs the tag is `div`. The `href` match is now `*=` since there's redirect domains. Images and ad text is wrapped in various tags for obfuscation, so the filters need to ignore the exact structure. We may want to consolidate the filters into just one line and remove the `> img`.
Whether we want to block text ads is another question, up to the maintainers to decide. Ads integrated into the prose are both less intrusive and harder to block, so blocking them would only happen on a case-by-case basis and with greater effort. I've included an example.

### DIFF
--- a/filters/filters-2024.txt
+++ b/filters/filters-2024.txt
@@ -2507,16 +2507,19 @@ myp2p.*##a[rel="sponsored"]
 game8.co##.js-side-ads-movie-container:remove()
 
 ! https://github.com/uBlockOrigin/uAssets/discussions/24177#discussioncomment-9821526
-modrinth.com##section.normal-page__content a[href*="/redirect/"] img
+modrinth.com##.normal-page__content a[href*="/redirect/"] img
 ! https://github.com/uBlockOrigin/uAssets/issues/24130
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
We intentionally use `img` to not block the links that have no image banners. You can remove `>` but don't remove `img`.

stephenhawk8054:
I'll fix this issue.

PR #29722, opened 2025-08-20
Update Gtlinks And tpi.li

BlazeFTL:
Domain update
Adblock update for tpi


stephenhawk8054:
For `csp`, you can use this instead:

```adb
*$doc,csp=worker-src 'none',to=blog.potterworld.co|hipsonyc.com|tech.pubghighdamage.com|blog.itijobalert.in|techkhulasha.com|tech.trendingword.com|gyanitheme.com|pubgaimassist.com
```

it will be better compatible to uBOL.

BlazeFTL:
Updated...
`https://gtlinks.me/9AUxf`
Can you reproduce the timer not getting boosted on first page ?
Testing on ff nightly android,ubo dev
This happens now only once

stephenhawk8054:
I can't reproduce, but I'm not sure.

PR #29827, opened 2025-08-29
Update filters-2025.txt

Hanifalm:

### Describe the issue

### Screenshot(s)

### Versions

- Browser/version: [here]

### Settings

### Notes

### DIFF
--- a/filters/filters-2025.txt
+++ b/filters/filters-2025.txt
@@ -1204,8 +1204,8 @@ darts-scoring.com##+js(no-fetch-if, googlesyndication)
 ###close-teaser
 ! https://github.com/uBlockOrigin/uAssets/issues/27519
 ###teaser3[style^="width: 100%;height:0;text-align: center;display: scroll;position:fixed;"]
-dramafren.org##+js(noeval-if, rejectMethod)
-dramafren.org##+js(nowoif, _blank)
+dramafren.com##+js(noeval-if, rejectMethod)
+dramafren.com##+js(nowoif, _blank)
 ||imasdk.googleapis.com/js/sdkloader/ima3.js$script,redirect=google-ima.js,domain=ryderjet.com
 ||ryderjet.com/assets/jquery/css100.js
 


Hanifalm:
Hello, I am the owner of a small website, and I am writing this message with sincerity and hope. I kindly request your consideration to remove my website from the filter. At the moment, my earnings are only about one dollar per day. This amount is very modest and makes it extremely difficult for me to cover even my basic living expenses, let alone the cost of maintaining hosting and keeping the site online.

Running this website is one of the few ways I can support myself, and I truly put in my best effort despite the limited resources I have. The filter has made it even more challenging for me to sustain this work. I deeply value the opportunity to reach people through my site, and I am only asking for a fair chance to continue without being held back.

I understand that filters are important and exist for good reasons, but I kindly hope you might reconsider in my case. Even a small change could make a meaningful difference in my ability to survive and keep moving forward. Thank you very much for your time, understanding, and compassion.

mapx-:
The best way for you is to ask your users for donations.

stephenhawk8054:
We don't whitelist sites. If it's soft anti-adblock that users can turn off and not return in a short time, we'll consider to move it to annoyance list.

Hanifalm:
I once added a donation button, but not a single person made a donation.

Hanifalm:
> The best way for you is to ask your users for donations.

I once added a donation button, but not a single person made a donation.

Hanifalm:
> We don't whitelist sites. If it's soft anti-adblock that users can turn off and not return in a short time, we'll consider to move it to annoyance list.

I kindly ask if you could reconsider my case. My website is very small and I earn only a minimal amount from it, just enough to cover hosting costs. I tried implementing a donation button before, but unfortunately no one donated. Advertising is currently the only way I can keep the website online.

stephenhawk8054:
No. We don't "reconsider" any case. 1 whitelisted website means any site can go here and ask for "reconsider" and this project is dead.

There's no "Acceptable Ads" program here: https://github.com/gorhill/uBlock/blob/master/MANIFESTO.md

As we said, we don't block donations, paywalls or soft anti-adblock by default. Other ads, trackers and anti-adblock will be blocked by default.

Hanifalm:
So the solution is that I have to change my website domain again 😮‍💨

mapx-:
@Hanifalm explain your situation to your users. If they want to still enjoy your stuff they could donate a small amount otherwise it's better to close it.

uBlock-user:
> So the solution is that I have to change my website domain again 


That's not even a solution, you can fight us by changing domains, but it will not help with your financial situation.

ghajini:
the detection is no longer **dismissable**
image
so it will remain in main list https://github.com/uBlockOrigin/uAssets/commit/df63244fcc88c17bbe78cf59d839c7f660f8ecc8

ghajini:
Why is so


even with `dramafren.org##+js(noeval-if, rejectMethod)`
Try reloading page 2-3 times

uBlock-user:
`dramafren.org##+js(noeval-if, /rejectMethod|getComputedStyle/)` works fine, no need to for new filters.

ghajini:
> `dramafren.org##+js(noeval-if, /rejectMethod|getComputedStyle/)` works fine, no need to for new filters.

I got detected on refresh hence changed

PR #29935, opened 2025-09-05
heise.de, kleinanzeigen.de: switch from accepting cookies to reject

crazytombi:

kleinanzeigen.de
heise.de

### Describe the issue

at the moment cookies are accepted, would be better to reject

### Versions

- Browser/version: FF 142.0.1
- uBlock Origin version: 1.65.0

### Settings

- Add almost all lists (for regional only German list)

### Notes

Additionally unlocked the decline button on heise.de
kleinanzeigen.de does A/B testing with the consent banner, solution should cover all types at the moment.

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -4145,9 +4145,6 @@ winparts.eu,winparts.ie,winparts.se##+js(trusted-click-element, a.aanvaarden.gre
 sportano.*##+js(trusted-click-element, button.button--preferences, , 900)
 sportano.*##+js(trusted-click-element, button.button--confirm, , 1100)
 
-accept-wall
-kleinanzeigen.de##+js(trusted-click-element, button#gdpr-banner-accept, , 1000)
-
 ! reject
 crocs.*##+js(trusted-click-element, button.js-btn-reject-all, , 1300)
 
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
I don't understand. Please explain with screenshots which options you choose on each site.

crazytombi:
### heise.de

First click on "Einstellungen"

<details>

grafik

</details>

Second click on "Ablehnen", which is now visible 
Third click on "Ausgewähltem zustimmen", which is now enabled
Please feel free to find a better element/button selection 😅

<details>

grafik

</details>


### kleinanzeigen.de
**kleinanzeigen is doing some A/B testing. it might appear differently**

<details>

click on "Datenschutzeinstellungen"
grafik
then on "ablehnen und widersprechen" or simliar
grafik

</details>

stephenhawk8054:
For `heise.de`, on my side it's English, there's no `Einstellungen` button.

With the above procedures, youtube video cannot be played here:

```yml
https://www.heise.de/news/Ja-Linux-ist-wirklich-so-einfach-Win-10-Ende-10632924.html
```

It's because you need to choose `Agree` to display external contents here:

<details>

image

</details>

Please note that there's a reason we don't complicate the filters too much. We need to make sure that it works on both uBO and uBOL and on all platforms (desktop, chromium mobile, safari mobile...). Using `:nth-child` is not a stable solution if the website just simply changes the order of the elements.

crazytombi:
You usually should be able to allow external content when it is appearing. 
You are right, I was not thinking about other languages for this button, maybe it should be "button.sp_choice_type_12". 
Right the selection is not stable, I need to find a better way. I'm new to this and not so experienced. So this need some more refinement.

stephenhawk8054:
As I said, we won't complicate the filters more by clicking more items when there are external contents like this.

stephenhawk8054:
`kleinanzeigen.de` filters are not too complicated, we can adjust to your filters for `kleinanzeigen.de`.

crazytombi:
okay done
some of my other submissions are also not in the right place (reject all, only necessary) 
i'll fix this with another extra pull request later.

thx for your support

PR #30159, opened 2025-09-24
remove duplicate

xandaronox:
servers-mc.net

### Describe the issue
Since those filters added into adgaurd turkish filters no need to them

https://github.com/AdguardTeam/AdguardFilters/commit/be093fe096d70389de0226b1bbf9211ddc635ba5

### Versions

- Browser/version: Firefox 143
- uBlock Origin version: 1.66.4

### Settings

All default


gorhill:
The page at `https://www.servers-mc.net/` is in English, so these are needed in the base list.

xandaronox:
> The page at `https://www.servers-mc.net/` is in English, so these are needed in the base list.

 you can see both banners and servers are turkish. they probably made it according to browser's language

PR #30416, opened 2025-10-14
update falixnodes

xandaronox:
i disabled ublock quick filters then added/changed something to bypass their new anti adblock. Ive tested they working both android and pc on firefox

edit: ive removed first 2 filters due to no need them for falix's new anti adblock


stephenhawk8054:
@JobcenterTycoon Do the filters work on your side?

PR #30505, opened 2025-10-21
petbarn.com.au: block inline ads

opeik:

`https://www.petbarn.com.au/c/dogs/dog-supplies/travel-carriers/carriers`

### Describe the issue

There's ads inline within the item listings.

### Screenshot(s)

No filter:
Screenshot 2025-10-21 at 18-16-10 Dog Carriers Petbarn

Filter:
Screenshot 2025-10-21 at 18-20-36 Dog Carriers Petbarn

### Versions

- Browser/version: Firefox 143.0.4
- uBlock Origin version: 1.66.4

### Settings

- N/A

### Notes

I'm a beginner at creating filter rules. This seemed to do the trick though.

### DIFF
--- a/filters/filters.txt
+++ b/filters/filters.txt
@@ -11520,6 +11520,9 @@ thejobsmovie.com##+js(nostif, adsBlocked)
 @@||techdracula.com^$ghide
 techdracula.com##ins.adsbygoogle
 
+! Ads inline with search results.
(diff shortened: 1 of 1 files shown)


mapx-:
there are only self promotions - which we do not block / hide

opeik:
Ah, I see. My bad!

PR #30703, opened 2025-11-06
notebookcheck.net: Anti-Adblock

brian6932:

`https://www.notebookcheck.net/`

### Describe the issue

Anti-AdBlock popup + overflow hidden.

### DIFF
--- a/filters/quick-fixes.txt
+++ b/filters/quick-fixes.txt
@@ -394,3 +394,6 @@ tvtropes.org##+js(set, HTMLScriptElement.prototype.onerror, undefined)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/30676
 /stormcaster.js$badfilter
+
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Test again in a new browser profile with only uBO and its default settings.

brian6932:
Hmm it seems to be caused by this filter: https://www.eff.org/files/cookieblocklist.txt.

uBlock-user:
> Hmm it seems to be caused by this filter: https://www.eff.org/files/cookieblocklist.txt.

That list doesn't block anything, all filters there start with `@@`

stephenhawk8054:
The `@@||fundingchoicesmessages.google.com^$third-party` is what causes anti-adb.

uBlock-user:
@gorhill Should we add this list address to badlists ?

gorhill:
Yes, should be added the badlists, whatever its purpose is, it's not made to be used with uBO.

PR #30745, opened 2025-11-08
Address guardian consentwall (EU IP)

ryanbr:


u-RraaLL:
Does blocking the consent iframe cause any breakages?

I've blocked it on my side with these and haven't seen any breakages so far (both EU and US IP):

    ||sourcepoint.theguardian.com/index.html*theguardian.com%2Fconsent$subdocument,1p
    ||sourcepoint.theguardian.com/us_pm/*usnat_notice$subdocument,1p

PR #30794, opened 2025-11-13
Add a built in distractions filter. 

JuniorKlump:
Just a suggestion, if you're open to it, to add a built in filter list for blocking common distractions if you want to. I've only put youTube in the list, but there might be others that may be relevant, and maybe could be added, like maybe xkcd as well, or as such.


stephenhawk8054:
https://github.com/uBlockOrigin/uBlock-issues?tab=readme-ov-file#filter-list-requests

> New filter list requests are not permitted and will be declined and closed.

uBlock-user:
Filterlists are decided by the team internally, not by public recommendations.

Issue #30908, opened 2025-11-24
forum.dobreprogramy.pl: breakage


krystian3w:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://forum.dobreprogramy.pl/latest`

### Category

breakage

### Description

Some new filter can not be used with the mobile version of the forum on a real phone.

https://github.com/uBlockOrigin/uAssets/blob/19c00a881881525460b9da0a917d88261f25b5d3/filters/filters-2023.txt#L1536-L1537

[...]

Issue #30894, opened 2025-11-23
dobrepgoramy.pl: detection


Roki100:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [x] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
https://www.dobreprogramy.pl/altsendme-proste-i-bezpieczne-udostepnianie-plikow-bez-chmury,7222483456531392a
```

### Description

The video player is not playing/triggering "disable adblock" popup, once i even got an entire-site overlay about adblock detection

[...]

PR #30909, opened 2025-11-24
Update filters-2023.txt

krystian3w:
#30908 / #30894 

I haven't tested the second filter for e.g. mobile Edge[^1] if needs to be fixed in same way, I don't want to be overly cautious when it's harmless on Chrome.

[^1]: The only browser that makes sense to check on Android after Firefox.


krystian3w:
There is no longer need to check the subdomain, as the forum was closed on March 3, and since March 5, it has been functioning more or less like a proxy version with the `www `prefix, until they decide to implement a gentle 301 redirect or landing page (e.g. a farewell page or a 404 page for search engine bots.).

PR #31001, opened 2025-12-03
Add filter rules for shortening Amazon URLs

PurpleRabbit75:
This file contains filter rules to shorten URLs on Amazon sites by removing unnecessary URL parameters.

`[amazon.com](https://www.amazon.com/PlayStation-Portal-Remote-Player-Gaming-Console/dp/B0CJT5DJ16?sr=1-1-a989f4b3-2c98-4889-a8ed-d1e4cfb2317c&th=1)`

### Describe the issue

Links on amazon.com (and other sites owned by the company, whose urls follow the pattern amazon.*) are filled with a very wide variety of trackers, which both bloat the size of the links and violate users' privacy. This proposed (optional) filter list removes all url parameters which do not actively interrupt user experience. (Parameters like "?node" were left alone, because when removed the website redirects to a 404 page)

### Screenshot(s)

No visual issues.

### Versions

- All browsers
- uBlock Origin version: 1.67.0

### Settings

- I added a file containing rules for removing trackers from urls of the pattern amazon.*

### Notes

The ?node parameter should not be blocked.
The ?th parameter is blocked here, but automatically restores itself.


stephenhawk8054:
Report the issue to `AdGuard URL Tracking Protection` list: https://github.com/AdguardTeam/AdguardFilters/issues

The parameters removal section in uAssets list is only for the parameters that don't cause breakages in that list for some time.

---

Filters like `||amazon.*^$removeparam=_encoding` will cause issues for sure. Encoding here in amazon means the character encoding (for example UTF-8...) for the websites.

PR #31321, opened 2025-12-26
preisjaeger.at: typo

53871:

`https://www.preisjaeger.at/`

### Describe the issue

https://github.com/uBlockOrigin/uAssets/commit/203151a9d826a08ef808fcf0d64dfa89e771b864 changed the matching URL from `preisjaeger.at` to `reisjaeger.at`. The latter one does not exist and reintroduces the cookie notice for `preisjaeger.at`. I assume it was a typo and set it back to `preisjaeger.at`. Then the removal of the cookie notice works again.

### Versions

- Browser/version: FF 146.0.1

### Settings

- Enable "uBlock filters – Cookie Notices"


stephenhawk8054:
Thanks.

PR #31653, opened 2026-01-27
Add rules to block Maxroll adblocker popups

H4vC:
Added rules to block Maxroll adblocker popups and scripts.

`https://maxroll.gg/poe`

### Describe the issue

Adblock detection widget with a 15s timeout before the user can proceed.

### Screenshot(s)

image

### Versions

- Browser/version: Firefox 147

### Settings

None

### Notes

Investigated the adblocker detection mechanism and implemented an update to the annoyances filter that prevents it from triggering.


stephenhawk8054:
Do we have to sign in to reproduce the issue?

H4vC:
> Do we have to sign in to reproduce the issue?

it ought to repro without logging in.

stephenhawk8054:
I can't reproduce the issue.

PR #31660, opened 2026-01-28
Fix inconsistent youtube cookie consent: reject all cookies

ZashIn:
Reject all cookies from GDPR YouTube cookie consent, instead of currently sometimes accepting them.

See https://github.com/uBlockOrigin/uAssets/issues/30157#issuecomment-3809310153

- www.youtube.com: cookies accepted
- consent.youtube.com (e.g. redirected from any channel, like www.youtube.com/@youtube): cookies rejected

### Describe the issue
YouTube GDPR cookie consent is currently inconsistent in the Cookie Notices list:
https://github.com/uBlockOrigin/uAssets/blob/1c90eb0783c75b53569ea3dcf0e2887946f8f66d/filters/annoyances-cookies.txt#L737-L744
- youtube.com: accept all with second button (`ytm-button-renderer + ytm-button-renderer`)
- m.youtube.com: accept all with `ytm-button-renderer.eom-accept`
- consent.youtube.com: reject all with `button[jsname="tWT92d"]`

YouTube page reloads again after the click on accept all.

### Versions

- Browser/version: Firefox 147.0.1

### Settings

- [x] uBlock filters - Cookie Notices

## Notes about changes

I tested the proposed changes with a custom filter and default settings (Cookie Notices disabled) under Firefox Desktop and Android (Fennec).

Rejecting the cookies by default has the following benefits:
- shorter load time, since the YouTube page reloads when clicking "Accept all", but not on "Reject all"
- less tracking with less cookies and no search/play history ("Your YouTube history is off")
- following GDPR default behavior to reject non-essential cookies without explicit consent
- especially useful for users who delete cookies regularly, thus getting the cookie message with every new session

### Enable additional features like play history tracking
With "uBlock filters - Cookie Notices" enabled (now  rejecting cookies):
- Change the setting explicitly, e.g. from www.youtube.com:
  image
- do not delete cookies for YouTube (browser settings) to keep settings between sessions

Alternative: overwrite the new default behavior with the following custom rules:
- [x] Allow custom filters requiring trust
```
! YoutTube: Accept all cookies = second button
www.youtube.com##+js(trusted-click-element, .eom-buttons > ytd-button-renderer:nth-of-type(2) button[aria-label], , 1000)
m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-reject button, , 2000)
consent.youtube.com##+js(trusted-click-element, form[action] button[jsname="b3VHJd"])
consent.youtube.com##[data-p*="/consent.youtube.com"]:has(button[jsname="b3VHJd"])
```

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -734,8 +734,9 @@ nordax.com##+js(trusted-set-cookie, cookie_consent, %7B%22allowEssentials%22%3At
 ! https://www.reddit.com/r/uBlockOrigin/comments/1693mka/
 ! https://github.com/uBlockOrigin/uAssets/issues/20586#issuecomment-2357294152
 ! youtube.com##+js(trusted-set-cookie, SOCS, CAISNQgDEitib3FfaWRlbnRpdHlmcm9udGVuZHVpc2VydmVyXzIwMjMwODI5LjA3X3AxGgJlbiADGgYIgJnPpwY, , , reload, 1, domain, youtube.com)
-www.youtube.com##+js(trusted-click-element, ytd-button-renderer.ytd-consent-bump-v2-lightbox + ytd-button-renderer.ytd-consent-bump-v2-lightbox button[aria-label], , 1000)
-m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-accept button, , 2000)
+! Reject all = first button
+www.youtube.com##+js(trusted-click-element, .eom-buttons button[aria-label], , 1000)
+m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-reject button, , 2000)
 ! GDPR dialog when moving from Google to Youtube
 ! https://github.com/uBlockOrigin/uAssets/discussions/18598#discussioncomment-7362961
 consent.youtube.com##+js(trusted-click-element, form[action] button[jsname="tWT92d"])


stephenhawk8054:
We had many users opposing us for rejecting the cookies due to their personalization and watch history of the site not appearing.

There's no ways to please everyone. Your "benefits" are the "breakages" of other users.

ZashIn:
> We had many users opposing us for rejecting the cookies due to their personalization and watch history of the site not appearing.

I do not get the issue here: your cookie consent choice is either saved with your account settings (history appears after login) or with the (persistent) cookie after you "Update setting".

So the only time when your history does not appear is:
- in a private window/tab => there should be no history here
- if you are not logged in => log into your YT account
- the first time you visit YouTube after cookies are cleared => login / update settings (once)

Which case did users complain about exactly?

stephenhawk8054:
It's about the "related" and "personalized" videos not appearing based on their watch behaviors.

Yes, people want that. They think that unrelated videos appearing is "breakages". I can't find the reddit thread any more (it's years ago and searching on reddit is difficult).

For a long time, we don't want to address cookie issues because we know exactly these situations would happen: there are people who want the sites' personalized contents and there are people who don't care about it.

We are the ones who would receive all these contradicted complaints, it's really tiring.

ZashIn:
These issues might also be outdated. E.g. easylist notes that [comments break when rejecting cookies](https://github.com/easylist/easylist/tree/master/easylist_cookie#the-following-sites-have-no-current-workaround), which is not the case (anymore).

And again, people who want to see their "personalized" videos either keep their cookies or use an YT account, so this setting has to be changed only once. And even when you first reject cookies, they are accepted after you login in, as the preference is stored with your account settings (which I tested again).

But even if you want to uBO to consent to cookies, which I strongly oppose, this needs to be addressed, since they are currently rejected when you are redirected from consent.youtube.com.

So I think it is worth to revisit this again and hope you opt for rejecting cookies with this PR.

ZashIn:
Additionally, I see currently no way to overwrite the accept rules without disabling the entire cookie filter list.
This is not the case when the list rejects the cookies, as I mentioned above.

stephenhawk8054:
> And even when you first reject cookies, they are accepted after you login in, as the preference is stored with your account settings (which I tested again).

People want the related videos even when they are not logged in. That's the issue.

I personally don't want uBO to address these cookies issues any more. It would just create constant loop of people wanting their own preferences and forcing us to spoil their demands.

Other volunteers might have time and interests in this.

Good luck.

ZashIn:
I get your point. The ideal global solution would be a proper standardized protocol (like a special cookie) to tell the site the locally stored default cookie setting. That this was not the case from the start is a prime example of passing laws without thinking and caring about the proper practical implementation.

But until then, a separate extension might be better suited if one default choice cannot fit everyone, which the (uBlock) filter lists usually manage to find.

Anyways, to address this issue a consistent choice is still needed:
1. reject cookies: change `(www|m).youtube.com` filter, as this PR proposes
2. accept cookies: change the `consent.youtube.com` filter instead
3. do nothing (show cookie banner): remove all cookie related youtube filters

krystian3w:
The description of EasyList table probably applies to situations where we do not want to use scriptlets other than (trusted-)set-cookie/storage/click (e.g. long debugging how to stop pop-up using old arsenal like `aost` (`abort-on-stack-trace`) when uBo build 1.50.0 did not exist yet).

Then the cosmetic filter (even if it is procedural) implying ` { display: none !important } ` probably doesn't allow lazy loading of the comment section (you would have to reprogram the logic to force `foo:style(visibility: hidden ! important)` with an attempt to move the close button to under the comment loading error icon (`foo foo:style(visibility: visible ! important)` will restore the button without the rest of the pop-up; it's worse to move it perfectly to the comments using only CSS when `position: relative`/`absolute` fails).

Rather updating the md file when someone remembers about it or it lost its importance when `set-` was run other than "constant" (like cookie/storage) and clicking with uBo.

---

I don't know what old uBo you used, but the trusted scriptlet filters can now be deactivated https://github.com/gorhill/uBlock/commit/d2b8d990e6d970cf6f88e80b254682e05e9f44ad (we assume that almost no one uses older than stable 1.54.0+ (consciously/deliberately/intentionally); clicking was implemented a bit earlier than the exception fix):

```adb
www.youtube.com#@#+js(trusted-click-element, ytd-button-renderer.ytd-consent-bump-v2-lightbox + ytd-button-renderer.ytd-consent-bump-v2-lightbox button[aria-label], , 1000) 
m.youtube.com#@#+js(trusted-click-element, ytm-button-renderer.eom-accept button, , 2000) 
consent.youtube.com#@#+js(trusted-click-element, form[action] button[jsname="tWT92d"]) 
consent.youtube.com#@#[data-p*="/consent.youtube.com"]:has(button[jsname="tWT92d"])
```

However, this requires some effort from the user, e.g. studying the uBo Wiki and using uBo version 1.54.0+.

ZashIn:
Hiding/removing the consent, without clicking any button, has currently only one effect as far as I can see: video is paused / cannot be played. Everything else looks normal, including all comments being shown.
So while a click is still necessary to play videos, the comments are not effected by hiding the consent form.

----

### Custom filter to reject cookies instead
Thank you for the deactivating rules, I forgot that they work with +js, too. So rejecting instead of accepting cookies is possible with the current consent filers and the following custom filter:
 ```adb
! Youtube cookies: disable accept all
www.youtube.com#@#+js(trusted-click-element, ytd-button-renderer.ytd-consent-bump-v2-lightbox + ytd-button-renderer.ytd-consent-bump-v2-lightbox button[aria-label], , 1000)
m.youtube.com#@#+js(trusted-click-element, ytm-button-renderer.eom-accept button, , 2000)
consent.youtube.com#@#+js(trusted-click-element, form[action] button[jsname="b3VHJd"])
! Use reject all
www.youtube.com##+js(trusted-click-element, .eom-buttons button[aria-label], , 1000)
m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-reject button, , 2000)
consent.youtube.com##+js(trusted-click-element, form[action] button[jsname="tWT92d"])
```

ryanbr:
Is there a chance to  breaking youtube here,  whats the main motivation exactly? 

I believe in some regions rejecting YT consents went into a loop.

ZashIn:
Main motivation of this PR is to fix the inconsistent cookie choices for the the 3 YT [subdomains](#issue-3864181261).
My preference would clearly be to reject cookies and tracking by default (one of the main reasons to use uBlock).
But I understand why accepting cookies by default cause less "issues", since all YT features are then enabled by default, which most users would expect.

Even though you have to accept the downsides then as well, especially for private browsing:
- page reloads = slower
- more/possible unwanted tracking
- no easy way to change the cookie consent / tracking on YT without logging in (the "Update Settings" is removed after accepting)

For the latter, the [above custom filter](#issuecomment-3811562179) to reject cookies instead is of cause a workaround.

### Fix for accepting cookies by default
In this case, for `consent.youtube.com` changing the following button to "Accept all", `button[jsname="b3VHJd"]`, is still necessary:
https://github.com/uBlockOrigin/uAssets/blob/1c90eb0783c75b53569ea3dcf0e2887946f8f66d/filters/annoyances-cookies.txt#L741-L744

So if you prefer to accept YT cookies instead, you can merge PR #31673 instead and close this one.

ryanbr:
We're still blocking the tracking in the lists, this PR could go bad, give how popular youtube.com are willing to make possible breakage? Not worth it.

PR #31673, opened 2026-01-29
Fix inconsistent youtube cookie consent: accept all cookies

ZashIn:
Alternative solution to PR #31660: accepting youtube cookies instead of rejecting them to enable full functionality by default, including watch history.

### Affected URLs
| URL | previous cookie choice | new cookie choice |
| ----| -- | -- |
| www.youtube.com | accept | accept |
| consent.youtube.com * | reject | accept |

*: getting redirected from e.g. any channel URL, like [www.youtube.com/@youtube](http://www.youtube.com/@youtube) (with cleared cookies)

#### Overview of youtube cookie behavior
| cookies              | accept / save                   | reject / delete             |
| -------------------- | ------------------------------- | --------------------------- |
| YT history           | on                              | off                         |
| personalized content | yes                             | no                          |
| banner is shown      | once (choice saved with cookie) | after every cookie deletion |
| page reloads         | at least once                   | no                          |
| change choice without deleting cookies | only when logged in (account settings) | use "Update Settings" from message "Your YouTube history is off" (e.g. on www.youtube.com) |

Choosing "Accept all" cookies to enable full youtube functionality, which most users would expect.

### How to overwrite the cookie consent choice
To reject all youtube cookies instead, a user can add the following custom filter, with
- [x] Allow custom filters requiring trust
```adb
! Youtube cookies: disable accept all filters
www.youtube.com#@#+js(trusted-click-element, ytd-button-renderer.ytd-consent-bump-v2-lightbox + ytd-button-renderer.ytd-consent-bump-v2-lightbox button[aria-label], , 1000)
m.youtube.com#@#+js(trusted-click-element, ytm-button-renderer.eom-accept button, , 2000)
consent.youtube.com#@#+js(trusted-click-element, form[action] button[jsname="b3VHJd"])
! reject all
www.youtube.com##+js(trusted-click-element, .eom-buttons button[aria-label], , 1000)
m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-reject button, , 2000)
consent.youtube.com##+js(trusted-click-element, form[action] button[jsname="tWT92d"])
```

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -734,14 +734,15 @@ nordax.com##+js(trusted-set-cookie, cookie_consent, %7B%22allowEssentials%22%3At
 ! https://www.reddit.com/r/uBlockOrigin/comments/1693mka/
 ! https://github.com/uBlockOrigin/uAssets/issues/20586#issuecomment-2357294152
 ! youtube.com##+js(trusted-set-cookie, SOCS, CAISNQgDEitib3FfaWRlbnRpdHlmcm9udGVuZHVpc2VydmVyXzIwMjMwODI5LjA3X3AxGgJlbiADGgYIgJnPpwY, , , reload, 1, domain, youtube.com)
+! Accept all = 2nd button
 www.youtube.com##+js(trusted-click-element, ytd-button-renderer.ytd-consent-bump-v2-lightbox + ytd-button-renderer.ytd-consent-bump-v2-lightbox button[aria-label], , 1000)
 m.youtube.com##+js(trusted-click-element, ytm-button-renderer.eom-accept button, , 2000)
 ! GDPR dialog when moving from Google to Youtube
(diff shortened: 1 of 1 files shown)


ZashIn:
Why was this PR also closed, too? I understand the reason for closing #31660, you do not want to break anything for existing users by rejecting the cookie.

But this PR fixes a bug without changing the current default state for users visiting youtube with Cookie Notices filters enabled: accepting cookies.
You can test the broken case like this:
1. with default uBlock settings, enable the cookie filter list "uBlock filters – Cookie Notices" or one of / all of the Cookie Notices parent groups.
2. open a private window
3. open www.youtube.com/@youtube or any other channel (either directly or from a search engine)
4. switch to the start page www.youtube.com

Result: cookie consent was rejected, so history is disabled!
image

In contrast to this, doing steps 1. and 2., but opening www.youtube.com directly results in history being enabled.

Issue #13017, opened 2022-04-29
Paste into customer number and password fields not allowed.


xpufx:
### Prerequisites

- [X] This is not a [support issue or a question](https://old.reddit.com/r/uBlockOrigin/)
- [X] I performed a [cursory search of the issue tracker](https://github.com/uBlockOrigin/uAssets/issues) to avoid opening a duplicate issue
- [X] I am running the latest version of uBlock Origin

### I tried to reproduce the issue when...

- [X] uBO is the only extension
- [X] uBO with default lists/settings
- [X] using a new, unmodified browser profile

### URL(s) where the issue occurs

`https://esube.albaraka.com.tr/LoginIB.aspx?sID=-1&type=corporate`

### Describe the issue

Site won't allow pasting into the Customer Number and Password fields.

### uBlock Origin version

 1.42.4

### Browser name and version

Firefox 99.0.1 

### Settings

Note: I didn't actually do the steps in the "I tried to reproduce the issue when..." as they are not relevant.

### Notes

None

gorhill:
The disabling of copy/paste of password is a security choice by the site, that choice is not something to be second-guessed by uBO's lists, as opposed to a website disabling selecting/copying text content for no good reasons.

zeref-dragneel:
You can use this extension.
[Don't Fuck With Paste](https://github.com/aaronraimist/DontFuckWithPaste)

PR #31883, opened 2026-02-16
Allow pasting 2FA code for TD Canada

xethlyx:

`https://authentication.td.com/uap-ui/?consumer=easyweb&locale=en_CA#/uap/login`

### Describe the issue

The "Enter Security Code" field does not allow pasting or selection (though this seems to be broken on Firefox so only pasting is disallowed). This screen is only triggered when signing in from an unknown IP address from the first time.

Pasting is denied via attributes on the element with id `code`, through this code:
`onpaste="return false;"`, `onselectstart="return false"`, etc.

### Screenshot(s)

Pasted image

I'm assuming I don't need the rest of the template since this is a pull request!

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -7935,3 +7935,6 @@ thestudentroom.co.uk##+js(set-cookie, _tsr_pc, 0)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/31841
 ft.com##+js(aeld, copy, clipboardData)
+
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
For security-related copy paste nuisances, we won't address them: https://github.com/uBlockOrigin/uAssets/issues/13017#issuecomment-1113341426

xethlyx:
I see. Thanks for the clarification

PR #32020, opened 2026-02-28
Update ubo-link-shorteners.txt with new rules

BlazeFTL:
``


ghajini:
why not use `exeygo.com##+js(prevent-fetch, method:HEAD)` instead of trusted scriptlet `##+js(trusted-create-html, head,  false }; }'>)`

BlazeFTL:
i didnt check the site first thing i tried was `##+js(trusted-create-html, head,  false }; }'>)`  which worked...

if prevent-fetch seems better then update

Issue #22198, opened 2024-01-25
open.spotify.com: ads / breakages


boromirmonk:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://github.com/uBlockOrigin/uAssets/discussions/17361) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.
### URL address of the web page

`https://open.spotify.com/`

### Category

ads

### Description

Today, the site starting showing ads while having uBlock Origin activated. 

### Other extensions used

None.

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.55.0
Chromium: 121
filterset (summary):
 network: 118987
 cosmetic: 51582
 scriptlet: 20570
[...]

PR #32124, opened 2026-03-08
add spotify to issue template

lunaflight:
### Describe the issue
there is a megathread for https://github.com/uBlockOrigin/uAssets/issues/22198 but the issue template doesn't include it. all other megathreads are in the issue template. would be nice to also list it here to try to reduce duplicates

feel free to reject if spotify: ads / breakages / nuisances != ALL spotify issues


lunaflight:
ugh, did i do something wrong? Every newly opened issue with this template is popping up as "mentioned" in the spotify issue megathread

stephenhawk8054:
You need to put `www.` as a trick to avoid the issue thread being mentioned in other newer issue threads.

PR #32192, opened 2026-03-13

BlazeFTL:


BlazeFTL:
@mapx- could we add this in shortner thread ?
```
### ⚙️ Recommended Environment
For the filters to work as intended (and to be eligible for support), your setup **must** match the following:

> [!CAUTION]
> **Recommended:** Use only with uBlock Origin's default filters. Adding third-party lists may cause conflicts or errors.

> [!TIP]
> - Keep filters updated regularly

* **Browser:** [Firefox](https://www.mozilla.org/firefox/) is highly recommended.
* **Extension:** **uBlock Origin (MV2)**. 
    * *Note: There is no support for uBlock Origin Lite (MV3).*
* **Clean Browser:** No extra extensions that perform blocking or page modification. **uBlock Origin is enough.**
* **No External Blocking:** You must not use any **DNS** or **3rd Party Apps** that block ads (e.g., NextDNS, AdGuard App, Pi-hole).

### 🛠️ uBlock Origin Configuration
Please use the **Default Filters**, plus these specific additions if needed:
* ✅ **EasyList + AdGuard Annoyances** (Cookie notices, Social widgets, and Annoyances *only*)**
```

Issue #32015, opened 2026-02-27
discord.com: nuisance


lunaflight:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
discord.com
```

## reproduce
go to any discord channel
paste `https://open.spotify.com/track/0zs01KE7vNOQePFZWmnQls`
you should get an embed; try to copy the title or artists,
copy paste blocked

## on duped title 

[...]

PR #32199, opened 2026-03-14
update #32015

lunaflight:

discord.com

### Notes

all outlined in https://github.com/uBlockOrigin/uAssets/issues/32015

### DIFF
--- a/filters/annoyances-others.txt
+++ b/filters/annoyances-others.txt
@@ -168,7 +168,8 @@ padlet.com##.select-none.surface-post:style(-webkit-touch-callout: default !impo
 padlet.com##.wish-wrapper > div[draggable="true"]:remove-attr(draggable)
 
 ! https://github.com/uBlockOrigin/uAssets/issues/32015
-discord.com##*:style(-webkit-touch-callout: default !important; -webkit-user-select: text !important; -khtml-user-select: text !important; -moz-user-select: text !important; -ms-user-select: text !important; user-select: text !important; -webkit-tap-highlight-color: revert !important;)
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Styling all elements is really risky (and I'm unsure about the performance for other users). I personally don't really want to add it to the stock lists.

Any users who want to resolve these annoyances can add the filter themselves.

lunaflight:
Would it be better if we enumerated and applied this to the important elements we deem fit, such as just the sidebar and the status pop-ups when you click on someone's profile?

stephenhawk8054:
I tried, but there were too many elements in the DOM to check. I didn't have time to look at the issue since then.

lunaflight:
I agree that we should avoid styling all elements.

On a preliminary skim, I feel like targeting just the following 2 could be sufficient to mitigate the annoyance, since these encapsulate game statuses / song statuses that are potentially dense in information and worth to copy & paste. I do admit maybe I'm not the best at picking selectors though.

```html
<ul class="cards__[NUMBERS]">
<ul class="cardsList__[NUMBERS]">
```

If any other users find more annoyances that they want to be addressed, they can leave a comment in the thread in the future maybe, but this could just be fine for now.

stephenhawk8054:
I don't see any elements with `cards` in class names on the link `https://open.spotify.com/album/2wQqd6G4SLmevXyGwoQbuZ`:

<details>

image

</details>

lunaflight:
<details>
image
The above contains `cardsList__`. 

Upon clicking on any of the cards, this pops up:
image
The above contains `cards__`.
</details>

to be clear, i'm referring to the now-playing statuses on `discord.com`. the reproduction steps are as outlined here. https://github.com/uBlockOrigin/uAssets/issues/32015#issuecomment-4025713486

PR #32247, opened 2026-03-18
Suppressing cookie popup on la7.it

griffinallen:
URL(s) where the issue occurs
`la7.it`

Describe the issue

Versions

Fix
Added trusted click element to suppress the notification

Notes
Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2230


stephenhawk8054:
Do these break anything?

```adb
||la7.it/la7_iabtcf2/cmp.js
||privacymanager.io^
```

griffinallen:
> Do these break anything?
> 
> ```
> ||la7.it/la7_iabtcf2/cmp.js
> ||privacymanager.io^
> ```

I have just tested that. Seems to work as expected, no breakage.

PR #32262, opened 2026-03-19
Rejecting cookies from theweathernetwork

griffinallen:
URL(s) where the issue occurs
`https://www.theweathernetwork.com/` -- when accessing the site from the UK

Describe the issue

Versions

Fix
Added trusted click element to suppress the notification

Notes
Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2041


mapx-:
no cookies using ELC on my side

griffinallen:
> no cookies using ELC on my side

Are you browsing from within the UK? Cleared cookies and local storage?

PR #32266, opened 2026-03-19
Rejecting unnecessary cookies for attheraces

griffinallen:
URL(s) where the issue occurs
attheraces.com

Describe the issue

Versions

Fix
Added trusted click element to suppress the cookie popup.

Notes
Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2001


stephenhawk8054:
Does this work?

```adb
||cdn-cookieyes.com^$domain=attheraces.com
```

PR #32267, opened 2026-03-19
NSFW - Rejecting unnecessary cookies on faphouse.tv

griffinallen:
URL(s) where the issue occurs
`faphouse.tv` - NSFW - reproducible from the UK

Describe the issue

Versions

Fix
Added trusted set cookie to suppress the cookie popup.

Notes
Related to this issue, however, this is not the full fix. I will also need to make a commit on ELC to hide the cookie banner that appears after rejecting cookies.
https://github.com/brave-experiments/cookiecrumbler-issues/issues/2095


stephenhawk8054:
```adb
faphouse.tv###cookie-wall-short
```

For simple cookie elements, no need to use scriptlets.

PR #32268, opened 2026-03-19
Rejecting cookies from humanartistrycampaign.com

griffinallen:
URL(s) where the issue occurs
`https://www.humanartistrycampaign.com/`

Describe the issue

Versions

Fix
Added trusted click element to suppress the cookie popup.


stephenhawk8054:
```adb
humanartistrycampaign.com##.consent-banner-root
```

griffinallen:
Could you explain this for me? 

My thought process behind this was that if the page shows an option to select between multiple cookie preferences, I should be making a change to uBO to essentially "save" the appropriate cookie selection for the page. However, a lot of my PRs are closed with suggestions for a cosmetic filter instead. If I can understand why, I can avoid making extra work for you guys.

mapx-:
In this case the essential cookies cannot be disabled, so using a cosmetic filter you'll keep the essential part and don't enable the others.

stephenhawk8054:
If the issue can be solved by cosmetic or network filters without breakages (or the elements don't flash when being hidden), we'll prefer that solution.

PR #32273, opened 2026-03-20
Only accepting necessary cookies for PostNL

griffinallen:
URL(s) where the issue occurs
`postnl.nl`

Describe the issue
Cookie popup is loaded on page load. It needs to be suppressed. Using a cosmetic filter will not work because the page will not scroll until the cookie has been set correctly.

Versions

Fix
Added trusted set cookie to satisfy this cookie popup. Note, the version of the cookie just needs to be greater than or equal to the current cookie wall version (3). I set this to 99 so that we will not have to adjust this in the near future.


ryanbr:
Seems even with accepting the consent, the will still show

ryanbr:
```
postnl.nl,postnl.be##pnl-cookie-wall-widget:remove()
postnl.nl,postnl.be##body:style(position: static !important; overflow: auto !important; width: auto !important;)
```

PR #32335, opened 2026-03-26
weddingbee.com cookies

griffinallen:

`www.weddingbee.com/find`

### Describe the issue

Users are reporting that searching doesn't work on this page. This is due to functional cookies being disabled by default on Brave. 

I did some digging through the source of this site. What is actually happening here is that the site looks at the navigator.globalPrivacyControl value and if it is true (which is the value when shields are on), sets groups C0002, C0003, c0004 to 0. If this value is false, they all get set to 1. This means functional cookies are being blocked when we haven't even seen a cookie consent.

### Versions

- Browser/version: Brave 1.88.136 (Official Build)

### Settings

Added a trusted-set-cookie to accept both the necessary and functional cookies for this site.


stephenhawk8054:
~~Which country are you when accessing the site?~~

I can reproduce with VPN to UK.

PR #32381, opened 2026-03-30
hiding neoxsoftworks.eu ad popup

griffinallen:

`https://neoxsoftworks.eu/`

### Describe the issue

When using the site with an adblocker, the site detects the adblocker and prompts the user to disable it.

### Screenshot(s)
Screenshot 2026-03-30 at 12 02 10 PM
Screenshot 2026-03-30 at 11 49 41 AM

### Versions

- Browser/version: Brave 1.88.136

### Settings

Added a session storage variable to this site that sets the adblockDismissed value to 1, so that it bypasses the adblock check.


stephenhawk8054:
For value `1`, you don't need `trusted-`.

PR #32401, opened 2026-03-31
Hiding popup on pinterest.com

griffinallen:

`pinterest.com`

### Describe the issue

pinterest.com has a cookie popup that appears to users. This popup needs to be suppressed.

### Screenshots
image

### Versions

- Browser/version: Brave 1.88.136

### Settings

Set a trusted click element on the "Reject All" setting in the cookie popup

Fixes https://github.com/brave-experiments/cookiecrumbler-issues/issues/2048

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -5468,3 +5468,7 @@ elysee.fr##+js(trusted-set-cookie, tarteaucitron, !eulerian-analytics=false!inst
 
 ! https://github.com/uBlockOrigin/uAssets/pull/32335 - Functional
 weddingbee.com##+js(trusted-set-cookie, OptanonConsent, groups=C0001%3A1%2CC0002%3A0%2CC0003%3A1%2CC0004%3A0, , , domain, weddingbee.com)
+
+! Reject
+! https://github.com/brave-experiments/cookiecrumbler-issues/issues/2048
+pinterest.com##+js(trusted-click-element, .unzjra button:first-child)


krystian3w:
What damage do you notice when using[^1]:
```adb
pinterest.*##body > div[class]:has(span > a[href*="/policy/cookie"])
```

Tags/profiles/catalogues may need these addition (The scrolling behaviour there is different compare to homepage):
```adb
pinterest.*##html:not([style]) body[style]:style(overflow-y: auto !important; margin-inline-end: 0 !important)
```

When regional suffixes only redirect, you can skip them and simplify them to `com`. 

----

[^1]: I only notice the ones that appear when the login pop-up is disabled incorrectly (if it’s leaking on my end, it means that the uBO Annoyances filters are out of date or aren’t keeping up with the site and are treating it as user-unfriendly).

mapx-:
I get no cookie popup

krystian3w:
In Poland, this GDPR cookie pop-up is already appearing.

I doubt that ELC has gone to the trouble of implementing a classic filter (something like my CSS4, or rigid, unreadable classes `.unzjra` like those in React) or effectively adapting to set-cookie/local-storage without trusted mode.


I would treat the issue of updating the ‘log in’ page as something difficult to reproduce, e.g. 30 clears and refreshes of profiles or ‘tags’ (so it’s more likely that the website’s mechanism went haywire that day).

griffinallen:
> What damage do you notice when using[1](#user-content-fn-1-1a86ddf719696ecc5cf2dfcb4f37a158):
> 
> ```
> pinterest.*##body > div[class]:has(span > a[href*="/policy/cookie"])
> ```
> 
> Tags/profiles/catalogues may need these addition (The scrolling behaviour there is different compare to homepage):
> 
> ```
> pinterest.*##html:not([style]) body[style]:style(overflow-y: auto !important; margin-inline-end: 0 !important)
> ```
> 
> When regional suffixes only redirect, you can skip them and simplify them to `com`.
> 
> ## Footnotes
> 1. I only notice the ones that appear when the login pop-up is disabled incorrectly (if it’s leaking on my end, it means that the uBO Annoyances filters are out of date or aren’t keeping up with the site and are treating it as user-unfriendly). [↩](#user-content-fnref-1-1a86ddf719696ecc5cf2dfcb4f37a158)

I have tested this and it seems to work as expected. I have made a PR in Easylist. Thanks

krystian3w:
The only problem is if the pull request gets stuck there, or if they insist on using that unreadable class instead of CSS4 (I usually set the number of divs to scan to 4 for users who aren't logged in).

The second issue is that it ended up in the wrong ELC file; this will corrupt profiles/tags/albums where scrolling is disabled with ABP/AdBlock/free AdGuard for iOS.

It would be better to move this to a [‘supplement’](https://github.com/griffinallen/easylist/blob/pinterest.com_cookies/easylist_cookie/easylist_cookie_specific_uBO.txt) file until a major release of ABP supports the CSS syntax `domain##selector { foo: bar }` (the syntax is still blocked by some flags, presumably compilation flags).

griffinallen:
> The only problem is if the pull request gets stuck there, or if they insist on using that unreadable class instead of CSS4 (I usually set the number of divs to scan to 4 for users who aren't logged in).
> 
> The second issue is that it ended up in the wrong ELC file; this will corrupt profiles/tags/albums where scrolling is disabled with ABP/AdBlock/free AdGuard for iOS.
> 
> It would be better to move this to a [‘supplement’](https://github.com/griffinallen/easylist/blob/pinterest.com_cookies/easylist_cookie/easylist_cookie_specific_uBO.txt) file until a major release of ABP supports the CSS syntax `domain##selector { foo: bar }` (the syntax is still blocked by some flags, presumably compilation flags).

Thanks for pointing this out. I was not aware of this limitation with ABP. I have moved it to the file you suggested.

PR #32408, opened 2026-04-01
Hiding cookie notice on kroger.com

griffinallen:
URL(s) where the issue occurs
`kroger.com`

Describe the issue
kroger.com has a cookie popup that appears to users. This popup needs to be suppressed.

Screenshots

image

Versions
Browser/version: Brave 1.88.136

Settings
Set a trusted local storage item to bypass the cookie popup


stephenhawk8054:
For value `true`, you don't need trusted scriptlets, and it can be added to EL Cookie: https://github.com/easylist/easylist/commit/679d78f70c2bd8a69ca8edc1182c272ed075c48e

stephenhawk8054:
Here are the values that don't need trusted scriptlets (it can be used for cookies and other storage items): https://github.com/gorhill/uBlock/wiki/Resources-Library#set-cookiejs-

PR #32447, opened 2026-04-05
feat(filters/badware): only allow aliucord.com

rushiiMachine:
This targets aliucord.org which is badware

`https://aliucord.org`

### Describe the issue

https://aliucord.com is the only official website for https://github.com/Aliucord (of which I am a member), and `aliucord.org` is an unauthorized impersonation that could distribute malware in the future.

### Notes

Currently, only `aliucord.org` is known of, but my changes block all `aliucord.*` domains that aren't aliucord.com


mapx-:
> could distribute malware in the future

- scanned using virustotal, urlscan => no threat
- any site in this world could be hacked in the future

Issue #32206, opened 2026-03-15
audiveris.com: badware


watappo:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184), [_Spotify_](https://www.github.com/uBlockOrigin/uAssets/issues/22198) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
https://audiveris.com/
```

### Category

badware

### Description

This website is NOT an official website of Audiveris (Optical Music Recognition Software) and may cause harm to users.
Here is a statement from the original repository (URLs are sanitized):

> Beware of site audiveris[.]com!
> 
[...]

PR #32449, opened 2026-04-05
feat(filters/badware): unofficial vesktop sites

rushiiMachine:

- `vesktop.org`
- `vesktop.softonic.com`
- `vesktop.softonic.nl`
- `vesktop.softonic.pl`
- `vesktop.en.softonic.com`

### Describe the issue

The only official download site for Vesktop (https://github.com/Vencord/Vesktop) is at https://vesktop.dev.

- `vesktop.org`, impersonates the real website
- softonic bundles PUPs with downloads

### DIFF
--- a/filters/badware.txt
+++ b/filters/badware.txt
@@ -6358,3 +6358,8 @@ ublockorigin.app###main::before:style(content: 'This is not the official uBlock 
 
 ! https://censys.com/blog/technique-based-approach-hunting-web-delivered-malware/
 ||orcanmedikal.com.tr^$doc
+
+! Vesktop - Official site: vesktop.dev
+! https://github.com/uBlockOrigin/uAssets/issues/32449
+||vesktop.$all,domain=~vesktop.dev
+||vesktop$all,domain=softonic.*


mapx-:
I get no threat using urlscan / virustotal
I see no analysis / prove about malware on those sites

magneticflux-:
I was about to file an issue about these same sites before I searched and found this.

There is a definite risk because the site instructs users to download and install a file they control. `vesktop.org` might send a copy of the official install if you check right now (or when malware scanners look), but there's nothing stopping it from silently switching out downloads for malware randomly or based on User-Agent or IP address. In contrast the official Vesktop downloads are all hosted on their official GitHub repository.

This is the exact same situation (and probably the same bad actor since **the website is identical**) as https://github.com/uBlockOrigin/uAssets/issues/32206 which @stephenhawk8054 contributed: a popular software download site gets impersonated and serves AI-generated blogspam and ads to boost its standing in Google search results, and its creators state that it is unofficial ([from 3 months ago](https://www.reddit.com/r/linuxquestions/comments/1q5qwlk/vesktop_official_site/)).

stephenhawk8054:
I added the filter because there's a statement from the original github itself.

DON'T use the argument that a site is added, so is other sites. We have many complaints whenever we add a site with weak justifying and the site owners come here. Any adding of those sites is a risk of receiving more of these complaints and we won't do these for all cases. Will you ever defend for us or you just go after reporting and let us defend these decisions ourselves when we receive the complaints?

Anyone who wants to use these kinds of arguments are free to contribute elsewhere.

magneticflux-:
I apologize, I think there was some miscommunication at the start, I believe @rushiiMachine is directly(?) communicating with the developers of Vesktop (they sponsor rushii, work on the same projects etc.). I was also specifically referring to the `vesktop.org` domain, I don't know anything about Softonic but it seems like it should be handled separately.

I don't know if they'd want to put a massive warning on top of their readme about the imposter website like Audiveris did (to avoid giving it any extra attention), but would an official statement from one of the developers saying that they believe it is an automated phishing website suffice?

I only bring up the Audiveris site to compare because it has literally the _exact_ same website pages with slightly changed text and images to fit. It seems like this is an automated attack on tons of open-source projects that don't own every com/org/net/etc permutation of their URL.

stephenhawk8054:
> I believe @rushiiMachine is directly(?) communicating with the developers of Vesktop (they sponsor rushii, work on the same projects etc.).

The thing is, how can the public know where those "official" communications are? We can't just say: "Complain this to this person because this person has communicated with other person".

PLEASE:
--

When you report some sites as badware, please always remind how you can answer these questions on behalf of uBO volunteers, because **WE** are the ones who will have to answer the complaints from anybody else after that:

1. Where are the reliable 3rd-party analysis/reports of those sites' malicious actions?
2. Where are the confirmations that those sites are malicious from original authors?

---

> I only bring up the Audiveris site to compare because it has literally the exact same website pages with slightly changed text and images to fit.

DON'T ever compare these decisions. We are volunteers. We have little free time to resolve issues from different websites everyday. We don't have time to be stressed over these comparisons and be obliged to answer these.

rushiiMachine:
> I believe @rushiiMachine is directly(?) communicating with the developers of Vesktop (they sponsor rushii, work on the same projects etc.)

Yes, I asked Vee (original author of the Vencord/Vesktop projects) before opening this pr. @Vendicated 

> Where are the reliable 3rd-party analysis/reports of those sites' malicious actions?

A negative result from an AV vendor means little to nothing, and it certainly does not mean that the site is not malicious. `vencord.org` seeks to deceive users by pretending to be the real Vesktop website (this is a very common problem for Discord client mods for some reason), and serves their own executable downloads. They could easily swap out a safe binary for a malicious one at any time, or only for a subset of users by geographical location, etc. Just by existing and seeking to deceive users, vesktop.org is badware

The only official and safe downloads that exist are hosted on Github, to which the real Vesktop website, vesktop.dev, redirects to.

stephenhawk8054:
Yes, I know. You still don't get it? 

The thing is, we need official/3rd-party evidences to present for any questions come later.

> They could easily swap out a safe binary for a malicious one at any time, or only for a subset of users by geographical location, etc. Just by existing and seeking to deceive users, vesktop.org is badware

We can't use just `could` words like this as answer to any of the above questions you know?

If others come here to complain, we could only ping you guys and hope you guys answer while they are pressing us?

---

> A negative result from an AV vendor means little to nothing, and it certainly does not mean that the site is not malicious. 

And we certainly cannot use just words to justify that it's malicious to block. We need evidences of `malicious` here.

---

> would an official statement from one of the developers saying that they believe it is an automated phishing website suffice?

Statement from where? This thread? You guys don't want to put the statements on the official projects but want to push all responsibilites to us instead?

mapx-:
> A negative result from an AV vendor means little to nothing, and it certainly does not mean that the site is not malicious.

yes: it is a good idea to block the entire internet (tomorrow any site could become harmful)

rushiiMachine:
> We need evidences of malicious here.

Is impersonation and redirecting traffic not enough of a reason to call it malicious behavior? I would consider it so

stephenhawk8054:
> impersonation

No, there are thousands of cloning project on github, hosted on github pages. We can't block all of them because of that. Should we block `https://ublockorigin.com` as well while we have 0 evidences of malicious activities from it?

> redirecting traffic

Redirecting from where to where? From beginning to here, no one gives any technical analysis of these sites' activities (literally what we are asking for from beginning).

PR #32576, opened 2026-04-16
reject all additional cookies on arena.ai

rhjdvsgsgks:

`arena.ai`

### Describe the issue

it have a cookie popup. and 4 switch for additional cookie turned on by default

### Versions

- Browser/version: firefox 148
- uBlock Origin version: 1.70.0

### Settings

- none


stephenhawk8054:
Which country are you when accessing the site?

mapx-:
I added the filter used by AG

rhjdvsgsgks:
> I added the filter used by AG

im not sure why `functionality` and  `socialMedia` turned on by default in your rule, they wont affect any feature even when turned off. plus `lastUpdated` have a incorrect format. plus without `reload, 1` flag, popup will still shown as normal.

stephenhawk8054:
The filter is `trusted-set-cookie-reload` which is an alias for `reload, 1`

Issue #32641, opened 2026-04-22
ah.nl: privacy


flexagoon:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184), [_Spotify_](https://www.github.com/uBlockOrigin/uAssets/issues/22198) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
https://zsp1.ah.nl/raylene//sl403/IIa4e3fc29-efad-4066-9ec7-243effd22074/ad2053647-19570/URhttps://www.ah.nl/shop/nieuw
```

### Description

[...]

PR #32647, opened 2026-04-23
Block external ads on ah.nl

flexagoon:

https://ah.nl

### Describe the issue

Most banners on ah.nl are self-promotion, which are out of scope per EasyList guidelines. However, some of them are actually ads for external website (see the screenshot).

Since they link to a tracking URL (#32641) instead of to the page directly, an extended `:not(:matches-attr(...))` rule has to be used to find the ones that link outside of ah.nl

### Screenshot(s)

The banner on the left is an advertisement:

image

With the rule applied:

image

### Versions

- Browser/version: Zen Browser 1.19.8b (Firefox 149.0.2)
- uBlock Origin version: 1.70.0


stephenhawk8054:
Did you enable `Adguard Dutch` list?

Anyways, this should be fixed in regional list, not here.

flexagoon:
@stephenhawk8054 I see, thanks

PR #32704, opened 2026-04-26
Combine anikai.to/animekai. rules

ryanbr:
Merge rules from https://animekai.pw/ same site, different mirrors.


uBlock-user:
bring in the rest of the rules, they're run by the same team

PR #33084, opened 2026-05-28
Hiding cookie notice on theguardian.com

griffinallen:

`theguardian.com`

### Describe the issue

There is a cookie notice on this page that should be hidden. When you reject this cookie notice, there is also a "rejection hurts... please enable cookies" message that appears. This should also be hidden.

### Screenshot(s)

image

image

### Versions

- Browser/version: Brave 1.90.124

### Notes

This is in response to numerous user reports on Brave. I am open to applying a different blocking approach if there is a more appropriate solution here.

### DIFF
--- a/filters/annoyances-cookies.txt
+++ b/filters/annoyances-cookies.txt
@@ -5583,3 +5583,7 @@ eventim.ro##+js(trusted-click-element, #cmpwrapper >>> .cmpboxbtnyes, , 1000)
 
 ! https://github.com/uBlockOrigin/uAssets/pull/33067
 lequipe.fr##+js(trusted-click-element, .CmpUnsubscriber .Cmp__action--yes, , 1000)
+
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Which country are you when accessing the site?

griffinallen:
> Which country are you when accessing the site?

Canada

stephenhawk8054:
I tested with VPN to Canada but I can't reproduce. The cookie banner is already addressed in the list

https://github.com/uBlockOrigin/uAssets/blob/37b42ccb49c263f6665f04fd6221fcad2b134acf/filters/annoyances-cookies.txt#L4983-L4989

Did you test with uBO and `EasyList/uBO – Cookie Notices` turned on? One of your filters is already conflicting with another filter setting `consentUUID` value.

griffinallen:
Yes, I was originally testing with the Cookie Notices turned on. I can no longer reproduce. I have also reached out to the issue reporter to see if they can reproduce still. This may no longer be an issue

PR #33150, opened 2026-06-02
Use essential cookies on filmpalast.net

griffinallen:

`filmpalast.net`

### Describe the issue

This site has a cookie notice that cannot be hidden cosmetically because there are scroll blockers and overlays that are present on the page. This change will use a trusted-set-cookie to accept essential cookies for the site.

### Screenshot(s)

image

### Versions

- Browser/version: Brave 1.90.128


stephenhawk8054:
Does this work?

```adb
||filmpalast.net/wp-content/*/borlabs-cookie/
```

griffinallen:
This works, thank you. PR to Easylist - https://github.com/easylist/easylist/pull/24350

Issue #22295, opened 2024-02-01
Schibsted sites - Blocked videos [GDPR]


THEtomaso:
### Prerequisites

- [X] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/20586), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367) or [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184) report. _These sites MUST be reported by clicking their **respective links**_.
- [X] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [X] I verified that this issue is not a duplicate. (Search [here](https://github.com/uBlockOrigin/uAssets/issues) to find out.)
- [X] I did not remove any of the [default filter lists](https://github.com/uBlockOrigin/uAssets/assets/66517106/6e8d024a-5869-47cf-b142-b2357eb6c7d0), or I have verified that the issue was not caused by removing any of the default lists.
- [X] I did not enable additional filter lists, or I have verified that the issue still occurs without enabling additional filter lists.
- [X] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [X] I am not using uBlock Origin (uBO) along with other content blocker extensions.
- [X] I have verified that the web browser's built-in blocker or [DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking) (standalone or through a VPN) is not causing the issue.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
https://e24.no/i/5B1GPE/
```

### Note:
The title for this report is generic, since it will probably be an ongoing issue, as Schibsted continues to update its various sites.
My initial reports was for `VG.no` here:
https://github.com/AdguardTeam/AdguardFilters/issues/170285/
https://github.com/easylist/easylist/issues/18245/

### Affected site:
`https://e24.no/`

### Issue:
Some videos won't load, due to a blocked cookie consent.

### Example:
`https://e24.no/i/5B1GPE/`

### Screenshots:
* **With blocked cookie consent:**

![filterissue-e24-2](https://github.com/uBlockOrigin/uAssets/assets/9382674/65abc05d-9e41-47bd-94c6-42eb0626dbcf)

* **After the cookie consent popup has been allowed, and manually declined:**

[...]

PR #33162, opened 2026-06-03
Fix voelkner.de consent

### LINKED ISSUE #33161: voelkner.de: breakage
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184), [_Spotify_](https://www.github.com/uBlockOrigin/uAssets/issues/22198) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://www.voelkner.de/`

### Category

breakage

### Description

Cookie banner is blocked by filter list "EasyList – Cookie Notices" but interaction with site is blocked unless cookie banner is shown and filled in

### Browser name and version

Firefox 151.0.3

### Other extensions used

None

### Country

Germany

### Screenshot(s)

<details>
<summary>Screenshot(s)</summary>

Image

https://github.com/user-attachments/assets/bf907c67-cd18-4980-acf1-ce47ddbb6b73

Note: These screenshots were made with a fresh Firefox profile, not the one that the metadata in the next section is from.

</details>

### Configuration

<details>

```yaml
uBlock Origin: 1.70.0
Firefox: 151
filterset (summary):
 network: 176592
 cosmetic: 184200
 scriptlet: 52693
 html: 3747
listset (total-discarded, last-updated):
 added:
  adguard-cookies: 38597-93, now
  ublock-cookies-adguard: 5670-80, now
  fanboy-cookiemonster: 46533-3791, now
  ublock-cookies-easylist: 5670-5670, now
  adguard-popup-overlays: 31277-226, now
  adguard-mobile-app-banners: 7396-104, now
  adguard-other-annoyances: 16107-201, now
  adguard-widgets: 3113-21, now
  easylist-annoyances: 5201-423, now
  easylist-chat: 275-18, now
  fanboy-ai-suggestions: 244-2, now
  easylist-newsletters: 9311-713, now
  [3 lists not shown]: [too many]
 default:
  user-filters: 4-0, never
  ublock-filters: 52467-70, now
  ublock-badware: 10340-28, now
  ublock-privacy: 4024-2, now
  ublock-unbreak: 2829-1, now
  ublock-quick-fixes: 490-14, now
  easylist: 91879-127, now
  easyprivacy: 55804-33, now
  urlhaus-1: 24285-6, now
  plowe-0: 3508-1024, now
filterset (user): [array of 4 redacted]
trustedset:
 added: [array of 2 redacted]
userSettings: [none]
hiddenSettings: [none]
supportStats:
 allReadyAfter: 461 ms (selfie)
 maxAssetCacheWait: 111 ms
 cacheBackend: indexedDB
popupPanel:
 blocked: 11
 network:
  voelkner.de: 1
  go-mpulse.net: 2
  googletagmanager.com: 1
  kameleoon.eu: 1
  newrelic.com: 1
  nr-data.net: 2
  re-in.de: 3
 extended:
  ###js_84fa0b8c9d7da99d162659e11c5e4028_container
```
</details>

ryanbr:


stephenhawk8054:
`vg.no` was addressed here: https://github.com/uBlockOrigin/uAssets/commit/0dc0d44009e3ebbd6ab898c5ff04edbb01b1f406

Better to address these sites separately, like at https://github.com/uBlockOrigin/uAssets/issues/22295

gorhill:
Yesterday I had tried that sort of filter shortly after my other proposed ones in the other thread:

    voelkner.de##+js(trusted-click-element, 'button[data-cookie_consent="0"]')

However I found out that a delay was needed otherwise this doesn't work. Stepping through the code, I see that without delay, the mutation observer somehow is not triggered after the DOM is finally filled with the `dialog` element.

So I have changed the code to no longer rely on mutation observer but rather on `requestIdleCallback` to wait for the target element to become part of the DOM. This allows the current case to be fixed without requiring a delay, and maybe this will fix more other such cases.

stephenhawk8054:
> with requiring a delay

You mean "without requiring a delay"?

gorhill:
Well, Safari doesn't support `requestIdleCallback` so I was testing with `requestAnimationFrame` and the problem still exists there with this method. Investigating further, turns out that without a delay the element is found and clicked too early -- `requestIdleCallback` was inherently causing a delay to occur so that in the end it worked.

So I am wondering now: should we click only elements which are visible? There is a method to check visibility, so the scriptlet could use this before clicking. Could this cause issues to existing `trusted-click-element` filters?

---

I added the `when-visible:` modifier for selectors: if a selector is prepended with `when-visible:`, the matching element will be clicked only once it becomes visible. <https://github.com/gorhill/uBlock/commit/aa028eacd0d60b75f43d8acf1daaaa3ff3c80da7>. Example:

    voelkner.de##+js(trusted-click-element, 'when-visible:button[data-cookie_consent="0"]')

So this changes nothing for existing `trusted-click-element` filters.

stephenhawk8054:
> should we click only elements which are visible

There are some cases where the sites hide the reject buttons but we can still click on them. Would this affect these cases?

gorhill:
I introduced the prefix `when-visible:` for when we want to wait for visibility, so all existing filters are unaffected by the addition of this prefix.

---

The biggest change from the latest commit is that the scriptlet no longer depends on mutation observer to wait for targeted elements to become available, hopefully no regression arise from this change.

PR #33175, opened 2026-06-04
Fixing scroll block on home-appliances.philips

griffinallen:

`home-appliances.philips`

### Describe the issue

This page is not usable or scrollable. This is because the cookie notice is cosmetically hidden, but there remains an overlay. In order to get rid of the overlay and restore functionality to the page, we need to set a trusted click element to action the dialog.

### Screenshot(s)

image

### Versions

- Browser/version: Brave 1.90.128

### Notes
- This fix is in response to numerous reports of scroll being blocked in the page on Brave


stephenhawk8054:
Do these work?

```adb
home-appliances.philips##vers-anonymous-consent-management-banner
home-appliances.philips##body.modal-open:style(overflow: auto !important;)
```

griffinallen:
Good suggestion @stephenhawk8054. I have tested and that works. I have created a PR for Easylist with those changes. Closing this PR.

PR #33205, opened 2026-06-06
unbreak: add toongo.io to wgplayer game player exceptions

Akshat752:

`toongo.io`

### Describe the issue

toongo.io uses wgplayer for its game player. The ||wgplayer.com^ block prevents games from loading on the site. I had to allow wgplayer on toongo.io so it would stop breaking games. 

### Versions

- Browser/version: Brave 1.91.168
- uBlock Origin version: 1.71.0

### Notes


gorhill:
Thanks.

Issue #33328, opened 2026-06-16
console.runpod.io: breakage


LarveyOfficial:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184), [_Spotify_](https://www.github.com/uBlockOrigin/uAssets/issues/22198) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). Comment in the old issue threads even when they are closed or even if you have a different problem.
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [ ] I did not answer truthfully to **ALL** the above checkboxes.

### URL(s) where the issue occurs.

```yaml
https://console.runpod.io/deploy
```

### Description

[...]

PR #33329, opened 2026-06-16
Remove posthog /flags/ block

LarveyOfficial:

`https://console.runpod.io/deploy`

### Describe the issue

Explained in full detail here: https://github.com/uBlockOrigin/uAssets/issues/33328
tldr; blocking `/flags/` doesn't add anything towards privacy, and instead just prevents websites that use PostHog's feature flags from loading

### Screenshot(s)

image

### Versions

- Browser/version: Firefox 151
- uBlock Origin version: 1.71.0

### Settings

- Default settings

### Notes
https://github.com/uBlockOrigin/uAssets/issues/33328

### DIFF
--- a/filters/privacy.txt
+++ b/filters/privacy.txt
@@ -2760,8 +2760,6 @@ abs-cbn.com##+js(set-session-storage-item, VR-INJECTOR-INSTANCES-MAP, $remove$)
 /ingress/*^ip=
 ! https://github.com/AdguardTeam/AdguardFilters/issues/213386
 /hog/*^ip=
-! https://github.com/uBlockOrigin/uAssets/issues/33267
-/flags/*^ip=
 
 ! https://edikted.com/
 ! https://github.com/uBlockOrigin/uAssets/issues/29131


VillainsRule:
pardon me if i misunderstand posthog, but aren't flags *based on* tracking that's unique to each user? are flags random A/B groups or do they have variable-based logic?

LarveyOfficial:
@VillainsRule it can be both, they *can* be based on tracking (but it doesn't do the tracking itself), `/flags` just requests the feature flag states. If no tracking has previously been provided (which it isn't since PostHog's `/hog` endpoint is already blocked by uBlock) then it just gives you either a random flag value (if its a 50/50 A/B group) or just the default state of the flag.

But tracking itself is not done by the `/flags` endpoint.

VillainsRule:
> @VillainsRule it can be both, they _can_ be based on tracking (but it doesn't do the tracking itself), `/flags` just requests the feature flag states. If no tracking has previously been provided (which it isn't since PostHog's `/hog` endpoint is already blocked by uBlock) then it just gives you either a random flag value (if its a 50/50 A/B group) or just the default state of the flag.
> 
> But tracking itself is not done by the `/flags` endpoint.

good to know, appreciate it :heart:

uBlock-user:
Answer the question asked in https://github.com/uBlockOrigin/uAssets/issues/33328

PR #33458, opened 2026-06-25
[nsfw] motherless.com moved to motherless.xxx

smartacephale:


Yuki2718:
Thanks, but please don't change `thirdparties/easylist/easylist-annoyances.txt`. It's an external list self-hosted for re-distribution.

PR #33826, opened 2026-07-23
Fixes cookie consent https://metlifeeap.one.telushealth.com/home

ryanbr:


Yuki2718:
I've merged https://github.com/uBlockOrigin/uAssets/pull/33829 , so the change for `20min.ch` won't be needed.

PR #34031, opened 2026-08-09
bilibili.com: remove tracking params

DeepChirp:
There is already a rule `||bilibili.com^$removeparam=vd_source` that can be used to remove the `vd_source` parameter. However, the parameter is automatically re-added to the URL a few seconds after video playback begins. I discovered that this is caused by the following JavaScript logic:

```js
this.$router.replace({
    query: zm(zm({
    }, this.$route.query), {
    }, {
    vd_source: this.vdSource
    }),
    hash: this.$route.hash
})
```

Therefore, we can replace `vd_source: this.vdSource` with `vd_source: undefined` to prevent this tracking parameter from being added to the address bar.


stephenhawk8054:
That list is just for `removeparam` filters. You can put that filter in `privacy.txt` list.

DeepChirp:
> That list is just for `removeparam` filters. You can put that filter in `privacy.txt` list.

I have moved this rule to `privacy.txt`.

PR #34117, opened 2026-08-14
Update unbreak.txt with ksp.co.il

venusddev:
Added rules to handle cookie prompts and potential adblock detection.

https://ksp.co.il

### Describe the issue

When an adblocker blocks the cookie prompt in ksp.co.il, an error appears that lets the user know there has been an error loading the page and they should try again in 10 seconds or later

### Screenshot(s)

<details>

image

</details>

### Versions

- Browser/version: 151.0.7922.137, brave v1.93.136
- uBlock Origin version: 1.73.0

### Settings

tried turning on annoyance filters, and well a workaround that works is disabling the ad blocker/whitelisting the script if it gets blocked(it seems to sometimes get blocked and sometimes not get blocked)

### Notes

Good thing to know is that the website appears to be blocking foreigners, might be blocking VPN use, and it's blocking brave browser though that should be mitigated soon:tm: as relevant pull requests I created were approved in brave's adblock list repo

### DIFF
--- a/filters/unbreak.txt
+++ b/filters/unbreak.txt
@@ -6035,3 +6035,7 @@ book.trivago.com##+js(set, RISKX.setSid, noopFunc)
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/1vkodxb/adblock_detection_on_minishortcom/
 ||imasdk.googleapis.com/js/sdkloader/ima3.js$redirect=noopjs,domain=minishort.com
+
+! ksp.co.il - breaks when cookie prompt js file is blocked
+ksp.co.il##+js(set-constant, gtag, noopFunc)
+ksp.co.il##+js(set-constant, dataLayer, emptyArr)


stephenhawk8054:
What are the steps to reproduce the issue?

I enabled `EasyList/uBO – Cookie Notices` lists and go to the link above and the site just looks like this to me:

<details>

image

</details>

I don't see any errors and I don't know what to do next.

venusddev:
ksp-cookie-consent.js needs to be blocked to replicate it properly, you can add the filter ||ksp.co.il/handle_cookies/ksp-cookie-consent.js

it should block it without it too but i found it to block it for some reason only sometimes, not sure why, the script isn't located in easylist, that js file is pretty new relatively speaking.

let me know if it still works for you, if so i'll list specific reproduction steps to make sure there isn't anything else that's playing as a factor

stephenhawk8054:
I said

> I enabled `EasyList/uBO – Cookie Notices` lists

which is already blocking `ksp.co.il/handle_cookies/ksp-cookie-consent.js` connection, but I can only reproduce the above screenshot and I can't reproduce the errors.

venusddev:
> I said
> 
> > I enabled `EasyList/uBO – Cookie Notices` lists
> 
> which is already blocking `ksp.co.il/handle_cookies/ksp-cookie-consent.js` connection, but I can only reproduce the above screenshot and I can't reproduce the errors.

almost forgot, set the page to mobile view, make sure url is ksp.co.il/mob ,and make sure in inspect element it's set to the mobile view option, it's currently affecting mobile use and not the desktop version of it for some reason

as for that, i found it to be blocking it still only half the time that's why I added to my filters that as well just in case but if it's blocking it for you then good enough, make sure site data is reset so the cookie prompt js will trigger since i found it to be a factor as well

reproduced it in titanium browser on android, 151.0.7922.71

stephenhawk8054:
Ok, I can reproduce now. Since the issue is caused by EL Cookie, it should be fixed in EL Cookie list.

```adb
ksp.co.il##+js(set, gtag, noopFunc)
ksp.co.il##+js(set, dataLayer, [])
```

@ryanbr

venusddev:
> Ok, I can reproduce now. Since the issue is caused by EL Cookie, it should be fixed in EL Cookie list.
> 
> ```
> ksp.co.il##+js(set, gtag, noopFunc)
> ksp.co.il##+js(set, dataLayer, [])
> ```
> 
> @ryanbr

alright thanks, do I create a pull request there since i see ryanbr is the easylist author over there

stephenhawk8054:
It's already fixed here: https://github.com/easylist/easylist/commit/500e4540aac12cb60913abcdd84de7e49e49a77b

venusddev:
> It's already fixed here: [easylist/easylist@500e454](https://github.com/easylist/easylist/commit/500e4540aac12cb60913abcdd84de7e49e49a77b)

oh wow that was fast thanks

PR #34148, opened 2026-08-17
Fix xhamster massive spam requests on videos [NSFW]

playonevideo:
These filters fix spam requests made by `xhamster.com` when you enter a video and/or after clicking the play button.

They spam a few `marzaent.com` in like the first second of the page, and then a lot of `hpyjmp.com` every second.

In addition, `juneworewyjyna.com` seems to be an ad server, it's already in AdGuard Base list:
https://github.com/user-attachments/assets/61328359-07a0-4ece-9f70-20b576ee2fcf

<details>

Screenshot 2026-08-16 174720
Screenshot 2026-08-17 185834
image

</details>

### DIFF
--- a/filters/resource-abuse.txt
+++ b/filters/resource-abuse.txt
@@ -213,3 +213,10 @@ igniteseurope.com##+js(no-xhr-if, igniteseurope.com/stats/)
 
 ! https://www.reddit.com/r/uBlockOrigin/comments/1vewpo7/how_to_stop_injectscriptadjustjs_spam/
 kick.com##+js(no-fetch-if, litix.io)
+
+! https://github.com/uBlockOrigin/uAssets/issues/23538#issuecomment-5306904084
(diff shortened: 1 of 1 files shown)


stephenhawk8054:
Can you share the link where you see that request?

playonevideo:
From `juneworewyjyna.com`? I did add a screenshot with the logger, where you can see the link, in my first post.

If you mean a video, it can be any video from the site, but here is one for test: `https://xhamster.com/videos/hetero-viejas-fotos-y-videos-de-maduros-y-abuelos-10-xhAt7FO` [NSFW]

playonevideo:
I've been thinking, maybe `juneworewyjyna.com` should be better placed at `EasyList` or `uBlock filters - Ads` list instead of `uBlock filters – Privacy`?

stephenhawk8054:
I can't see that domain in the above link so it's hard to report to EL. I'll move it to uBO ads lists later.

PR #34162, opened 2026-08-18
Fixing Youtube infinite reload on Brave with uBO extension

griffinallen:

`youtube.com`

### Describe the issue

This filter rule has an issue that surfaces when this rule is run in multiple environments simultaneously. For example, Brave Shields and uBO extension running on the same Brave browser instance. The script has a mutation observer and when this observer fires, the video refreshes. This triggers the mutation observer in the other script, which causes a video refresh and triggers the mutation observer in the first script. This starts a video refresh loop.

To avoid this loop, this change creates a shared variable between the two environments which indicates that the other script has triggered a video refresh. Once the other script starts running, if it sees that a video refresh is in progress, it aborts and breaks the loop.

### Screenshot(s)

image

### Versions

- Browser/version: Brave 1.95.70
- uBlock Origin version: uBlock Origin Lite 2026.818.1422

### Notes

Typically, users have been told to "choose one" adblock tool because having more than one active tool can create issues like this. However, a considerable amount of Brave users have the uBO extension enabled. As a test, we added an exception rule for this filter on Brave and saw a 30% reduction in issues being reported for Youtube on Brave. When we removed our exception rule, we saw a 30% increase in issues being reported for Youtube on Brave. 

If necessary, I can provide video or further testing. Please give suggestions if you have a better approach to this.


stephenhawk8054:
We don't want to create a global variable so that YouTube can easily know that there's uBO installed just because there's another blocker interfering.

Users should not use multiple blockers at the same time.

griffinallen:
That's understandable. Would you have any other suggestions for a fix to this? 

This is affecting a large number of Brave users  and we are hoping to find a solution that works while Brave users are using the uBO extension, instead of forcing users to pick one tool.

stephenhawk8054:
We always tell users to just use 1 blocker for years, regardless of which one users choose:

- https://github.com/uBlockOrigin/uAssets#similarly-purposed-blockers
- https://x.com/gorhill/status/1033706103782170625

Making exceptions to one tool will just lead to endless requests of making exceptions to all other tools, and make us waste more time for issues that don't present with uBO alone.

ryanbr:
@gorhill  can you consider `env_brave` since we're still shipping MV2 uBO and Shields users? https://github.com/ryanbr/uBlock/pull/2

gorhill:
https://github.com/gorhill/uBlock/commit/0c56103a40124eb8b7ac8d6b8d0b7c04b4a4e05d

ShivanKaul:
> We don't want to create a global variable so that YouTube can easily know that there's uBO installed just because there's another blocker interfering.

Yep agreed.

> Making exceptions to one tool will just lead to endless requests of making exceptions to all other tools, and make us waste more time for issues that don't present with uBO alone.

This is a pretty disappointing stance. We are trying our best to support uBO in Brave, including going out of our way to offer it in brave://settings/extensions/v2 which means a lot of work both in the browser ([with MV2-related patching](https://brave.com/blog/brave-shields-manifest-v3/)) and on our backend (especially once Google removes uBO MV2 from Chrome Web Store in 10 days). 

This is a pretty bad bug: a uBO user on Brave basically can't watch YouTube videos. We want to support uBO on Brave, but we can't if uBO's stance is "don't use uBO with other ad & tracker blocking tools".

stephenhawk8054:
> a uBO user on Brave basically can't watch YouTube videos

If they use Brave Shields' content blocking function with uBO.

---

Yes, multiple blockers interfering with each other happens all the time. That's the correct stance. We are volunteers here, not a company like Brave. We won't spend more of our little free time resources left to resolve every single conflict which don't appear with uBO alone. And vice versa, we don't go around and ask other tools to make them always compatible with uBO.

---

> We are trying our best to support uBO in Brave, including going out of our way to offer it in brave://settings/extensions/v2 which means a lot of work both in the browser (with MV2-related patching) and on our backend (especially once Google removes uBO MV2 from Chrome Web Store in 10 days).

And what is your point here? Brave is also using uBO's filters for many complicated issues (which also take a lot of work and time of the volunteers to resolve too). This sounds like we are asking for favors and Brave receives nothing at the end? What kind of logic is this?

---

Brave ignores the discussion of `env_brave` for years: https://github.com/brave/adblock-rust/issues/234 and now we are the ones who receive the complaints here?

ShivanKaul:
> And what is your point here? Brave is also using uBO's filters for many complicated issues (which also take a lot of work and time of the volunteers to resolve too). This sounds like we are asking for favors and Brave receives nothing at the end? What kind of logic is this?

I never said anyone is asking for favors, where are you getting that from? Not sure why you had the worst-possible interpretation of my comment. Brave contributes heavily to the open-source adblocking ecosystem as well. We just want uBO to work for Brave users, full stop, and are happy to put in the time and energy to make that happen.

stephenhawk8054:
You are the one who started about "mean a lot of work", not me, and I just stated the truth back. And you are the one who calls our stance "a disappointing stance", not me started talking anything about Brave. In which that stance is correct, and one of the stances we consistently tell users for years, to reserve our time and not burn out resolving conflicts with countless different blocking tools, which happen all the time for years. Again, we are volunteers here.

You are free to do what you want on your sides. If the changes on our sides don't affect how websites see uBO and easy for us to understand, maintain and adjust, we can consider. I also said about both points from beginning: https://github.com/uBlockOrigin/uAssets/pull/34162#issuecomment-5335836700 . If the proposed solutions don't fit us, we won't spend more time to resolve those conflicts. Again, we are volunteers here.

ShivanKaul:
> If the changes on our sides don't affect how websites see uBO and easy for us to understand, maintain and adjust, we can consider.

Sounds good, thanks.

uBlock-user:
> We are trying our best to support uBO in Brave, including going out of our way to offer it in brave://settings/extensions/v2 which means a lot of work both in the browser

You're doing this to promote your browser amoung other browsers, you're not doing this as a favour to uBO, don't try to paint this as a favour with that sentence.


> This is a pretty disappointing stance.

This is not a 'stance', it's a fact -- https://github.com/uBlockOrigin/uAssets#similarly-purposed-blockers You embed Brave with it's blocker which is similar to uBO in functionality and then complain about disappointing stance, if you cared about uBO so much you should have promoted uBO instead of creating your own blocker. Again, do not try to show that you're doing us any favour, we're volunteer org and nobody gets paid here unlike you, an employee who draws salary along with the rest of of your kin.

Yuki2718:
>This is a pretty bad bug: a uBO user on Brave basically can't watch YouTube videos. 

No, uBO works fine on Brave. Did you miss that the issue occurs only if both uBO and Shields are enabled at the same time? We have been doing all the best to prevent running more than one blocker concurrently and really don't want to make any exception for this stance, and I personally hope Brave to be in line with us. I remember Brave's official X account validated running uBO on top of the Shield, which is quite against this: `https://x.com/brave/status/1845950855322747046`

PR #34253, opened 2026-08-26
Hiding anti adblock annoyance on finance.kooorazoom.net

griffinallen:

`https://finance.kooorazoom.net/`

### Describe the issue

There is an anti-adblock popup on this page that needs to be hidden so users can access the page while ads continue to be blocked.

### Screenshot(s)

<details>

image

</details>

### Versions

- Browser/version: Brave 1.96.4


stephenhawk8054:
I can't reproduce. Is it a redirection from a shortener link?

griffinallen:
Good catch @stephenhawk8054. Closing here and making this fix specific to Brave as I can't repro on Firefox + uBO

PR #34292, opened 2026-08-30
Hiding anti adblock notice on hdstreamverse.xyz

griffinallen:

`hdstreamverse.xyz`

### Describe the issue

This site contains an anti-adblock notice that is easily removed with a network filter, blocking the script that serves the notice to users. This site has been reported to Brave, but the notice is also visible on Firefox+uBO.

### Screenshot(s)

image

### Versions

- Browser/version: Brave 1.96.14


stephenhawk8054:
It seems users can't disable that anti-adb modal so I'll address it in main list.

Issue #34187, opened 2026-08-21
igorslab.de: ads


ghost:
### Prerequisites

- [x] This is **NOT** a [_YouTube_](https://www.github.com/uBlockOrigin/uAssets/issues/30157), [_Facebook_](https://www.github.com/uBlockOrigin/uAssets/issues/3367), [_Twitch_](https://www.github.com/uBlockOrigin/uAssets/issues/5184), [_Spotify_](https://www.github.com/uBlockOrigin/uAssets/issues/22198) or [_a shortener/hosting site_](https://www.github.com/uBlockOrigin/uAssets/discussions/27472) report. _These sites MUST be reported by clicking their **respective links**_.
- [x] I read and understand the [policy about what is a valid filter issue](https://github.com/uBlockOrigin/uAssets/blob/master/README.md#uassets).
- [x] I verified that this issue is not a duplicate. (Use this [button](https://user-images.githubusercontent.com/585534/146582579-c32707a0-36a1-4cc5-ad50-83172c9f67a8.png) to find out.). **Comment in the old issue threads even when they are closed or even if you have a different problem.**
- [x] I did not remove any of the [default filter lists](https://github.com/user-attachments/assets/1879fc88-09bc-4de4-9e2f-3834a6ab396a), or I have verified that the issue was not caused by removing any of the default lists.
- [x] I did not enable additional or non-default filter lists, or I have identified which specific stock list is causing the issue.
- [x] I do not have custom filters/rules, or I have verified that the issue still occurs without custom filters/rules.
- [x] I have verified that the web browser's built-in content blocker/tracking protection, [network wide/DNS blocking](https://en.wikipedia.org/wiki/DNS_blocking), or my VPN is not causing the issue.
- [x] I have turned off all other extensions and the issue still persists. (exception "[Firefox Multi-Account Containers](https://addons.mozilla.org/en-US/firefox/addon/multi-account-containers)").
- [x] If this is about a breakage or detection, I have verified that it is caused by uBlock Origin and isn't a site or browser issue.
- [x] I have verified that the browser I am using is up to date with no pending updates.
- [x] I did not answer truthfully to **ALL** the above checkboxes.

### URL address of the web page

`https://www.igorslab.de/`

### Category

ads

### Description

The ads on igorslab.de didn't block. If i choose them manually to block, the first time they are blocked, the second time they didn't blocked.

### Browser name and version

the latest firefox

### Other extensions used

adblock plus

### Country

Germany

[...]

PR #34302, opened 2026-08-30
update igorslab.de

playonevideo:
The filters were again countered, so with this one, the ads are hidden again. For this one I used the element picker from uBO.
image

### DIFF
--- a/filters/quick-fixes.txt
+++ b/filters/quick-fixes.txt
@@ -529,6 +529,7 @@ igorslab.de###extras > #single-sticky + div[class]:not([id]):has(a[href][target=
 igorslab.de##.entry-content > #skio-article-bar ~ div[id][class]:not([class*="igorslab"]):has(a[href][target="_blank"] img)
 igorslab.de##a[target="_blank"]:has(img[src^="blob:"])
 igorslab.de##.herald-above-footer
+igorslab.de##[src^="blob:https://www.igorslab.de/"]
 /^https:\/\/www\.igorslab\.de\/[a-z0-9]+\/[\/a-zA-Z0-9_-]+$/$image,1p,match-case,domain=igorslab.de
 
 ! https://github.com/uBlockOrigin/uAssets/issues/34237


igorwallossek-sys:
@stephenhawk8054

I'd like to raise several specific, factual concerns about the scope of the rules merged in `73909e7` (originally proposed in #34302, following the discussion in #34187), separate from the general question of whether advertising may be blocked on this domain.

**1. Architecture context**

igorslab.de runs no programmatic advertising, no third-party ad networks, and builds no personalized user profiles. All advertising is self-hosted, manually vetted for content and destination, and served without third-party ad-tech calls. A significant portion of these banners come from industry partners the site cooperates with editorially (review samples, long-term testing relationships); in that sense they function closer to self-promotion of an existing partnership than to third-party ad-tech placement. Readers who wish to hide these specific, clearly identifiable ad elements can already do so; that is not in dispute.

**2. Editorial independence and funding**

The site does not accept payment for reviews or awards, and does not intend to. Maintaining that position requires a sustainable revenue model - producing technically detailed, independently verified content (measurement equipment, long-term testing, dedicated staff time) has real costs, and self-hosted, manually-vetted advertising is the mechanism that funds it without compromising editorial judgment on individual products. Rules that indiscriminately strip revenue-generating, non-tracking elements make that funding model harder to sustain, which is a relevant consideration when assessing the actual impact of a filter, not just its intended target.

**3. Price-comparison module (Geizhals)**

The price-comparison box is a first-party, editorially-triggered component tied to specific product articles, not a rotating ad slot. Product data, images, and shop logos are cached and served from igorslab.de itself; the reader's browser does not contact Geizhals, retailers, or any external CDN until a deliberate click occurs. No third-party cookies, tracking identifiers, or profiling data are involved in simply displaying the module. Rules that remove this element on the assumption that it constitutes advertising or tracking are, in this specific case, factually inaccurate, and the resulting breakage has caused a documented five-figure revenue loss this year.

**4. VG WORT counting pixels**

These pixels are unrelated to advertising or marketing tracking. They serve a German collective-rights-management system (VG WORT / METIS) that determines author compensation for text usage. No advertising profile, cross-site tracking, or personalization is built from this data; its sole function is usage measurement for a statutory remuneration scheme. Treating it identically to an ad tracker conflates two technically and legally distinct categories.

**5. Copyright notices, authorship attribution, and logos**

Selectors broad enough to hit generic structural containers (e.g. above-footer wrappers, div:last-of-type inside an article container) have also affected the copyright footer, authorship attribution, and the site logo in some configurations. None of these carry any advertising or tracking function; they identify the author/publisher of the content and the site itself. German copyright law includes a right of attribution and provisions protecting rights-management information, and while I'm not asserting a specific legal violation here, removing these elements is clearly outside the stated purpose of an ad/tracking filter and should not occur as an incidental side effect of a broad selector.

**6. Forum and compliance controls**

The forum embed is a separate, deliberately decoupled system (no shared login with the main site) that displays discussion metadata only. Rules that remove it based on container position or class name affect functionality unrelated to advertising. Similarly, any rule that hides privacy/consent-setting controls removes the exact interface a reader would use to exercise the control such rules are meant to protect.

**7. Selector precision**

Several of the rules in question (e.g., position-based selectors like a footer-adjacent container, or structural patterns like `div:last-of-type`, or attribute combinations such as `a[target] img[src^="blob:"]`) do not identify advertising by content or function, only by incidental markup position. Such patterns are prone to false positives as the site's template evolves, independent of any intent on the site operator's side to evade detection.

**Request**

I am not asking that advertising on this domain be exempted from blocking. I am asking that filter rules be scoped narrowly enough to target actual ad/tracking elements without incidentally removing first-party, non-commercial, partner-relationship, or compliance-related functionality that funds independent, non-pay-to-play editorial work. :)

playonevideo:
Please, no AI-written essays here.

igorwallossek-sys:
You can google me, so it's not AI, it's me. Only AI translated and corrected. If you dislike my work and my content, why you visit my site? 

@playonevideo Fair point on length. I'll shorten it if that helps, but I'd rather you engage with whether the Geizhals module actually contacts third parties before display, whether VG WORT pixels constitute tracking in any meaningful sense, or whether a footer copyright notice should be within scope of an ad filter than with how the comment was drafted. The technical claims stand or fall on their own regardless of what tool helped format them.

igorwallossek-sys:
One more quick note: If all the sites that are fighting against complete commercialization. at great effort and expense, end up closing due to a lack of revenue, you won’t need an ad blocker in the end either. That’s because the ads I host myself are visually unobtrusive, free of trackers, and don’t interfere with reading. On the contrary. And it shouldn’t have escaped your notice that we make our living from precisely this revenue and don’t sell article content or awards, unlike all those who can no longer make a living from advertising. You decide which is better. 

If you have a personal issue with me or my work, that's your business. But in the end, that kind of thing just leads to a game of cat and mouse that annoys everyone involved.

uBlock-user:
> I'd like to raise several specific, factual concerns about the scope of the rules 

This is not up for discussion. Ads are always blocked, period.

>  end up closing due to a lack of revenue, you won’t need an ad blocker in the end either

Good, can't wait for the day when I don't have to enable an extension to be able to browse the web.

> we make our living from precisely this revenue

Not our problem. We block all ads and lot more than ads if you read the description.

stephenhawk8054:
@igorwallossek-sys Users have the rights to block whatever they want on their _**OWN**_ browser clients. Those are the words from World Wide Web Consortium Ethical Web Principles, not us: https://www.w3.org/TR/ethical-web-principles/#render

> People must be able to change web pages according to their needs. For example, people should be able to install style sheets, assistive browser extensions, and blockers of unwanted content or scripts.

_**ANY**_ moral and ethical websites should respect users' choices. You are also using users' resources to display your contents, it's not like you are using your own resources for that. Users have total rights to protect themselves from unsolicited contents and network requests creeping into their _**OWN**_ machines. Those machines are theirs, not yours.

---

Adtech industry is borderline a criminal industry. When even the FBI warns people to use content blockers to protect themselves, you know that your industry is beyond saving now: https://www.ic3.gov/PSA/2022/PSA221221

> Tips to Protect Yourself
> 
> The FBI recommends individuals take the following precautions:
> 
> Use an ad blocking extension when performing internet searches. Most internet browsers allow a user to add extensions, including extensions that block advertisements.

If your business depends totally on that immoral and unethical industry, it deserves to be gone. If you have your own ethical products that value, pitch it and sell to users. Stop relying on ads and trackers.

PR #34318, opened 2026-08-31
Hiding anti adblock warning on games360rgh.com

griffinallen:

`games360rgh.com`

### Describe the issue

This site has an anti adblock popup that can be easily dismissed through disabling the function that serves the popup. This popup was reported on Brave, but is also reproducible on Firefox+uBO.

### Screenshot(s)

image

### Versions

- Browser/version: Brave 1.96.14


griffinallen:
This fix may belong in a different list. If so, could you please explain the rationale as to where to put different types of anti-adblock fixes?

PR #34322, opened 2026-08-31
Fix empty alternative in ClickFix clipboard regex

### LINKED ISSUE #34321: prevent-clipboard-write regex contains empty alternative and matches all clipboard text
### Description

The global `prevent-clipboard-write` regex in `filters/filters-general.txt` contains an empty alternative:

```regex
\bcmd\b||^iex\(
```

Because `A||B` has an empty alternative, the regex matches every clipboard string. This causes benign clipboard writes to be blocked with the ClickFix warning.

For example, copying this Home Assistant entity ID is blocked:

```text
sensor.chambre_humidite
```

Minimal JavaScript reproduction:

```js
const pattern = /\bcmd\b||^iex\(/msi;
pattern.exec("sensor.chambre_humidite"); // [""]
```

The empty matched string confirms that no suspicious token in the clipboard text caused the match.

### Suggested fix

Remove the duplicate alternation character:

```diff
- ^conhost .*?--headless .*?\bcmd\b||^iex\(i(rm|wr) 
+ ^conhost .*?--headless .*?\bcmd\b|^iex\(i(rm|wr) 
```

Source: https://github.com/uBlockOrigin/uAssets/blob/master/filters/filters-general.txt#L3122

cnotin:
Fixes #34321.

Remove the duplicate alternation character from the ClickFix clipboard regex. The previous `||` introduced an empty alternative, causing the regex to match every clipboard string.

All three synchronized copies are updated:

- the global `prevent-clipboard-write` rule
- the exception for `community-scripts.org` and `massgrave.dev`
- the specific `sites.google.com` rule

Tested that `sensor.chambre_humidite` no longer matches, while representative `cmd` and `iex(` inputs still match.


cnotin:
This regression was introduced by @stephenhawk8054 in commit 2c848f1c95ea83eace1b65fd9c32fbf7c75ec0a0 while broadening the iex(irm ...) pattern to also cover iwr: https://github.com/uBlockOrigin/uAssets/commit/2c848f1c95ea83eace1b65fd9c32fbf7c75ec0a0
If you could confirm please

u-RraaLL:
Thanks. gorhill missed the PR and already fixed it in: https://github.com/uBlockOrigin/uAssets/commit/dffea528430508e3220a8485f399766f8d3ec6a8

PR #34387, opened 2026-09-05
fake featherwallet site | https://github.com/feather-wallet/feather/issues/303

adblockenjoyer:
phishing

flagged by Kaspersky

https://www.virustotal.com/gui/url/299764cd2af6f3b52dc2597416e5808d04a0f967366558c21c231bbe5bfb9ee7

https://github.com/feather-wallet/feather/issues/303


stephenhawk8054:
About the wildcard filter, currently there are no other domains. We don't add wildcard filters based on theoretical cases. We only consider when there are many more examples to form a pattern.

adblockenjoyer:
> About the wildcard filter, currently there are no other domains. We don't add wildcard filters based on theoretical cases. We only consider when there are many more examples to form a pattern.

I remember there was at least one other phishing domain in the past. It seems to be specifically targeted over time. I'm hoping you can help with being a bit more proactive for the users.

I understand your stance though.

stephenhawk8054:
The thing is, we are the ones who will receive all kinds of false positive complaints, not the people who **_want_** to block those domains. And false positive situation is worse given that uBOL can't update the filter lists like uBO. Most people just want us to block all the things they want and then they are gone, leaving us to deal with all issues later. Any filters added need a firm ground in order to prepare arguments for those complaints. At least give us 3 cases of badwares or the pattern involves randomizing in order for us to consider wildcards.

adblockenjoyer:
in the past: feather-wallet.org - https://github.com/feather-wallet/feather/issues/198

